PR #11, opened 2022-01-13
Support STM32F3DISCOVERY board

marnix:
So this is my first humble success in the microzig world: a red blinking led on STM32F3DISCOVERY.

Completely incomplete, of course, but it is _something_. :-)

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -38,7 +38,7 @@ pub fn build(b: *std.build.Builder) !void {
 
             const exe = try microzig.addEmbeddedExecutable(
                 b,
-                "test-" ++ tst.name ++ "-" ++ cfg.name,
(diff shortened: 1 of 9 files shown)


marnix:
For now I don't know how to deal with the duplication between existing CPU file cortex-m3.zig and new cortex-m4.zig: the only difference is the top-of-stack (which actually seems to be part of the _chip_ definition, not the _CPU_ definition), and I don't see how to abstract that out currently.

Apart from that, this PR seems pretty complete to me.

Removing draft status; let me know if anything else is required.  Thanks!

marnix:
@vesim987 @mattnite @MasterQ32 Will someone merge this? Or is there something I have to do first? Thanks!

marnix:
Thanks @MasterQ32 !

PR #18, opened 2022-02-15
Packages and stack

vesim987:


mattnite:
This degrades the initial UX because it looks like you have to manually set the stack size. Is there any reason to distinguish between stack and heap anyways since we have allocators in userspace?

MasterQ32:
> This degrades the initial UX because it looks like you have to manually set the stack size. Is there any reason to distinguish between stack and heap anyways since we have allocators in userspace?

This actually increases long-term UX as you will now easily find stack overflows by triggering "SIGBUS" on most devices instead of silently corrupting the heap. Imho, this is an important security and safety feature

mattnite:
I understand the value but the defaults are wrong. If someone just uses the default, as soon as they use more than 1024 bytes of the stack they'll just have a runtime crash which will be especially difficult to debug if they are new.

Instead the default value for the stack size config should be null, so that people who know what they're doing can opt-in. Otherwise the full ram section is used for the stack.

PR #20, opened 2022-02-19
stm32f3discovery/stm32f303 Initial UART1 support, including @panic()

marnix:


marnix:
This is currently not complete, but for now this will have to be it from me for a little while.

Please merge if this is deemed OK, and/or let me know if there are things you'd like to see different.

Thanks in advance!

marnix:
It looks like UART1 support is pretty complete now.  What would be needed to get this merged?

marnix:
Back to draft, because unfortunately :-( #23 was merged in the meantime, and with that
- this branch doesn't compile anymore;
- the new regz-generated `Mmio()` doesn't support the `writeRaw()`
  which I added in core/mmio.zig in 860cb8e.
   * Note that I'm wondering why this code is now in the generated registers.zig files,
     instead of in mmio.zig,
     since this `Mmio()` and `MmioInt()` code seems to be completely generic...

marnix:
@mattnite After #24 has been merged, this should work fine.  Removing 'draft' status; could you please review/merge?  Thanks in advance!

mattnite:
🙏

PR #21, opened 2022-02-20
Gd32v Support

kassane:
## References

[SVD - Sipeed](https://github.com/sipeed/platform-gd32v/blob/master/misc/svd/GD32VF103.svd)
[GigaDevice device-info](https://www.gigadevice.com/microcontroller/gd32vf103cbt6/)
[Sipeed -Logan Nano](https://longan.sipeed.com/en/)
[Manual - Firmware Library (PDF)](http://www.gd32mcu.com/data/documents/yingyongbiji/GD32VF103_Firmware_Library_User_Guide_V1.0.pdf)


kassane:
cc: @mattnite @vesim987

Is it possible to automate the translation or read the start.S file in zig build or microzig?
I found the files needed to support the chip in the GD32v firmaryLibrary [¹], but am having difficulty converting them.
And comparing on the rust solution [²] for this I was in doubt of the real support with llvm.


[¹]: https://github.com/riscv-mcu/GD32VF103_Firmware_Library/blob/master/Firmware/RISCV/env_Eclipse/start.S
[²]: https://github.com/riscv-rust/gd32vf103xx-hal/blob/master/assemble.sh

kassane:
# Status

- Logan Nano board added

By adding some changes you find errors in the api during the build:

**Uart error:**
```bash
./src/core/uart.zig:6:28: error: container '.chip' has no member called 'Uart'
    const SystemUart = chip.Uart(index);
                           ^
./src/modules/boards/logan-nano/logan-nano.zig:60:29: note: called from here
    const uart0 = micro.Uart(0).getOrInit(.{
                            ^
./src/modules/boards/logan-nano/logan-nano.zig:59:44: note: called from here
pub fn debugWrite(string: []const u8) void {
```

**Pin error:**
```bash
./src/core/pin.zig:22:13: error: container '.chip' has no member called 'parsePin'
        chip.parsePin(@field(board.pin_map, spec))
            ^
./tests/blinky.zig:9:36: note: called from here
        .@"Logan Nano" => micro.Pin("PC13"), // error in 'parsePin'
                                   ^
./tests/blinky.zig:4:17: note: called from here
const led_pin = if (micro.config.has_board)
```

ekliptik:
I suggest writing a small HAL for STM32F103 first. It is better documented, the GD32VF is a binary compatible clone. Right now your PR uses F303 HAL which is incompatible, for example, instead of a single GPIO mode register there are two with different fields, so it's not just renaming registers that you have to make it work
By the way, it's Longan, not Logan, also, [never git push force](https://salferrarello.com/never-git-push-force/).

kassane:
Hi @ekliptik 

Nice tip! I already made some improvements, however didn't have time to validate and send the new commit containing these improvements and including the name of the fixed board.

But thanks for the review!

mattnite:
@kassane looks like some unused parameters are causing the build to fail

kassane:
> @kassane looks like some unused parameters are causing the build to fail

fix

build.zig
```diff
@@ -24,6 +24,9 @@ pub fn build(b: *std.build.Builder) !void {
         BuildConfig{ .name = "boards.stm32f3discovery", .backing = Backing{ .board = boards.stm32f3discovery } },
         BuildConfig{ .name = "boards.stm32f4discovery", .backing = Backing{ .board = boards.stm32f4discovery } },
         BuildConfig{ .name = "boards.stm32f429idiscovery", .backing = Backing{ .board = boards.stm32f429idiscovery }, .supports_uart_test = false },
+        // BuildConfig{ .name = "chips.stm32f103x8", .backing = Backing{ .chip = chips.stm32f103x8 } },
+        // BuildConfig{ .name = "chips.gd32vf103x8", .backing = Backing{ .chip = chips.gd32vf103x8 } },
+        // BuildConfig{ .name = "boards.longan_nano", .backing = Backing{ .board = boards.longan_nano } },
```
mattnite, line 29: these should be uncommented

mattnite (changes requested):

build.zig
kassane: done!!

mattnite:
are you able to confirm that this is how interrupts work on the chip, and is it consistent with other risc chips?

kassane:
> are you able to confirm that this is how interrupts work on the chip, and is it consistent with other risc chips?

First, you can't confuse `gd32` (bluepill clone) with `gd32v` even though the pinouts are compatible. The main difference is that gd32v has a higher clock speed.

The IRQ mechanism (ECLIC) of this device is unique to the Nuclei manufacturer (core Bumblebee) since RISC-V has no default interrupt mechanism.
So far the interrupt implementation is not finished as it would need to use pieces of code from the official SDK and toolchain (nuclei-gcc fork), even attach an assembly file as it was done in the rust project (see [this](https://github.com/riscv-rust/gd32vf103xx-hal/pull/14/files)).

The SVD provided is also not complete, many complain about certain details like [undeclared core timer](https://github.com/riscv-rust/gd32vf103-pac/issues/2).
And `regz` did not translate it completely (missing clusters), so it will have to be done manually with the respective time.

### Reference

- [ECLIC - Nuclei doc](https://doc.nucleisys.com/nuclei_spec/isa/eclic.html)

mattnite:
okay if that's the case then we shouldn't be putting arm interrupt table code in the risc processor code, similar to how we add a skip for the uart example, please add a skip mechanism for the interrupt example.

src/modules/cpus/rv32-imac/riscv32.zig
```diff
@@ -0,0 +1,921 @@
...
+pub const CAUSE_FAULT_STORE = 0x7;
+pub const CAUSE_USER_ECALL = 0x8;
+pub const CAUSE_SUPERVISOR_ECALL = 0x9;
+pub const CAUSE_HYPERVISOR_ECALL = 0xa;
+pub const CAUSE_MACHINE_ECALL = 0xb;
+
+const VectorTable = microzig.chip.VectorTable;
+
+// will be imported by microzig.zig to allow system startup.
+pub var vector_table: VectorTable = blk: {
```
mattnite, line 908: oh do you have to do this because of #64 ? if interrupt implementations are per-chip for riscv then this code shouldn't be here

mattnite (changes requested):

src/modules/cpus/rv32-imac/riscv32.zig
kassane: So faced with this problem in addition to removing the tableVector would also need to comment out the existing builds.
Like the stm32f103 it cannot be built at the present time because the project requires interrupts.

build.zig
```diff
@@ -42,6 +44,8 @@ pub fn build(b: *std.build.Builder) !void {
         for (all_tests) |tst| {
             if (tst.uses_uart and !cfg.supports_uart_test) continue;
             if ((cfg.backing.getTarget().cpu_arch.?) == .avr and tst.on_avr == false) continue;
+            // Error: "gd32v - missing interrupts"
+            if ((cfg.backing.getTarget().cpu_arch.?) == .riscv32) continue;
```
mattnite, line 48: this will make it so that no test programs are compiled :/, can you just add a flag for `supports_interrupt_test` so that we get something compiling?

mattnite (changes requested):

build.zig
```diff
@@ -24,16 +24,18 @@ pub fn build(b: *std.build.Builder) !void {
         BuildConfig{ .name = "boards.stm32f3discovery", .backing = Backing{ .board = boards.stm32f3discovery } },
         BuildConfig{ .name = "boards.stm32f4discovery", .backing = Backing{ .board = boards.stm32f4discovery } },
         BuildConfig{ .name = "boards.stm32f429idiscovery", .backing = Backing{ .board = boards.stm32f429idiscovery }, .supports_uart_test = false },
+        BuildConfig{ .name = "chips.gd32vf103x8", .backing = Backing{ .chip = chips.gd32vf103x8 }, .supports_interrupt_test = false },
+        BuildConfig{ .name = "boards.longan_nano", .backing = Backing{ .board = boards.longan_nano }, .supports_interrupt_test = false },
     };
 
-    const Test = struct { name: []const u8, source: []const u8, uses_uart: bool = false, on_avr: bool = true };
+    const Test = struct { name: []const u8, source: []const u8, uses_uart: bool = false, on_riscv32: bool = true, on_avr: bool = true };
```
mattnite, line 31: why do we have both a `supports_interrupt_test` and `on_riscv32`? isn't one enough? otherwise it will get confusing have to turn off both when we get this working.

mattnite (changes requested):

mattnite (approved):

mattnite:
🙏 thanks for getting the first riscv patch in!

kassane:
> 🙏 thanks for getting the first riscv patch in!

Thank you for the patience. 😅

PR #23, opened 2022-03-03
use register code generated by regz

mattnite:


marnix:
@mattnite Is there a specific reason why core/mmio.zig `mmio()` and `MMIO()` are now ignored, preferring equivalent generated code (now called `mmio()` and `Mmio()` + `MmioInt()`)?  It looks like that generated code is completely generic?

Asking because I had extended the original generic code (with a `writeRaw()` that was helpful for one use case), but now it looks like regz has to be updated instead, and all registers.zig regenerated?

That doesn't seem to be an improvement to me...

Or am I missing something?

marnix:
@mattnite Thanks for that clarification! Could you perhaps add `writeRaw()` to regz and to these generated files, as I did in https://github.com/ZigEmbeddedGroup/microzig/pull/20/commits/860cb8e00f1e2a8e1eeab7b672a6425af3925dd2 ?

I'd do it myself, but
- I'm not sure I see how regz does things yet, and
- I don't know which exact .svd files you used...

Thanks in advance!

PR #26, opened 2022-03-12
Initial I2C support

marnix:
This is an initial I2C API, based on discussions with @MasterQ32 in Discord \#microzig, with an initial implementation for STM32F303.


src/core/i2c.zig
```diff
@@ -0,0 +1,121 @@
...
+                    /// RESTART to a new transfer, invalidating this object.
+                    pub fn restartTransfer(self: Self, comptime new_direction: Direction) Transfer(new_direction) {
+                        self.device.internal.restart();
+                        return self.device.transfer(new_direction);
+                    }
+                },
+            };
+        }
+
+        pub fn transfer(self: *const Device, comptime direction: Direction) Transfer(direction) {
```
MasterQ32, line 74: This should send the device selector as well as the address

src/core/i2c.zig
```diff
@@ -0,0 +1,121 @@
...
+            return switch (direction) {
+                .read => struct {
+                    const Self = @This();
+
+                    device: *const Device,
+
+                    pub const Reader = std.io.Reader(Self, ReadError, readSome);
+
+                    pub fn reader(self: Self) Reader {
+                        return Reader{ .context = self };
```
MasterQ32, line 25: With this implementation, several reads will perform restarted reads instead of a single continuous read. These are semantically different in I²C

MasterQ32 (changes requested):
Please make the `transfer` function start the transfer and make `readSome` and `writeSome` don't send start conditons.

marnix:
Open issue (apart from the merge conflict that has arisen): @MasterQ32 correctly observed that with the current code, "if i read two bytes from the reader() i will actually read one byte twice".

The issue  seems to be that STM32 microcontrollers do some I2C things automatically, making e.g. "do only START" impossible (?) to implement.  I think is why I implemented things the way I did, with an implicit START.
Sources where others had similar issues:
- <https://community.st.com/s/question/0D50X0000C3CFCWSQ4/using-stm32h743-i2c-with-cr2-reload-and-nbytes0-gives-tc-interrupt-instead-of-tcr> which led me to
- <http://adastra-soft.com/an-easy-to-use-i2c-library-for-stm32h7xx/>

Setting to Draft for now.

marnix:
@MasterQ32 I've rewritten large parts-- could you perhaps have a look again, and see what you think?

Keeping on Draft, still need to test with latest master.

marnix:
@MasterQ32 Un-Draft-ing.  As I said, in [009d3fc](https://github.com/ZigEmbeddedGroup/microzig/pull/26/commits/009d3fc3fa5c7636197f04cadad7448ffb845b70) and later commits I've rewritten the STM32F303 implementation of the API, and documented the API to give STM32-specific restrictions.

Could you please have a look at this again?

Thanks very much!

src/core/i2c.zig
marnix: 'Fixed' in commit [009d3fc](https://github.com/ZigEmbeddedGroup/microzig/pull/26/commits/009d3fc3fa5c7636197f04cadad7448ffb845b70) and later: The second read call on an STM32F303 will now panic, since it is basically impossible to implement a multi-part read.

Note that a multi-part write will work on STM32F303, at least if the number of bytes is < 256; I cannot really test >= 256 writes (single-part or multi-part), but I'm not at all sure they will work...

Both have been achieved by doing a 'lazy' (re)START during the first read or write call-- that really seems to be the only way the current API can be properly implemented on STM32 microcontrollers...

MasterQ32 (approved):

MasterQ32:
Yep, looks good now!

PR #28, opened 2022-03-14
Greatly simplifies the package logic by using a proxy-to-root package.

MasterQ32:
This will simplify the package dependency tree for all packages:
```
src/core/microzig.zig
├── app
│   └── microzig
├── board
│   └── microzig
├── chip
│   └── microzig
├── cpu
│   └── microzig
└── microzig-config
```


MasterQ32:
I need to test the successful compilation for the device i have, so check if i didn't break anything, but i think this is a huge success

mattnite:
:pray:

PR #29, opened 2022-03-15
Avr interrupt overloading

mattnite:
requires #28


src/modules/cpus/avr/avr5.zig
```diff
@@ -24,39 +25,63 @@ pub inline fn cbi(comptime reg: u5, comptime bit: u3) void {
...
+                    .Unspecified => struct {
+                        fn wrapper() callconv(.C) void {
+                            if (calling_convention == .Unspecified) // TODO: workaround for some weird stage1 bug
+                                @call(.{ .modifier = .always_inline }, handler, .{});
+                        }
+                    }.wrapper,
+                    else => @compileError("Just leave interrupt handlers with an unspecified calling convention"),
+                };
+
+                const options = .{ .name = field.name, .linkage = .Strong };
```
MasterQ32, line 70: I would probably add an `_isr_` prefix to the symbol name here, and we need to make sure the symbol is located directly after the vector table (same for `_start` :thinking: )

PR #33, opened 2022-03-19
Make sure vector_table is correctly exported

marnix:
It needs to be a non-`comptime` `var` for `@export` to work properly.

The only 'documentation' for this behavior currently seems GitHub comment
https://github.com/ziglang/zig/issues/5157#issuecomment-618933196 .

This issue was introduced in 1c17304 for PR #27,
which broke at least ARM-based STM32F303.


marnix:
Replaced by better #34, closing.

PR #48, opened 2022-05-16
add hal package option

mattnite:
The idea here is that someone could write an `nrf52` package that makes working with registers more readable. Similar to how the sdk has some nice-to-use functions for setting up gpio. This sort of package would depend on microzig only. 

Wrt to the generalized interfaces we've discussed, I see packages like these having implementations that can fully take advantage of hardware, and can return a generalized structure. This would be analogous to how you could have a `CountingReader`, and a `GzipReader`, but you call `reader()` on them to get a generalized reader.

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -28,6 +28,10 @@ fn root() []const u8 {
 
 pub const BuildOptions = struct {
     packages: ?[]const Pkg = null,
+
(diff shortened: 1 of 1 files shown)


cvoges12:
Just making sure this is mentioned #48

PR #50, opened 2022-05-27
Adds logo

MasterQ32:
I designed a project logo based on an idea by Loris and a scribble i did on the phone:

It also has two variants with different fonts and the stylized writing `MicroZig`:

**Variant 1:**

**Variant 2:**

I personally dig variant 1 more. When we chose one, i'll create optimized SVGs for the chosen text variant and also create a light/dark variant for the right background.

Which variant do we chose? @mattnite @vesim987


MasterQ32:
Looks like everyone likes the clean font more, so let's settle on this.

MasterQ32:
Created bright and dark variant:

MasterQ32:
Also added a variant that uses CSS styles to adjust to color scheme:

PR #53, opened 2022-07-01
Add support for multiple clock domains

rbino:
While this does not automate clock configuration yet, at least it allows using
microzig with applications that configure the clock themselves and then declare
the speeds of the various domains.
Without this, all peripheral depending on different clock domanins (e.g. Uart or
I2C on STM32 devices) could only work with the reset frequency.


mattnite (approved):
I think this is a great step in the right direction 🚀

PR #77, opened 2022-09-07
Fixes #72

### LINKED ISSUE #72: Expand EmbeddedExecutable
Expose common functions like `install()` and `setBuildMode()`, also `installRaw()`

r4gus:
Added some function wrappers to EmbeddedExecutable that expose commonly used functionalities of LibExeObjStep.

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -48,6 +48,37 @@ pub const EmbeddedExecutable = struct {
             }
         } else @panic("app package not found");
     }
+
+    pub fn addPackagePath(exe: *EmbeddedExecutable, name: []const u8, pkg_index_path: []const u8) void {
+        exe.addPackage(Pkg{
+            .name = exe.inner.builder.allocator.dupe(u8, name) catch unreachable,
+            .source = .{ .path = exe.inner.builder.allocator.dupe(u8, pkg_index_path) catch unreachable },
+        });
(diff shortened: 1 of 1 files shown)


src/main.zig
```diff
@@ -48,6 +48,37 @@ pub const EmbeddedExecutable = struct {
             }
         } else @panic("app package not found");
     }
+
+    pub fn addPackagePath(exe: *EmbeddedExecutable, name: []const u8, pkg_index_path: []const u8) void {
```
mattnite, line 52: sorry for the late review! btw you could simplify and just call `exe.inner.addPackagePath()`

mattnite (changes requested):

src/main.zig
r4gus: Hi, no problem. Unfortunately, the added package can't be found if you call `inner.exe.addPackagePath()` directly.

Example:
```
./src/main.zig:3:14: error: unable to find 'zbor'
const cbor = @import("zbor");
```
I use Linux and Zig version `0.10.0-dev.3685+dae7aeb33`.

mattnite: ooh yes, I forgot about the package structure, you did the correct thing

mattnite (approved):

PR #79, opened 2022-09-14
Add executable methods and fix recursive call

kuon:
Add executable methods (addOptions, addObjectFile) and fix recursive call in addIncludePath.

For #72.


mattnite:
probably works now, let's merge and :pray:

PR #88, opened 2022-10-15
 ATmega328p uart

PhilippWendel:
uart-sync.zig
- Added support for ATmega328p
  - "PB5" is the builtin led of Arduino UNO

atmega328p.zig:
- Fixed compile error in fn computeDivider
  - std.math.cast changed from error  to optional
  - https://github.com/ziglang/zig/commit/0e6285c8fc31ff866df96847fe34e660da38b4a9
- Added pub const clock_frequencies
  - 16 MHz is the default freq. of Arduino UNO

### DIFF
--- a/src/modules/chips/atmega328p/atmega328p.zig
+++ b/src/modules/chips/atmega328p/atmega328p.zig
@@ -17,6 +17,10 @@ pub const clock = struct {
     };
 };
 
+pub const clock_frequencies = .{
+    .cpu = 16_000_000,
+};
+
 pub fn parsePin(comptime spec: []const u8) type {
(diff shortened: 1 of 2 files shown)


PhilippWendel:
It might make more sense to add the arduino uno as a board, move the clock definition over there and use that for the uart example.

MasterQ32:
Yes, please. The Atmega328p has no builtin clock or LED pin. Just create a new board next to the Arduino Nano :+1:

mattnite:
yup @PhilippWendel get that in and it'll be good to merge

PhilippWendel:
Moved to https://github.com/ZigEmbeddedGroup/microzig/pull/89

PR #95, opened 2022-12-23
stm32f4xx: add systick register

### LINKED ISSUE #94: STM32F4XX does not support SysTick register
I think this may be a bug in regz, not microzig, or maybe whatever svd file that was used to generate the register definitions. afaik All Cortex devices have the same definition of this register. I coppied a samd one and it worked without issue

ConnorRigby:
Fixes #94 

I didn't use regz to generate this, so it might not be suitable for  merging, just wanted a trail to show how i solved the issue


mattnite:
This is sort of working with Regz (some bugs need fixing)

PR #99, opened 2023-01-15
Use `root.std_options` to override log behaviour

aoldershaw:
ziglang/zig#14181 changed the interface for overriding the log function. Previously, you would define a public `root.log` function. Now, you define a `root.std_options` namespace that has a `logFn` decl.


MasterQ32:
Thanks!

mattnite:
:pray:

PR #101, opened 2023-02-04
Catchup to changes introduced by ziglang/zig#14498

PhilippWendel:
Catchup to changes introduced by https://github.com/ziglang/zig/pull/14498

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -10,7 +10,7 @@ const chips = microzig.chips;
 const Backing = microzig.Backing;
 
 pub fn build(b: *std.build.Builder) !void {
-    const mode = b.standardReleaseOptions();
+    const optimize = b.standardOptimizeOption(.{});
 
     const test_step = b.step("test", "Builds and runs the library test suite");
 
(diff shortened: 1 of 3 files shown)


build.zig
```diff
@@ -56,7 +56,7 @@ pub fn build(b: *std.build.Builder) !void {
             );
 
             if (filter == null or exe.inner.target.cpu_arch.? == filter.?) {
-                exe.inner.setBuildMode(mode);
+                exe.inner.optimize = optimize;
```
mattnite, line 59: Could you instead add `optimize` to the build options for `addEmbeddedExecutable()`? And default it to debug.

src/main.zig
```diff
@@ -186,7 +186,7 @@ pub fn addEmbeddedExecutable(
     // might not be true for all machines (Pi Pico), but
     // for the HAL it's true (it doesn't know the concept of threading)
     exe.inner.single_threaded = true;
-    exe.inner.setTarget(chip.cpu.target);
+    exe.inner.target = chip.cpu.target;
```
mattnite, line 189: Set this in the invocation of `addExecutable()`

mattnite (changes requested):

mattnite:
Thanks for the updates! Just a couple comments and we're gtg. Not sure why the windows build is failing but I'm not going to let it block merging

build.zig
mattnite: Oh you can nuke it np

mattnite:
Lol I think Pkg changed to Module, we'll leave that for another PR

mattnite (approved):

PR #120, opened 2023-04-02
Updated parts of gpio and pin to reflect code changes from regz rewrite

PhilippWendel:
I have also updated the readme build example as that has changed as well.


mattnite:
🙏

PR #124, opened 2023-04-19
Abstract USB device implementation

r4gus:
I've extracted the platform independent code from the rp2040 device impl (see ZigEmbeddedGroup/raspberrypi-rp2040#40).

One can create a concrete implementation by calling `Usb(F)`, where `F` is a struct that contains a set of functions to handle the low level details (register access etc.).

Check out the usb examples in the [rp2040 repo](https://github.com/r4gus/raspberrypi-rp2040/blob/main/examples/usb_hid.zig)!

@MasterQ32 @mattnite

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -270,8 +270,16 @@ pub fn build(b: *std.build.Builder) !void {
         .optimize = optimize,
     });
 
+    const core_tests = b.addTest(.{
+        .root_source_file = .{
+            .path = comptime root_dir() ++ "/src/core.zig",
+        },
+        .optimize = optimize,
+    });
+
     const test_step = b.step("test", "build test programs");
     test_step.dependOn(&minimal.inner.step);
     test_step.dependOn(&has_hal.inner.step);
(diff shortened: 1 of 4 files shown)


mattnite:
🤌

mattnite:
In general I would prefer removing  `Usb` from the beginning of the identifiers since they're in a `usb` namespace. That being said, I'm also for keeping them as is since I know that's how the USB standard refers to them. What do you think?

r4gus:
Just write everything down you dislike, I'll take care of it

mattnite:
The only other change I'd like to see is to ensure that the unit tests run when you execute `zig build test`, once that's done with the removal of `Usb` we're good to merge :)

mattnite:
@r4gus is this ready to merge?

r4gus:
@mattnite yes. Sending multiple packets between Host and Client seems to Work fine (with the rp2040). Adding Support for other Chips will probably reveal some shortcomings.

mattnite:
Awesome

PR #128, opened 2023-05-01
update to zig master: use @memset, @memcpy builtin

binarycraft007:
with PR: [15278](https://github.com/ziglang/zig/pull/15278) merged, zig master now uses @memcpy and @memset builtin instead of mem.copy, mem.set, this MR update all usage of mem.copy and mem.set to use builtins.


binarycraft007:
I don't think the ci is using the latest zig:
```
$ trap 'kill -- $' INT TERM QUIT; zig version
--
0.11.0-dev.2892+fd6200eda
```

mattnite:
🙏

PR #142, opened 2023-08-20
Regz

mattnite:


MasterQ32:
We could either merge this "for now" or apply the changes i proposed on the ZEG server:
```zig
mz = microzig.init(b);
mz.addExecutable(...);
```

and ship a self-initializing microzig that will use the provided builder to fetch deps

mattnite:
I like that approach much better, closing this.

PR #144, opened 2023-09-21
Microzig Generation 2 Build Interface 

MasterQ32:


mattnite:
This is amazing 🤌. Only comment I have is to remove the optimize argument in github actions now that the build doesn’t take it.

PR #147, opened 2023-09-23
Implements some things to use avr-gcc for AVR5 target.

MasterQ32:


mattnite:
@ikskuh this is super old, could you update it if these changes are needed?

ikskuh:
I guess we can close it and rework the stuff later. It's kinda "obvious" what to do as soon as you try to emit C code and considering my spare time right now i'd rather close the PR than updating it

PR #151, opened 2023-10-26
Change the HAL API to snake_case() from camelCase()

marnix:
This is to follow the STM32 HAL implementations in stmicro-stm32, which were converted to snake_case() when they were extracted from the main microzig repository into their own separate repository.


mattnite:
Thank you!

PR #153, opened 2023-11-06
Work around ziglang/zig #17882

marnix:
As discussed on Discord with @mattnite, working around (suspected) Zig 0.11.0 bug https://github.com/ziglang/zig/issues/17882 in microzig.

(Issue #154 created to revert this in the next Zig release that fixes this.)


ikskuh:
I guess we can close this as we're on 0.13 now?

PR #157, opened 2023-11-13
Fix mmio toggle

akiroz:
In zig 0.11.0, struct field `default_value` always returns an opaque pointer which can't be used in `@tagName`.
I'm not entirely sure how the original API worked but this implementation allows selective toggling of bits in each field of registers via a xor mask:
```
REG.toggle(.{ .FIELD = MASK });
```

### DIFF
--- a/core/src/mmio.zig
+++ b/core/src/mmio.zig
@@ -47,7 +47,7 @@ pub fn Mmio(comptime PackedT: type) type {
         pub inline fn toggle(addr: *volatile Self, fields: anytype) void {
             var val = read(addr);
             inline for (@typeInfo(@TypeOf(fields)).Struct.fields) |field| {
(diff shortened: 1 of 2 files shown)


ikskuh:
This function changes the semantics of the original function.

The idea was that usage looked like thiS:
```zig
REG.toggle(.{ .pin0, .pin5, .pin7 });
```
which would be equivalent to
```zig
var inter = REG.read();
inter.pin0 = !inter.pin0;
inter.pin5 = !inter.pin5;
inter.pin7 = !inter.pin7;
RED.write(inter);
```

I think we could use your function as well, we never really used `toggle` i guess.

akiroz:
The original way it's written likely no longer works with the current version of zig, I think it can be useful to have a xor mask as well since fields can be more than 1 bit and you may want to flip a particular bit in a field.

PR #165, opened 2024-02-17
fix documentation for patch tool

tobiaskohlbau:
The patch tool expects the build.zig.zon file as
it's argument not the project folder.


mattnite:
Thanks!

PR #166, opened 2024-02-17
support bundling only specific boards

tobiaskohlbau:
Support bundling only specified boards and skipping the examples. This speedsup the process in case
the developer is only working on specific boards
at a time. By default every example and every board are bundled.


mattnite:
Thanks!

PR #168, opened 2024-02-17
fix ci by cloning complete history

tobiaskohlbau:


tobiaskohlbau:
Changes incoperated into #164

PR #174, opened 2024-02-20
rp2040: Add support for PIO's jmp_pin

Grazfather:
This field is a bit tricky: It belongs to the EXECCTRL register, while most of the other fields can determined based on the directives used in the assembled program, this one has to be set explicitly. This makes it so that we have to plumb some way to explicitly set the field. I did this by making `LoadAndStartProgramOptions` take `ExecOptions`, but this isn't perfect, because if a user sets other fields, they will be ignored in favour of the values set in the program.

### DIFF
--- a/board-support/raspberrypi-rp2040/src/hal/pio.zig
+++ b/board-support/raspberrypi-rp2040/src/hal/pio.zig
@@ -83,6 +83,7 @@ pub const ClkDivOptions = struct {
 };
 
 pub const ExecOptions = struct {
+    jmp_pin: u5 = 0,
     wrap: u5 = 31,
     wrap_target: u5 = 0,
     side_pindir: bool = false,
@@ -132,6 +133,7 @@ pub const LoadAndStartProgramOptions = struct {
     clkdiv: ClkDivOptions,
     shift: ShiftOptions = .{},
     pin_mappings: PinMappingOptions = .{},
+    exec: ExecOptions = .{},
 };
 
 pub const Pio = enum(u1) {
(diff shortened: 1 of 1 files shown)


Grazfather:
Tested with a simple pio program:

```
        \\.program blink
        \\
        \\mainloop:
        \\    jmp pin mainloop
        \\    set y, 31
        \\    set pins 0 [31]
        \\delay_off:
        \\    jmp y--, delay_off
        \\    set y, 31
        \\    set pins, 1 [31]
        \\delay_on:
        \\    jmp y--, delay_on
        \\    jmp mainloop
```

This simply blinks the built-in LED unless the jmp_pin is high.

Grazfather:
This is what the fix should look like, I just don't get how to get it to compile (It complains because the config value isn't available at comptime)

```diff
diff --git a/board-support/raspberrypi-rp2040/src/hal/dma.zig b/board-support/raspberrypi-rp2040/src/hal/dma.zig
index fcbc3d8..b5eb749 100644
--- a/board-support/raspberrypi-rp2040/src/hal/dma.zig
+++ b/board-support/raspberrypi-rp2040/src/hal/dma.zig
@@ -107,7 +107,12 @@ pub const Channel = enum(u4) {
         regs.ctrl_trig.modify(.{
             .EN = @intFromBool(config.enable),
             .DATA_SIZE = .{
-                .value = .SIZE_BYTE,
+                .value = switch (config.transfer_size_bytes) {
+                    1 => .SIZE_BYTE,
+                    2 => .SIZE_HALFWORD,
+                    3 => .SIZE_WORD,
+                    else => unreachable,
+                },
             },
             .INCR_READ = @intFromBool(config.read_increment),
             .INCR_WRITE = @intFromBool(config.write_increment),
```

```
src/hal/dma.zig:110:26: error: value with comptime-only type '@TypeOf(.enum_literal)' depends on runtime control flo
  w
                  .value = switch (config.transfer_size_bytes) {
```

PR #177, opened 2024-02-25
usb: unify handling of length parameters

tobiaskohlbau:
The length parameters in USB packets are strict and the existing implementation does already hardcode the length within the serialize functions. Therefore the user should not provide these lengths during creation time. In order to make this more verbose use defaults values within the structs and reference these defaults values within the serialization functions. It's not possible to remove these fields completly as this fields are used within deserialization from the hardware itself.

### DIFF
--- a/core/src/core/usb.zig
+++ b/core/src/core/usb.zig
@@ -513,7 +513,7 @@ pub const Dir = enum(u8) {
 /// Describes an endpoint within an interface
 pub const EndpointDescriptor = extern struct {
     /// Length of this struct, must be 7.
-    length: u8,
+    length: u8 = 7,
     /// Type of this descriptor, must be `Endpoint`.
     descriptor_type: DescType,
     /// Address of this endpoint, where the bottom 4 bits give the endpoint
@@ -530,7 +530,7 @@ pub const EndpointDescriptor = extern struct {
 
     pub fn serialize(self: *const @This()) [7]u8 {
         var out: [7]u8 = undefined;
-        out[0] = 7; // length
+        out[0] = self.length;
         out[1] = @intFromEnum(self.descriptor_type);
         out[2] = self.endpoint_address;
         out[3] = self.attributes;
@@ -544,7 +544,7 @@ pub const EndpointDescriptor = extern struct {
 /// Description of an interface within a configuration.
 pub const InterfaceDescriptor = extern struct {
     /// Length of this structure, must be 9.
-    length: u8,
+    length: u8 = 9,
     /// Type of this descriptor, must be `Interface`.
     descriptor_type: DescType,
     /// ID of this interface.
@@ -566,7 +566,7 @@ pub const InterfaceDescriptor = extern struct {
 
     pub fn serialize(self: *const @This()) [9]u8 {
         var out: [9]u8 = undefined;
-        out[0] = 9; // length
+        out[0] = self.length;
         out[1] = @intFromEnum(self.descriptor_type);
         out[2] = self.interface_number;
         out[3] = self.alternate_setting;
@@ -582,7 +582,7 @@ pub const InterfaceDescriptor = extern struct {
 /// Description of a single available device configuration.
 pub const ConfigurationDescriptor = extern struct {
     /// Length of this structure, must be 9.
-    length: u8,
+    length: u8 = 9,
     /// Type of this descriptor, must be `Config`.
     descriptor_type: DescType,
     /// Total length of all descriptors in this configuration, concatenated.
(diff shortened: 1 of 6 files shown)


core/src/core/usb.zig
```diff
@@ -513,7 +513,7 @@ pub const Dir = enum(u8) {
 /// Describes an endpoint within an interface
 pub const EndpointDescriptor = extern struct {
     /// Length of this struct, must be 7.
-    length: u8,
+    length: u8 = 7,
```
MasterQ32, line 516: I guess this could be replaced with a `@sizeOf(@This())`, i'd probably create a helper function like `sizeOfCheck`:

```zig
fn sizeOfCheck(comptime T: type, comptime expected: comptime_int) comptime_int {
    const actual = @sizeOf(T);
    if(actual != expected) {
        @compileError(std.fmt.comptimePrint("{s} was expected to be {d} bytes large, but actually has {d} bytes size.", .{
            @typeName(T), expected, actual,
        });
    }
    return actual;
}
```

This way, you hardcode the size of the field, but you also validate that it is actually `7`.

core/src/core/usb.zig
```diff
@@ -530,7 +530,7 @@ pub const EndpointDescriptor = extern struct {
 
     pub fn serialize(self: *const @This()) [7]u8 {
         var out: [7]u8 = undefined;
-        out[0] = 7; // length
+        out[0] = self.length;
```
MasterQ32, line 533: This function should assert `self.length` to be 7, and hardcode it with either `@sizeOf(@This())` or use `out.len`.

Otherwise, a user could pass in a bad `self` which has length = 6 or similar.

MasterQ32 (changes requested):
The two comments are relevant for all changes

core/src/core/usb.zig
tobiaskohlbau: This would be possible indeed, thought about that too.

core/src/core/usb.zig
tobiaskohlbau: That a pretty good idea, will implement it.

core/src/core/usb.zig
tobiaskohlbau: I've decided to remove the length fields altogether, as they're only used for calculating the length of the descriptors and the calculation is indeed wrong. An extern fields are padded and therefore do not always result in the correct size returned by sizeOf. For e.g. the example EndpointDescriptor is reported as 10 bytes even though the acutal size is 9. I've updated the examples to show how the _magicnumber_ is calculated and we should strive to hide the manual calculation from the user. Will do this in a followup PR.

tobiaskohlbau:
Looks like the CI needs a retry, maybe GitHub still has issues?

> Error: Error: server was started but never notified its presence.

https://github.com/ZigEmbeddedGroup/microzig/actions/runs/8054829310/job/22000295495?pr=177

core/src/core/usb.zig
```diff
@@ -610,7 +604,7 @@ pub const ConfigurationDescriptor = extern struct {
 
     pub fn serialize(self: *const @This()) [9]u8 {
         var out: [9]u8 = undefined;
-        out[0] = 9; // length
+        out[0] = out.len;
         out[1] = @intFromEnum(self.descriptor_type);
         out[2] = @intCast(self.total_length & 0xff);
         out[3] = @intCast((self.total_length >> 8) & 0xff);
```
MasterQ32, line 610: There's `std.mem.writeInt` which can be used to write big/little endian integers of bigger sizes into arrays

examples/raspberrypi-rp2040/src/usb_device.zig
```diff
@@ -101,9 +97,10 @@ pub var DEVICE_CONFIGURATION: usb.DeviceConfiguration = .{
         .interface_s = 0,
     },
     .config_descriptor = &.{
-        .length = @as(u8, @intCast(@sizeOf(usb.ConfigurationDescriptor))),
         .descriptor_type = usb.DescType.Config,
-        .total_length = @as(u8, @intCast(@sizeOf(usb.ConfigurationDescriptor) + @sizeOf(usb.InterfaceDescriptor) + @sizeOf(usb.EndpointDescriptor) + @sizeOf(usb.EndpointDescriptor))),
+        // This is calculated via the sizes of underlying descriptors contained in this configuration.
+        // ConfigurationDescriptor(9) + InterfaceDescriptor(9) * 1 + EndpointDescriptor(8) * 2
+        .total_length = 34,
```
MasterQ32, line 103: I wonder if we can compute this value of the configuration. can we actually ever have more than a single config descriptor active? cc @r4gus

MasterQ32 (changes requested):

examples/raspberrypi-rp2040/src/usb_device.zig
tobiaskohlbau: Yes this could be calculated, as stated above I'm tinkering with a solution to resolve this. But this shouldn't be in this CL it's a followup.

Issue #194, opened 2024-05-16
Support for ATmega32U4


biom4st3r:
I've been working on a Keyboard for a while based on the ATmega32U4 and would love to someday have the firmware written in Zig.
@MasterQ32 
I have:
* Double checked all the addresses from the 328p matched the 32u4 datasheet
* bumped the ram from 2kb -> 2.5kb
* Changed the USART0 -> USART1 per the json
* Added ATmega32u4.json and manually patched the `COMM_SCK_RATE_3BIT` enum, because it was missing a `size` field
* Add the Adafruit itsy bitsy 32u4 board
https://github.com/biom4st3r/microzig

I cannot test until the dependencies are updated to latest zig

PR #195, opened 2024-05-16
Basic support for ATmega32u4

biom4st3r:
#194


MasterQ32:
Thanks! I hope i find time soon!

ikskuh:
Where did you get the `json` file? Merging now, but could you do a follow up MR if possible when you find a ATDF for the register definitions and commit that instead of the json?

PR #199, opened 2024-05-20
Properly export startup logic for espressif-riscv

DNedic:
This fixes an issue where startup logic was not properly exported for the espressif-riscv CPU, leading to failed builds at first and lack of direct boot bytes on second and further builds.

The issue can be reproduced by building the `espressif/esp` example with Zig `0.12`.

First output:
```bash
install
└─ install generated to esp32-c3_blinky.bin
   └─ objcopy generated
      └─ zig build-exe esp32-c3_blinky Debug riscv32-freestanding-eabi failure
error: warning(link): unexpected LLD stderr:
ld.lld: warning: cannot find entry symbol _start; not setting start address
```

### DIFF
--- a/bsp/espressif/esp/src/cpus/espressif-riscv.zig
+++ b/bsp/espressif/esp/src/cpus/espressif-riscv.zig
@@ -61,7 +61,7 @@ pub const startup_logic = struct {
 
     extern fn microzig_main() noreturn;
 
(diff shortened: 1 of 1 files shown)


DNedic:
Note: There is currently another issue I am currently resolving relating to the `.sdata` section being created, so ignore the first part of the binary and the following zeroes.

PR #200, opened 2024-05-20
Linking and linker relaxation fixes

DNedic:
This PR fixes the following issues:
1. Accounts for the possibility of the `.sdata*` sections being generated. These sections are created by compilers in order to optimize access for small data on some architectures. Right now this breaks the `espressif/esp` example the `.sdata` output section is being automatically placed at the wrong place, creating an incorrect and enormous binary.
2. Properly provide the `__global_pointer$` used in the RISC-V architecture CPUs to initialize the GP register. This register is then used in order to improve access times to frequently used data, as long as it is +-2048 bytes away by fitting the offset in the immideate part of the instruction.
3. Disables relaxation when initializing the GP register on `espressif-riscv`

References:
https://www.sifive.com/blog/all-aboard-part-3-linker-relaxation-in-riscv-toolchain
https://five-embeddev.com/quickref/global_pointer.html
https://d3s.mff.cuni.cz/files/teaching/nswi200/202324/doc/riscv-abi.pdf

### DIFF
--- a/bsp/espressif/esp/src/cpus/espressif-riscv.zig
+++ b/bsp/espressif/esp/src/cpus/espressif-riscv.zig
@@ -67,7 +67,12 @@ pub const startup_logic = struct {
             :
             : [eos] "r" (@as(u32, microzig.config.end_of_stack)),
         );
-        asm volatile ("la gp, __global_pointer$");
+        asm volatile (
+            \\.option push
+            \\.option norelax
+            \\la gp, __global_pointer$
+            \\.option pop
+        );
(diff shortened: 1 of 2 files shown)


DNedic:
The toolchain does not appear to generate gp relative loads still, does zig utilize this feature?

DNedic:
It does appear that relaxation is not enabled in zig: https://github.com/ziglang/zig/issues/3451, however we should still merge this as `espressif/esp` example (and other projects utilizing `espressif-riscv`) are broken.

mattnite:
@DNedic thank you!

PR #201, opened 2024-06-02
USB driver: support chunked data

arkadiuszwojcik:
This PR adds support for chunked data in the USB driver. This is especially required for configuration sections larger than the endpoint's maximum packet size.

### DIFF
--- a/core/src/core/usb.zig
+++ b/core/src/core/usb.zig
@@ -94,10 +94,10 @@ pub fn Usb(comptime f: anytype) type {
             const CmdEndpoint = struct {
 
                 /// Command response utility function that can split long data in multiple packets
-                fn send_cmd_response(data: []const u8, expected_length: u16) void {
+                fn send_cmd_response(data: []const u8, expected_max_length: u16) void {
                     const cmd_in_endpoint = usb_config.?.endpoints[EP0_IN_IDX];
 
-                    S.buffer_reader = BufferReader { .buffer = data[0..@min(data.len, expected_length)] };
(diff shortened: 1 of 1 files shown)


r4gus:
I'll look at this as soon as I got time (probably the weekend).

Did you test this somehow? If you have used any script etc. for testing, it would be nice if you attach them to this thread.

arkadiuszwojcik:
I will try to put something on weekend. I have draft code for RP2040 that uses USB (CEC ADC) config so windows detect it as serial port, it required quite long config descriptor so it have to be splited in 2 packets.

r4gus:
If you have a working implementation you could add a minimal working project to the examples.

arkadiuszwojcik:
Such example require some futher changes in current API but I created for you a [branch](https://github.com/arkadiuszwojcik/microzig/tree/usb_chunked_data_cdc_example) with working CDC example above changes introduced by this PR for testing purpose. By working example I mean point where serial port is discovered by host. There is hovever one issue, this handshake process is very slow (up to 15s?) I don't think this is related to this PR but to CDC config from my example. I should take a look on that today or tomorrow.

arkadiuszwojcik:
@r4gus I just found issue responsible for that bug with long delay. My code didn't send fianal empty packet after command response. PR and example were updated. Now Serial detection from CDC example branch work instantly. This PR introduce only chunked data, CDC support require some more discussions.

r4gus (approved):

r4gus:
@ikskuh

ikskuh:
Wonderful, nice work!

PR #203, opened 2024-06-08
Complete Gd32f103 Support

asterix24:
I am new of zig, but I have developed in embedded RTOS.

I would to learn to use the zig language programming, starting from the longan clone gd32f103 board, that have on hand. 
In this pull request I start to implement the gpio driver take example from the stmicro support.

I do the blinky example and works, but I would ask you if I am correct and the code architecture could be good.

Now I am continuing with the uart.

Let me know what do think.


ikskuh:
Looks good to me!

PR #206, opened 2024-06-12
USB driver: new config API and CDC support example

arkadiuszwojcik:
This PR proposes the following changes:

1. Namespaces rearrangement in USB driver
2. New USB configuration API with descriptor templates (this simpify usb driver examples)
3. New USB example code with CDC support


arkadiuszwojcik:
This require some more work and discussion

PR #216, opened 2024-07-09
Initial ESP32-C3 GPIO HAL driver

DNedic:
This adds a first iteration of the `ESP32-C3` (later to be universal) GPIO HAL driver, and factors it out into its own file.

### DIFF
--- a/bsp/espressif/esp/src/hals/esp32-c3/gpio.zig
+++ b/bsp/espressif/esp/src/hals/esp32-c3/gpio.zig
@@ -30,10 +30,8 @@ pub const Pin = struct {
         drive_strength: DriveStrength = DriveStrength.@"5mA",
     };
 
-    pub fn init(number: u5, config: Config) Pin {
-        std.debug.assert(number < 21);
-
-        IO_MUX.GPIO[number].modify(.{
+    pub fn apply(self: Pin, config: Config) void {
+        IO_MUX.GPIO[self.number].modify(.{
             .SLP_SEL = 0,
             .FUN_WPD = @intFromBool(config.pulldown_enable),
             .FUN_WPU = @intFromBool(config.pullup_enable),
@@ -43,7 +41,7 @@ pub const Pin = struct {
             .MCU_SEL = 1,
         });
 
-        GPIO.FUNC_OUT_SEL_CFG[number].write(.{
+        GPIO.FUNC_OUT_SEL_CFG[self.number].write(.{
             .OUT_SEL = 0x80,
             .OEN_SEL = @intFromBool(config.output_enable),
             .INV_SEL = @intFromBool(config.output_invert),
@@ -51,25 +49,25 @@ pub const Pin = struct {
             .padding = 0,
         });
 
-        GPIO.PIN[number].modify(.{
+        GPIO.PIN[self.number].modify(.{
             .PIN_PAD_DRIVER = @intFromBool(config.open_drain),
         });
 
         // NOTE: Assert that the USB_SERIAL_JTAG peripheral which uses pins GPIO18 and GPIO19
         //       is disabled and USB pullup/down resistors are disabled.
-        if (number == 18 or number == 19) {
+        if (self.number == 18 or self.number == 19) {
             const usb_conf0 = peripherals.USB_DEVICE.CONF0.read();
-            std.debug.assert(usb_conf0.USB_PAD_ENABLE == 0 and usb_conf0.DP_PULLUP == 0 and usb_conf0.DP_PULLDOWN == 0 and usb_conf0.DM_PULLUP == 0 and usb_conf0.DM_PULLDOWN == 0);
+            std.debug.assert(usb_conf0.USB_PAD_ENABLE == 0 and
+                usb_conf0.DP_PULLUP == 0 and
+                usb_conf0.DP_PULLDOWN == 0 and
+                usb_conf0.DM_PULLUP == 0 and
+                usb_conf0.DM_PULLDOWN == 0);
         }
 
-        GPIO.ENABLE_W1TS.write(.{ .ENABLE_W1TS = @as(u17, 1) << number, .padding = 0 });
-
-        return Pin{
-            .number = number,
-        };
+        GPIO.ENABLE_W1TS.write(.{ .ENABLE_W1TS = @as(u17, 1) << self.number, .padding = 0 });
     }
 
-    pub fn deinit(self: Pin) void {
+    pub fn reset(self: Pin) void {
         IO_MUX.GPIO[self.number].write_raw(0x00);
         GPIO.FUNC_OUT_SEL_CFG[self.number].write_raw(0x00);
         GPIO.PIN[self.number].write_raw(0x00);
@@ -144,5 +142,34 @@ pub const Pin = struct {
             Level.low => write(self, Level.high),
             Level.high => write(self, Level.low),
         }
+    }
+};
+
+pub const instance = struct {
+    pub const GPIO0 = Pin{ .number = 0 };
+    pub const GPIO1 = Pin{ .number = 1 };
+    pub const GPIO2 = Pin{ .number = 2 };
+    pub const GPIO3 = Pin{ .number = 3 };
+    pub const GPIO4 = Pin{ .number = 4 };
+    pub const GPIO5 = Pin{ .number = 5 };
+    pub const GPIO6 = Pin{ .number = 6 };
+    pub const GPIO7 = Pin{ .number = 7 };
+    pub const GPIO8 = Pin{ .number = 8 };
+    pub const GPIO9 = Pin{ .number = 9 };
+    pub const GPIO10 = Pin{ .number = 10 };
+    pub const GPIO11 = Pin{ .number = 11 };
+    pub const GPIO12 = Pin{ .number = 12 };
+    pub const GPIO13 = Pin{ .number = 13 };
+    pub const GPIO14 = Pin{ .number = 14 };
+    pub const GPIO15 = Pin{ .number = 15 };
+    pub const GPIO16 = Pin{ .number = 16 };
+    pub const GPIO17 = Pin{ .number = 17 };
+    pub const GPIO18 = Pin{ .number = 18 };
+    pub const GPIO19 = Pin{ .number = 19 };
+    pub const GPIO20 = Pin{ .number = 20 };
+
+    pub fn num(instance_number: u5) Pin {
+        std.debug.assert(instance_number < 21);
+        return Pin{ .number = instance_number };
     }
 };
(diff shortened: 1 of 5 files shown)


DNedic:
I will be changing the approach to the API, there is simply too much variation in possible configurations to have everything as a typestate, and having `Output` and `Input` separately when there will still be runtime differences is a half-measure.

bsp/espressif/esp/src/hals/esp32-c3/gpio.zig
```diff
@@ -0,0 +1,148 @@
...
+
+pub const DriveStrength = enum(u2) {
+    @"5mA" = 0,
+    @"10mA" = 1,
+    @"20mA" = 2,
+    @"40mA" = 3,
+};
+
+pub const Pin = struct {
+    number: u5,
```
haydenridd, line 20: Since the only field in this struct is a `u5`, you can actually just turn this into an `enum(u5)` instead of a struct! Then, any call to `self.number` can become `@intFromEnum(self)`.

bsp/espressif/esp/src/hals/esp32-c3/gpio.zig
```diff
@@ -0,0 +1,148 @@
...
+        output_enable: bool = false,
+        open_drain: bool = false,
+        pullup_enable: bool = false,
+        pulldown_enable: bool = false,
+        input_filter_enable: bool = false,
+        output_invert: bool = false,
+        drive_strength: DriveStrength = DriveStrength.@"5mA",
+    };
+
+    pub fn init(number: u5, config: Config) Pin {
```
haydenridd, line 33: I would split this out into `from_instance_number(number: u5) Pin` and then an `init(self: Pin) void`.

Also I'm not committed to this, but at least the current standard terminology in MicroZig for these types of functions would be `apply()` and `reset()` (instead of `deinit()`):
https://github.com/ZigEmbeddedGroup/.github/blob/main/GUIDELINES.md#terminology

bsp/espressif/esp/src/hals/esp32-c3/gpio.zig
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    pub fn set_input_filter(self: Pin, enable: bool) void {
+        IO_MUX.GPIO[self.number].modify(.{
+            .FILTER_EN = @intFromBool(enable),
+        });
+    }
+
+    pub fn write(self: Pin, level: Level) void {
+        std.debug.assert(GPIO.FUNC_OUT_SEL_CFG[self.number].read().OEN_SEL == 1);
```
haydenridd, line 122: IMO I don't think this assert is required and could potentially hurt flexibility. Maybe someone for some reason wants to "pre-set" their output register before enabling the GPIO via `OEN_SEL`?

bsp/espressif/esp/src/hals/esp32-c3/gpio.zig
```diff
@@ -0,0 +1,148 @@
...
+
+        switch (level) {
+            Level.low => GPIO.OUT_W1TC.write(.{ .OUT_W1TC = @as(u17, 1) << self.number, .padding = 0 }),
+            Level.high => GPIO.OUT_W1TS.write(.{ .OUT_W1TS = @as(u17, 1) << self.number, .padding = 0 }),
+        }
+    }
+
+    pub fn get_output_state(self: Pin) Level {
+        std.debug.assert(GPIO.FUNC_OUT_SEL_CFG[self.number].read().OEN_SEL == 1);
```
haydenridd, line 131: Similar sentiment to previous comment, a lot of these asserts I think can just be deleted and it's up to the user to correctly configure the pin!

DNedic:
Thanks for the review @haydenridd !

bsp/espressif/esp/src/hals/esp32-c3/gpio.zig
DNedic: I would keep this in, as forgetting to enable something before using it is so extremely common that not asserting here would probably waste too much of people's time.

Generally, yes I agree that only invariants should be asserted and all other errors should be returned for the user to handle, but I believe very few people expect `pin.write(Level.high)` to be fallible.

DNedic: Alternative for setting initial states: Make `Config` also contain optional `initial_output`, and have `set_output_enable()` also take that parameter?

bsp/espressif/esp/src/hals/esp32-c3/gpio.zig
DNedic: True, but then that would create bigger changes down the line if we want to add things to the struct. I will definitely think about this though!

haydenridd: For sure, personal preference!

bsp/espressif/esp/src/hals/esp32-c3/gpio.zig
haydenridd: Yup agreed, I think in this case it's cool to leave in.

ikskuh (approved):
Looks good so far!

haydenridd:
Only comment would be to humbly consider this approach to creating "instances" of a `Pin` for starting to create some consistency between vendor HALs:
https://github.com/haydenridd/microzig/blob/0a4f51774adb0bcba7b6770f7cbc930dbd39b940/bsp/raspberrypi/rp2040/src/hal/i2c.zig#L171-L177

But... We absolutely don't have a standard for this yet so up to you :) Looks good!

haydenridd (approved):
Only comment would be to humbly consider this approach to creating "instances" of a Pin for starting to create some consistency between vendor HALs:
https://github.com/haydenridd/microzig/blob/0a4f51774adb0bcba7b6770f7cbc930dbd39b940/bsp/raspberrypi/rp2040/src/hal/i2c.zig#L171-L177

But... We absolutely don't have a standard for this yet so up to you :) Looks good!

DNedic:
Thanks for reviews @haydenridd @ikskuh ! 

Could you explain what benefits the approach taken in the Raspberry Pi Pico HAL provides in this case?

haydenridd:
> Thanks for reviews @haydenridd @ikskuh !
> 
> Could you explain what benefits the approach taken in the Raspberry Pi Pico HAL provides in this case?

For sure! Also just for clarity, the example I gave was general to any peripheral, not just a "Pin". The benefits I see are as follows:
- Given there will be a variety of types/functions name-spaced under `hal.i2c`, for example `I2C`, `TransactionError`, `TimingRegisterValues`, etc, it's nice that there is a standardized namespace _within_ `hal.i2c` to "get a given instance of a peripheral"
    - No matter the peripheral, there is a consistent experience of _accessing_ said peripheral via `hal.peripheral.instance`
- In the general case, a user would access a peripheral via `hal.i2c.instance.I2C0/1`
- If for some reason a user needs to programatically access the instance via number, that's where `hal.i2c.instance.num(0)` comes in. Use cases for this will be fewer, but adds convenience if you want to configure all instances in a loop.
- Finally, if there is a peripheral where there is only ever a single instance (ADC on the RP2040), its methods would just be directly accessed via `hal.i2c.instance.method()`

So, essentially, this approach is just for consistency and convenience. It also attempts to provide some intuition that peripherals are not something that can be created or destroyed, but rather are inherent global entities that preserve state throughout program execution.

DNedic:
Thanks for the clarification @haydenridd , this makes sense to me. I have some conflicts of taste with the approach though, though nothing deal-breaking.

* Even though we cannot 'create' peripherals, we can copy the instance like so: `led_pin = gpio.instance.GPIO8`, multiple times as well, so in a sense this does not seem too different from `led_pin = gpio.Pin.init(8)`
* Besides this, by using `init()`, we are ensuring that the peripheral is always initialized before attempting to use it, while using `apply` to initialize in the other case can be easily forgotten
* (nitpick) The final convenience feature, single instance case `hal.i2c.instance.method()` is a bit of an outlier and might confuse people if they are used to using `instance.instance`, can we have global instances without the `instance` namespace and save some typing?
* (nitpick, does not conflict with the approach mentioned in a fundamental way) Using enums as underlying types and then `@enumfromint()` is less verbose while I feel it is more confusing to newcomers and less readable overall than using single-member structures

@ikskuh cc

DNedic:
> The only thing I'll caution is if `init()` configures the peripheral _and_ passes you an instance, does it handle this peripheral being previously configured? As in, does it know/care about this sort of behavior:

Good point there, a check could be added to peripherals where double initialization would for some reason cause problems by either adding an `initialized: bool` member, or infering from registers.

Generally, I would let users track this and would not attempt to protect them from a case like this, as the solutions can get unwieldly very quickly. The global instance situation has the same problem, so there is really no difference when it comes to that in the approaches, the only difference is that you can't have an uninitialized peripheral unless you manually deinitialize/reset.

A counter question would be, `apply(config)` has configure semantics, while `init()` has configure + enable  in my opinion (very subjective), so would a separate `enable()` function be considered for `apply()`?

I'm going to adapt my solution to the WIP guidelines we have nevertheless in the meantime and we can merge the changes, however these are some things I'd like to see evaluated further.

DNedic:
@haydenridd PTAL again

DNedic:
Dilemma: Should I use the functions for setting individual options in the constructor to make it more verbose? 

Whether this makes code easier to read is up to discussion (I am leaning towards yes), but there is the issue of register writes being `volatile`, and with that approach more write operations would be created by the compiler, bloating code size. In GPIO this might not be terrible but I imagine for more complex peripherals this can add up.

haydenridd:
> A counter question would be, `apply(config)` has configure semantics, while `init()` has configure + enable in cases where there is no `enable` in `config` in my opinion (very subjective), so would a separate `enable()` function be considered for `apply()`?

A good question, I would say it's a little case dependent. "Enabling" the peripheral can in many cases do "nothing" other than make it ready for use. For instance, enabling the I2C block, but understanding that nothing will happen until you call some sort of `read` or `write` function. In that case it would make sense to enable in `apply`. I agree it's a bit subjective and case by case!

haydenridd (approved):
Looking good!

haydenridd:
> Dilemma: Should I use the functions for setting individual options in the constructor to make it more verbose?
> 
> Whether this makes code easier to read is up to discussion (I am leaning towards yes), but there is the issue of register writes being `volatile`, and with that approach more write operations would be created by the compiler, bloating code size. In GPIO this might not be terrible but I imagine for more complex peripherals this can add up.

I think you can have the best of both worlds if you mark those configuration functions that just touch a register `inline`...

DNedic:
> I think you can have the best of both worlds if you mark those configuration functions that just touch a register `inline`... Then it would be the "same" behavior whether you use those functions or not in `apply`

Are you sure this would merge the `volatile` writes? In C or C++ this would not happen even if they were in the same function, line after line.

haydenridd:
> Are you sure this would merge the `volatile` writes? In C or C++ this would not happen even if they were in the same function, line after line.

Great question... I think so. `inline` works quite differently in Zig than in C/C++, it's a [mandate to the compiler](https://ziglang.org/documentation/master/#inline-fn) rather than a suggestion. Maybe I'm misinterpreting what you mean by "merge" though!

PR #222, opened 2024-07-17
RP2040 SPI HAL Improvements

haydenridd:
Relatively minor improvements, but the highlights:

- Minor renaming for clarity similar to my I2C PR
- Support for all the different frame formats the peripheral supports
- Support for all the different bit widths the peripheral supports
- TX FIFO preloading for performance improvements:

Example transaction before:
RigolDS1

Example transaction with preloading (no gap!):
RigolDS0

 - Removed configuration of GPIO pins in peripheral driver
     - This actually had some immediate benefits, as the way the RP2040 peripheral controls the `CSN` pin automatically is a bit odd, it toggles CSN high after every byte when a GPIO is configured to `.spi` function:
      
RigolDS2

     - For the flash chip I tested this driver on, this would actually make it unusable was I forced to use this
     - Instead, I manually drive CSN via a GPIO, which works perfectly well
     - Not configuring the GPIOs within the peripheral driver gives users of the HAL this kind of flexibility
 
 The full test suite I wrote for this driver can be found here:
 https://github.com/haydenridd/microzig/tree/rp2040-spi


bsp/raspberrypi/rp2040/src/hal/spi.zig
```diff
@@ -12,57 +12,141 @@ const hw = @import("hw.zig");
 
 const SpiRegs = microzig.chip.types.peripherals.SPI0;
 
+pub const Motorola = struct {
+    pub const Polarity = enum(u1) {
+        low = 0,
+        high = 1,
```
ikskuh, line 18: i think `default_low` and `default_high` would be better names

bsp/raspberrypi/rp2040/src/hal/spi.zig
```diff
@@ -12,57 +12,141 @@ const hw = @import("hw.zig");
...
+
 pub const Config = struct {
     clock_config: clocks.GlobalConfiguration,
-    tx_pin: ?gpio.Pin = gpio.num(19),
-    rx_pin: ?gpio.Pin = gpio.num(16),
-    sck_pin: ?gpio.Pin = gpio.num(18),
-    csn_pin: ?gpio.Pin = gpio.num(17),
-    baud_rate: u32 = 1000 * 1000,
+    /// Number of bits in a frame can't be less than 4 or more than 16
+    data_width: u4 = 8,
```
ikskuh, line 42: Use u5 here, as u4 cannot encode `16`

bsp/raspberrypi/rp2040/src/hal/spi.zig
```diff
@@ -12,57 +12,141 @@ const hw = @import("hw.zig");
...
     }
 
-    pub fn apply(spi: SPI, comptime config: Config) void {
+    /// Initializes the SPI HW block per the Config provided
+    /// Per the API limitations discussed above, the following settings are fixed and not configurable:
+    /// - Controller Mode only
+    /// - DREQ signalling is always enabled, harmless if DMA isn't configured to listen for this
+    pub fn apply(spi: SPI, comptime config: Config) ConfigError!void {
+        if (config.data_width < 4) return ConfigError.InvalidDataWidth;
+
```
ikskuh, line 86: Add a check for upper range as well, or we can replace `data_width` with an enum that exposes all legal values

bsp/raspberrypi/rp2040/src/hal/spi.zig
```diff
@@ -72,39 +156,104 @@ pub const SPI = enum(u1) {
...
+            if (std.meta.eql(PacketType, acceptable_type)) return;
+        }
+        @compileError("PacketType must be one of u4, u5, ..., u16");
+    }
+
+    const fifo_depth = 8;
+
+    /// Disable SPI, pre-fill the TX FIFO as much as possible, and then re-enable to start transmission.
+    /// Leads to performance gains in thoroughput. Returns how many bytes were consumed from src.
+    fn prime_tx_fifo(spi: SPI, PacketType: type, src: []const PacketType) usize {
```
ikskuh, line 172: `PacketType` should be comptime

bsp/raspberrypi/rp2040/src/hal/spi.zig
```diff
@@ -72,39 +156,104 @@ pub const SPI = enum(u1) {
     pub inline fn is_readable(spi: SPI) bool {
         return spi.get_regs().SSPSR.read().RNE == 1;
     }
-    pub fn transceive(spi: SPI, src: []const u8, dst: []u8) usize {
+
+    fn validate_bitwidth(PacketType: type) void {
+        const acceptable_types = .{ u4, u5, u6, u7, u8, u9, u10, u11, u12, u13, u14, u15, u16 };
+        for (acceptable_types) |acceptable_type| {
+            if (std.meta.eql(PacketType, acceptable_type)) return;
+        }
```
ikskuh, line 164:
```zig
if(@bitSizeOf(PacketType) < 4 or @bitSizeOf(PacketType) > 16)
  @compileError("PacketType must have between 4 and 16 bits");
```
This allows us to use arbitrary data items instead of forcing uints :)

ikskuh (changes requested):
Looks pretty good so far. Do we need `clock_config` to be `comptime` known. If not, we can remove the `comptime config` requirement on the `apply` fn

bsp/raspberrypi/rp2040/src/hal/spi.zig
haydenridd: Lol great catch, I think I'll just bite the bullet and make this an `enum(u5)`, it's not that many values and it makes it more explicit.

bsp/raspberrypi/rp2040/src/hal/spi.zig
haydenridd: General question, I usually omit `comptime` on a `type` arg because they _have_ to be `comptime` given they're types. I've seen this a decent amount in other Zig code. No issue adding `comptime` to be explicit though!

bsp/raspberrypi/rp2040/src/hal/spi.zig
haydenridd: Ahhh nice ,this is what I was fumbling for when I cooked up this :)

bsp/raspberrypi/rp2040/src/hal/spi.zig
```diff
@@ -12,57 +12,141 @@ const hw = @import("hw.zig");
...
+        first_edge = 0,
+        second_edge = 1,
+    };
+
+    /// Low means clock is steady state low, high means steady state high
+    clock_polarity: Polarity = .low,
+
+    /// Controls whether data is captured on the first or second clock edge transition
+    clock_phase: Phase = .first_edge,
+};
```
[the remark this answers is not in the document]
haydenridd: Done!

haydenridd:
@ikskuh Requested changes applied!

bsp/raspberrypi/rp2040/src/hal/spi.zig
ikskuh: I would add them just to make clear that this parameter is comptime only. It was necessary for a long time, so it might not be "best practise" nowadays, but i still like to see quickly if something has comptime effects

ikskuh (approved):
In general, i'm pretty happy now. Can you maybe add an example where you use a non-u8 type for data width?

haydenridd:
For sure! I'll add an example then merge :)

PR #228, opened 2024-07-30
RP2040 Uart Refactor

haydenridd:
Should start looking pretty familiar to the other PRs!

The highlights:
- Removed configuration of GPIO pins by driver
- Added full transaction error handling
- Consolidated to new "uart.instance.[]" design pattern
- Added new "apply()" and "apply_runtime()" pattern for dual comptime/runtime error checking
- Added querying of current active UART errors
- Tweaked + renamed to write_blocking() and read_blocking() and added a timeout to match other APIs
- Tweaked syntax for using the UART as a logger + added ability to remove uart as logger

### DIFF
--- a/bsp/raspberrypi/rp2040/src/hal.zig
+++ b/bsp/raspberrypi/rp2040/src/hal.zig
@@ -60,4 +60,5 @@ test "hal tests" {
     _ = pio;
     _ = usb;
     _ = i2c;
+    _ = uart;
 }
--- a/bsp/raspberrypi/rp2040/src/hal/uart.zig
+++ b/bsp/raspberrypi/rp2040/src/hal/uart.zig
@@ -42,7 +42,11 @@ pub const Config = struct {
     parity: Parity = .none,
 };
 
-pub const TransactionError = error{
+pub const TransmitError = error{
+    Timeout,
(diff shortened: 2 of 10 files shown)


mattnite (approved):
Love these changes, thank you for all this hard work. Especially like apply vs apply_runtime.

haydenridd:
> Very good work, but what i just asked: Can we even have any transmission error except Timeout during write operations? I mean we can't overrun, can't parity-error, can't break and all of that.
> 
> Apart from that, i'm fine

Yep indeed we can only get a timeout error from a write. Would you suggest separating out into `WriteError`/`ReadError`?

ikskuh:
> Yep indeed we can only get a timeout error from a write. Would you suggest separating out into `WriteError`/`ReadError`?

Yeah, that's the idea. Maybe you could do `TransmitError` and `ReceiveError`?

haydenridd:
> > Yep indeed we can only get a timeout error from a write. Would you suggest separating out into `WriteError`/`ReadError`?
> 
> Yeah, that's the idea. Maybe you could do `TransmitError` and `ReceiveError`?

For sure! I'll update then merge :)

PR #229, opened 2024-08-01
Update website and zine (Zig 0.13) and fix broken links

MatthiasPortzel:
- Updates Zine for Zig 0.13
    - Frontmatter format changed to zon from JSON
    - Requires valid HTML—no self closing tags
- Updated broken links
    - regz and uf2 now point to the monorepo folder
    - Dropped the extension from the "Getting Started" link, that's now fixed
    - The Getting Started page needed some absolute links to assets since it's now a folder not a top-level file
    - Updated the ziglings link to point to Codeberg
- Deleted the "legacy-ssg" folder (hopefully that wasn't still needed)


website/content/tutorials/02-embedded-programming.md
```diff
@@ -60,7 +60,7 @@ The RAM will contain garbage after startup. Our variables aren't initialized and
 
 To do this, we first have to tell the compiler to store our pre-initialized data somewhere in flash memory, so we can copy it into RAM at startup. This is done via the linker script:
 
-```ld
+```
```
MatthiasPortzel, line 63: New version of Zine errors on syntax highlighting it doesn't support, so I dropped `ld` (https://github.com/kristoff-it/zine/issues/46)

MatthiasPortzel:
@EliSauder I think this duplicates some work you did previously, does this look right?

website/build.zig.zon
```diff
@@ -4,8 +4,8 @@
     .paths = .{"."},
     .dependencies = .{
         .zine = .{
-            .url = "https://github.com/kristoff-it/zine/archive/03f80646b83cadb2e693ff5d97445d3e16c8e222.tar.gz",
-            .hash = "1220e3e4938edf652776349c45b3bb58774d540a050034488ff8dab7dbe410cc2977",
+            .url = "git+https://github.com/kristoff-it/zine#e21b179b78de54ab48efe38d2378e87a56cb3542",
+            .hash = "1220781f118454bbc87d7efbd244b4d1ef029e76a6113ddf4752a6f3cd85879dbb3b",
```
EliSauder, line 8: I would suggest going with the tagged release just since kristoff has made one.

website/content/getting-started.md
```diff
@@ -1,11 +1,11 @@
 ---
 {
-  "title": "Home",
-  "date": "2020-07-06T00:00:00",
-  "author": "Felix Queißner",
-  "draft": false,
-  "layout": "getting-started.html",
-  "tags": []
+  .title = "Home",
+  .date = "2020-07-06T00:00:00",
```
EliSauder, line 4: Does this build without the `@date(..)`?

website/content/tutorials/02-embedded-programming.md
EliSauder: You can use cpp instead of ld and it works well enough.

EliSauder (commented):
The only last thing other than the other comments, is the layout files should be moved to .shtml over .html as that is the file extension kristoff chose for the SuperHTML templating language.

EliSauder:
Also this won't automatically deploy to the website yet since the workflows are broken (what I've been working on updating). So this should probably wait until that gets merged in to main.

EliSauder:
Resolves https://github.com/ZigEmbeddedGroup/website/issues/13

website/content/getting-started.md
MatthiasPortzel: It does build but it's not correct. I'll fix them up.

website/build.zig.zon
MatthiasPortzel: I don't understand the Zig package system to know what advantage we get from using a different URL.

ikskuh:
Nice work!

website/build.zig.zon
EliSauder: The main version is using a well defined release version that the library creator has determined is good to go rather than a specific commit. If we want to stick with a commit that is alright, I just think using explicit versions tends to make more sense to me. Using zon, you can do something like ...#v0.4.1 as the # just expects a particular git ref.

ikskuh:
@EliSauder: Merging this should be fine with your changes, right?

EliSauder:
Yeah, i just rebased some of the changes so there shouldn't be any conflicts.

EliSauder:
@ikskuh the only remaining concern for the website is that zine has been updated. But we could have that be its own PR.

ikskuh:
> @ikskuh the only remaining concern for the website is that zine has been updated. But we could have that be its own PR.

Yeah, sounds like a good idea :)

PR #233, opened 2024-08-26
Add GPIO Config to SPI Example

haydenridd:
- Added GPIO configuration to SPI example
- Updated terminology


haydenridd:
Merging as this is incredibly minor :)

PR #242, opened 2024-10-03
Update gen.zig to produce just enums

taylorh140:
Changed the gen.zig to produce just enums instead of packed unions.

### DIFF
--- a/tools/regz/src/gen.zig
+++ b/tools/regz/src/gen.zig
@@ -1113,10 +1113,7 @@ test "gen.field with named enum" {
         \\            };
         \\
         \\            TEST_REGISTER: mmio.Mmio(packed struct(u8) {
-        \\                TEST_FIELD: packed union {
-        \\                    raw: u4,
-        \\                    value: TEST_ENUM,
-        \\                },
+        \\                TEST_FIELD: TEST_ENUM,
         \\                padding: u4,
         \\            }),
         \\        };
@@ -1142,13 +1139,10 @@ test "gen.field with anonymous enum" {
         \\    pub const peripherals = struct {
         \\        pub const TEST_PERIPHERAL = extern struct {
         \\            TEST_REGISTER: mmio.Mmio(packed struct(u8) {
-        \\                TEST_FIELD: packed union {
-        \\                    raw: u4,
-        \\                    value: enum(u4) {
-        \\                        TEST_ENUM_FIELD1 = 0x0,
-        \\                        TEST_ENUM_FIELD2 = 0x1,
-        \\                        _,
-        \\                    },
+        \\                TEST_FIELD: enum(u4) {
+        \\                    TEST_ENUM_FIELD1 = 0x0,
+        \\                    TEST_ENUM_FIELD2 = 0x1,
+        \\                    _,
         \\                },
         \\                padding: u4,
         \\            }),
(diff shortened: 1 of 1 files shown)


taylorh140:
as per @ikskuh's statement. that built-ins remove the need for the packed unions.

tools/regz/src/gen.zig
```diff
@@ -824,10 +824,12 @@ fn write_fields(
         } else if (db.attrs.@"enum".get(fields[i].id)) |enum_id| {
             if (db.attrs.name.get(enum_id)) |enum_name| {
                 try writer.print(
-                    \\{}: packed union {{
-                    \\    raw: u{},
-                    \\    value: {},
-                    \\}},
+                // These were phased out in favor of using new builtins
```
mattnite, line 827: Don't need this comment, and remove the commented out code

tools/regz/src/gen.zig
```diff
@@ -824,10 +824,12 @@ fn write_fields(
...
+                //    \\{}: packed union {{
+                //    \\    raw: u{},
+                //    \\    value: {},
+                //    \\}},
+                    \\{}: {},
```
mattnite, line 832: You're going to have to remove `next.size` from the arguments now it's only printing two things.

tools/regz/src/gen.zig
```diff
@@ -836,9 +838,10 @@ fn write_fields(
                 });
             } else {
                 try writer.print(
-                    \\{}: packed union {{
-                    \\    raw: u{},
-                    \\    value: enum(u{}) {{
+                // \\{}: packed union {{
+                // \\    raw: u{},
+                // \\    value: enum(u{}) {{
+                    \\{}: enum(u{}) {{
```
mattnite, line 844: Same as above.

mattnite (changes requested):
You'll have to change some of the unit tests in addition to the comments I've made thus far.

tools/regz/src/gen.zig
taylorh140: This was a terrible error. so embarrassing. I got the right number of items included now.

tools/regz/src/gen.zig
taylorh140: ditto.

taylorh140:
I'm not quite sure where the test are located... yet. Ill find them and get this polished.

mattnite:
@taylorh140 tests are at the bottom of the file

tools/regz/src/gen.zig
taylorh140: Done.

taylorh140:
Ok, I think I have the changes you requested. I'm not sure why the build is failing though.

I think I have the test cases aligned but. However i can seem to build zig test. I'll have to see if I added that.

taylorh140:
It doesnt look like added it:
regz 
zig build test  gives:
image

mattnite:
@taylorh140 check the generated code, we're failing AST checks. Could be something like a missing curly brace

taylorh140:
@mattnite Hey it looks like the AST problem is better, I missed removing a closing bracket in the extended cases.

I think that the next thing is maybe updating how some of the items use the packed union as it looks like it's causing some issues. I think i wouldn't mind hunting these down if you don't thing it would be too much of a scope creep for the PR.

mattnite:
@taylorh140 Yup, a low level API has been changed so now all uses of it need to be updated. I care more about commit atomicity than perceived patch complexity, so to me it's not even scope creep, it's an unfinished part of the task you're trying to do.

mattnite:
Closing because #328 implements this now that the regz rewrite is complete.

PR #243, opened 2024-10-04
ESP 32 blinky example - typo fix

RokKos:
I was looking through the example for esp 32 for micro zig and I noticed a small bug/typo in setting blue led to high.


ikskuh:
Thanks!

PR #262, opened 2024-10-16
Support I2C and SPI functions in the pins abstraction

MatthiasPortzel:
Just got around to updating microzig to past @haydenridd's SPI and I2C re-writes. Formerly, the function and pull were set by the I2C interface.

=> https://github.com/ZigEmbeddedGroup/microzig/pull/213/files#diff-8a4efe08a401e079bae57d66c0dd3ef49a3839a6065c3c1058ac8a0eaf142623L110

Since that's no longer the case, I'd like to set the function in the pins interface. This Microzig patch lets me do that.

=> https://tildegit.org/matthias/Micromouse/src/branch/master/src/physical-bot.zig#L212

Arguably we could/should also set the direction and pull based off of these properties (previously the I2C interface used `set_pull(.up)` on both pins), but it's slightly less obvious how to do that correctly and doesn't seem to be necessary in my test.

### DIFF
--- a/port/raspberrypi/rp2xxx/src/hal/pins.zig
+++ b/port/raspberrypi/rp2xxx/src/hal/pins.zig
@@ -185,6 +185,32 @@ pub const Function = enum {
         };
     }
 
+    pub fn is_spi(function: Function) bool {
+        return switch (function) {
+            .SPI0_RX,
+            .SPI0_CSn,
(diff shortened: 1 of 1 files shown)


ikskuh (commented):
Thanks!

ikskuh (approved):

PR #263, opened 2024-10-19
Added general purpose bare metal allocator.

leibupro:
Howdy,

currently using this bare metal allocator on a M4 CPU micro controller. Perhaps you have a use for it too.

Wrote already a bunch of unit tests but plan to write further test cases when I find time.

Cheers P.

### DIFF
--- a/core/src/core/alloc.zig
+++ b/core/src/core/alloc.zig
@@ -1,0 +1,1403 @@
+//
+//  ############################################################
+//
+//   File:     alloc.zig
+//
+//
+//   Purpose:  General purpose memory allocator
+//             to be used in bare metal applications
+//             which do not have a backing operating
+//             system nor a memory management unit.
+//
+//
+//   Remarks:  - Operates on a finite byte field
+//               (u8 array) within the .data segment.
+//
+//
+//   Author:   P. Leibundgut
+//             <https://github.com/leibupro>
+//
+//
+//   Date:     09/2024
+//
+//  ############################################################
+//
+
+const std = @import( "std" );
+const ArrayBitSet = std.bit_set.ArrayBitSet;
+
+const assert = std.debug.assert;
+const print = std.debug.print;
+const alignPointerOffset = std.mem.alignPointerOffset;
+
+const Allocator = std.mem.Allocator;
+
+
+const FreeCellAuditResult = struct
+{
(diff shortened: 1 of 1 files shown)


ikskuh:
Are you open to do some adjustments on the allocator?

leibupro:
In general, yes. Depends a little on the change set size and currently available time.

ikskuh:
> ```zig
> pub fn BareMetalAllocator( comptime memory_area: []u8 ) type
> ```

Can you write at least a doc comment on what `memory_area` is supposed to do? You allocate `memory_area.len` chunks of pointers, so i don't assume it's the memory area you use for user allocations, right?

leibupro:
`memory_area` (assigned to `memory_vault` internally) is indeed a fat pointer to the memory area which is used for the memory allocations. The byte field which is managed by the allocator. The `metadata_space` is an array of memory cell metadata (specific memory slice and a pointer to the next cell) of length `memory_area.len`. This means, that in the extreme case, the allocator can hold `n` memory cells of 1 byte length. Where `n` is the length in bytes of the `memory_area`.

You're right, I'm going to add a descriptive doc comment like written in the previous paragraph.

ikskuh:
Just to get it right?

```zig

pub fn BareMetalAllocator( comptime memory_area: []u8 ) type {
    return struct {
        const MemoryCell = struct {
            memory: []u8,
            next: ?*MemoryCell,
        };
        memory_vault: []u8 = memory_area,
        metadata_space: [ memory_area.len ]MemoryCell = undefined,
    };
}
```

So i pass a slice with 16kB to `memory_arena`, when i want to have 16kB of heap? But wouldn't that mean that the allocator will be 12*16kB = 192kB bytes large (assuming 32 bit pointers), as `MemoryCell` is 12 bytes large

leibupro:
Yes, that is the case. Since the smallest memory cell can be one byte and the allocator has to manage `memory_area.len` 1 byte cells theoretically. It is described before the test code starts, line 710 ff.

ikskuh:
It doesn't make sense to use such an allocator in a memory constraint area, considering that it has such a crazy memory overhead. It would allow you having less than 150 bytes of heap on a typical AVR device, and on a lot of other MCUs less than 2k

PR #264, opened 2024-10-21
rp2xxx: improve RP2350 port

vesim987:
I didn't test this on a real hardware, i don't have pico 2 yet.


haydenridd:
It will be a few days but I will review this!

haydenridd (changes requested):
Great idea on changing `compatibility.zig` to just be a constant, definitely no need for a function :) No problem with any of the code changes, only request is:
- Let's not move over any new examples to "chip agnostic examples" yet until we've done some deep verification on hardware the HAL actually still works on the RP2350 vs. the RP2040. I spot checked a couple of the new ones on my HW and they don't appear to be working yet (for instance, multi-core doesn't blink as expected).  This makes sense though since from my initial stab there was just enough little differences in each peripheral to warrant going through with a fine-toothed comb.

Spot checking the examples and diagnosing why they don't work is a great window into getting the HALs RP2350 compatible, so I'll keep doing that but don't want this PR to get any more stale!

vesim987:
I rebased against main and pushed the commit with examples to `rp2xxx-examples` branch  so it doesn't get lost.

haydenridd (approved):
Awesome, LGTM! Just verified all the known working examples for RP2350 on my HW and all is still well.

PR #265, opened 2024-10-22
Fixed RP2xxx usb reset register masks

eugenepentland:
Before the rework on resets.zig the mask had all of the rp2xxx hal/resets.zig Mask had all of the values in the packed struct as false by default, so when reset.reset(.{pll_usb}) was run, it would make a bitmask for only that register. But now it makes a bitmask for every register and takes all pins out of reset mode.

I just made it so it gets the bit offset of the register and creates a bitmask for only the register in question. This should support rp2040 and rp2350, but currently is only tested with the usb cdc example on the rp2040.

### DIFF
--- a/port/raspberrypi/rp2xxx/src/hal/usb.zig
+++ b/port/raspberrypi/rp2xxx/src/hal/usb.zig
@@ -119,8 +119,12 @@ pub const F = struct {
         // already moved over to the 12MHz XOSC. We just need to make it x4 that
         // clock.
         //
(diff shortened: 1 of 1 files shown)


eugenepentland:
Turns out controlling those reset bits aren't needed because in reset.masks, it sets the usbctrl and pll_usb to be on by default.

arkadiuszwojcik:
@eugenepentland I can confirm this fix works, but if those values are default why setting them once again frezzes execution?

PR #268, opened 2024-11-01
Deploy website to github pages

mattnite:


EliSauder:
How will this relate to #232?

mattnite:
@EliSauder This is the first step in a redesign of the website where #232 is automation. Switching to github pages makes deployment for ZEG simpler so that's why I'm going with that option.

I'd like to explore hosting packages on github pages as well since it's also static content, but that's a bit more complicated.

EliSauder:
We could host the packages using GitHub releases and then link to them potentially.

PR #284, opened 2024-11-11
Comptime USB endpoints number

arkadiuszwojcik:
Previous USB refactor for simplification assumed a fixed number of 4 USB endpoints. This PR introduces a configuration option to specify the desired number of endpoints, but no more than the board's maximum number of endpoints. Every board have to provide its own config with proper defaults which coresponds to max number of supported endpoints.


haydenridd (approved):
All these changes seem sensible to me! I haven't messed around a ton with the USB drivers yet so I would want someone a little more familiar to take a peek before we merge but no problems on my end.

mattnite:
@haydenridd I'm going to merge this, it looks good to me and we'll have a USB initiative in the future.

PR #287, opened 2024-11-15
Use program origin when computing label addresses

tomasz-lisowski:
When using the `.origin` directive in PIO programs, labels would receive addresses relative to the first instruction in the program instead of absolute addresses that take the origin into consideration. This led to invalid jump addresses leading to incorrect encodings for `JMP` instructions.

I also added a test that checks the behaviour of this fix.


mattnite (approved):
Thank you for finding and fixing this bug! great test too.

tomasz-lisowski:
Thanks for the quick merge!

PR #289, opened 2024-11-17
Add helpers for reading from PIO RX FIFO and for PIO SM pin configuration

tomasz-lisowski:
When configuring PINs for use with PIO SMs, currently we only set the pin bases and counts for each group of pins inside the `PINCTRL` register. This does not affect the actual direction and default value that these pins will have from the point-of-view of the state machine.
The pico-sdk will [internally configure the desired pins](https://github.com/raspberrypi/pico-sdk/blob/efe2103f9b28458a1615ff096054479743ade236/src/rp2_common/hardware_pio/pio.c#L234-L268). It does that by saving the `PINCTRL` and `EXECTRL` registers, then it configures the `SET` group to the pin we want to configure so that it can execute `SET PINDIRS, n` or `SET PINS, n` for the pin(s) being configured.

Eventually it would be ideal to include pin value and pin direction configuration as part of the `sm_set_pin_mappings` function. For now I think it's best to keep this as separate functions as I have done here.

I also included an ability to read and blocking-read the RX FIFOs. Currently we only have this capability for the TX FIFOs.

### DIFF
--- a/port/raspberrypi/rp2xxx/src/hal/pio.zig
+++ b/port/raspberrypi/rp2xxx/src/hal/pio.zig
@@ -308,6 +308,124 @@ pub const Pio = enum(u1) {
         });
     }
 
+    pub fn sm_set_pindir(self: Pio, sm: StateMachine, base: u5, count: u5, dir: gpio.Direction) void {
+        const sm_regs = self.get_sm_regs(sm);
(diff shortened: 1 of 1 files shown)


mattnite (approved):
Thank you! I

PR #290, opened 2024-11-20
Regz patching

mattnite:
This patch shows a minimal example of patching the register schema _after_ it has been loaded into Regz. It was a goal to allow the patches to be written in zig for strong typing and potential for code completion.

We can't instantiate the Regz code directly in build scripts as it has C dependencies, so Regz is built as an executable and it operates on files.

`Chip` now has a `patches` field which is a list of patches that are applied to the Database in order. These patches are serialized to a newline-delimited JSON file that is passed to the Regz executable. Regz loads the register schema into its database and then applies the patches. In the future the plan will be to allow to use a newlined-delimted JSON file directly, I have plans to create a graphical program for patching register schemas.

Right now creating enums and assigning that shared type to different fields is all that's supported.


build/build.zig
```diff
@@ -98,13 +110,23 @@ pub fn add_firmware(
     // On demand, generate chip definitions via regz:
     const chip_source = switch (chip.register_definition) {
         .json, .atdf, .svd => |file| blk: {
-            const regz_exe = mz.dependency("microzig/tools/regz", .{ .optimize = .ReleaseSafe }).artifact("regz");
+            const regz_exe = mz.dependency("microzig/tools/regz", .{ .optimize = .Debug }).artifact("regz");
```
haydenridd, line 113: Intentional?

tools/regz/src/main.zig
```diff
@@ -9,6 +9,10 @@ const assert = std.debug.assert;
 
 const svd_schema = @embedFile("cmsis-svd.xsd");
 
+pub const std_options = std.Options{
+    .log_level = .debug,
```
haydenridd, line 13: Revert once finalized?

haydenridd (approved):
Loving where this is going! No complaints on my end.

Only question I had is how easy is it for a user to override/add their own patches in the build interface? I think that might be a desirable feature for someone to put their own tweaks in in `build.zig` via `firmware.add_patches()` or something of the sort...

build/build.zig
mattnite: Yes! Since regz isn't run very often, I think we'll end up wasting less time if we don't optimize regz, and run it in debug mode. On first run of the project the release build time is significant.

mattnite:
@haydenridd I wasn't able to make `firmware.add_patches()` without a bunch of rework. BUT I did add a patches field to the `add_firmware()` options, which should be just as good :)

PR #259, opened 2024-10-15
Build system rewrite

tact1m4n3:
## Build system rewrite

If anyone has any other ideas of things to add to the new build system (even stuff that was there before and I missed it), please leave a comment in this PR and I will add it to the list. 

### Tasks to complete
- [x] Initial version of the build system done.
- [x] Better way for targets to select a cpu/arch
- [x] Patch elf
- [x] Build formats
- [x] Better ux for port writing
- [ ] Change all ports to work with the new build system (wip)
    - [ ] rp2xxx: find a way to build bootrom and also expose it so as to allow user targets to reference it
    - [x] lpc
    - [x] stm32
    - [x] nrf5x
    - [x] gd32
    - [x] atsam
    - [x] avr
    - [ ] esp
- [ ] Change all examples to work with the new build system
    - [x] rp2xxx
    - [x] lpc
    - [x] stm32
    - [x] nrf5x
    - [x] gd32
    - [x] atsam
    - [x] avr
    - [ ] esp
- [x] Final touches
    - [x] Add the remaining options from the previous build system
    - [x] Add the `package` command back to the root `build.zig`

PR #223, opened 2024-07-19
[WIP] Support for WCH CH32V003xx

akiroz:
Implement preliminary support for WCH CH32V003xx targets
Pending hardware verification.

PR #298, opened 2024-11-28
Fixed building error for WCH CH32V

nosuz:
I have uncommented to build examples for the WCH CH32V series and fixed an error: "error: evaluation exceeded 1000 backwards branches" when compiling them.

In addition to this fix, I have restored the configuration to import the CPU-specific startup module. Without this startup module, the absolute jump to the initialization function _start() is omitted.
I think this configuration was dropped by the commit 62734a7 "Build system rewrite (#259)". 

If the default riscv32 startup module (core/src/cpus/riscv32.zig) is used, the main routine microzig_main() is placed before _start() on a flash memory. The WCH CH32V series cpu has no startup vector table and runs from the top of the memory after the reset is released, and the microzig_main() is executed without initialization by _start().
And the builded binaries are asumed to run on the flash memory but cpu execute the mirror of flash memory starting from address 0. So, an absolute jump to the region of flash memory is required at the begining of binaries.

This modification might be also needed for the Gigadevices GD32 series port.

I am not sure this is acceptable way to call _start() for initialization, and I am very happy to hear your comments and advices.


mattnite:
@nosuz please include a LICENSE file for the ch32v package, and I think we'll be good to go

nosuz:
@mattnite Thank you for advice. I have included LICENSE files and also rebased from the current head.

nosuz:
@mattnite Thank you for approving and merging my Pull Request. I will make a new PR for PR #223 soon.

PR #303, opened 2024-12-03
[RP2xxx] Update pins helper to create pins as comptime-available data instead of zero-sized-types

MatthiasPortzel:
**TL;DR:** Replaces `pins = pin_config.apply()` with `pins = comptime pin_config.pins(); pin_config.apply()`.

**Long version:**
This PR updates the "pin configuration helper," `pins.GlobalConfiguration`, to return a `pins` type which is more suitable for normal usage than the ad-hoc zero-sized-types previously created. For example, a `pins.led_a` could be saved into a runtime struct or passed to a struct initializer at comptime, just like a primitive Zig value. In order to make this possible, there are a couple of distinct changes being made:

1. For .sio-function pins, the resulting pin is now a `gpio.Pin`. This removes the previous ZSTs and unifies the API. As well as making these data that can be used at runtime.
2. For pwm-function pins, the Pwm type has been rewritten to be a runtime-compatible struct, instead of a zero-sized type. (Previously interacting with runtime-known Pwm pin numbers was impossible.)
3. Instead of having `pin_config.apply()` return the `pins` type, as was previously the case, `.apply()` now returns nothing and `pin_config.pins()` returns the `pins`. This was necessary in order to allow `pins` to exist as normal Zig data. (Since `apply` has to be called at runtime, only the type of the return value is known at `comptime`.)

The only breaking change for users is splitting calls to `apply()` into a new call to `pins()`. (See the change to the examples for an example of what that looks like.)

This represents a regression in that you will no longer get compile-errors for e.g. reading from an output pin. This is acceptable to me for the time being because it matches our philosophy for the rest of the HAL—configuration is the user's responsibility. However, I'm not opposed to changing the GPIO interface to be stricter, and I have a couple of ideas on how to do that. But fundamentally I think that change would need to happen to `gpio.Pin` (e.g. maybe it splits into `gpio.ReadPin` and `gpio.WritePin`)—I don't think it's tenable to have a split in the type system between a configured pin and an unconfigured pin.

I'm 99% sure that accessing an ADC-type pin on main right now doesn't work. Since the ADC code is old, it could really use a rewrite, so I've left it broken in this PR (replaced it with a TODO). I'm happy to update the pin configuration helper to support ADC if it's clear what we want to do with the HAL.

### DIFF
--- a/port/raspberrypi/rp2xxx/src/hal/pins.zig
+++ b/port/raspberrypi/rp2xxx/src/hal/pins.zig
@@ -395,7 +395,7 @@ pub const GlobalConfiguration = struct {
                 if (pin_config.function == .SIO) {
                     pin_field.type = gpio.Pin;
                 } else if (pin_config.function.is_pwm()) {
-                    pin_field.type = pwm.RuntimePwm;
+                    pin_field.type = pwm.Pwm;
                 } else if (pin_config.function.is_adc()) {
                     // TODO
                     // pin_field.type = adc.Input;
@@ -427,7 +427,7 @@ pub const GlobalConfiguration = struct {
                 if (pin_config.function == .SIO) {
                     @field(ret, pin_config.name orelse field.name) = gpio.num(@intFromEnum(@field(Pin, field.name)));
                 } else if (pin_config.function.is_pwm()) {
-                    @field(ret, pin_config.name orelse field.name) = pwm.RuntimePwm {
+                    @field(ret, pin_config.name orelse field.name) = pwm.Pwm {
                         .slice_number = pin_config.function.pwm_slice(),
                         .channel = pin_config.function.pwm_channel(),
                     };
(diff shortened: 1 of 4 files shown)


port/raspberrypi/rp2xxx/src/hal/pins.zig
```diff
@@ -425,48 +418,36 @@ pub fn Pins(comptime config: GlobalConfiguration) type {
...
+                        .channel = pin_config.function.pwm_channel(),
+                    };
+                }
 
-    comptime {
-        const pin_field_count = @typeInfo(Pin).Enum.fields.len;
-        const config_field_count = @typeInfo(GlobalConfiguration).Struct.fields.len;
-        if (pin_field_count != config_field_count)
-            @compileError(comptimePrint("{} {}", .{ pin_field_count, config_field_count }));
+                // TODO
```
mattnite, line 436: Please resolve a TODO, or create a FIXME with a link to a github issue where you can provide some insight.

mattnite (changes requested):

port/raspberrypi/rp2xxx/src/hal/pins.zig
MatthiasPortzel: Created https://github.com/ZigEmbeddedGroup/microzig/issues/311 since I don't know how the ADC works and I don't want this PR to wait for me to learn.

mattnite (approved):

PR #304, opened 2024-12-03
rp2xxx: Update rp2350 SVD

Grazfather:


Grazfather:
I believe that this will require also updating the rp2040 SVD. I tried doing so, but the patches no longer apply, and I don't know how to update them, so I might need a hand there.

mattnite:
@Grazfather Please document the changes you want to make to the SVD. The plan is to vendor SVDs from the manufacturer and not alter them, but apply semantic patches so that we know how we got there -- The size of your patch lends more credence to this approach: `11,084 additions, 2,656 deletions`. Using line based diffs and git history will not scale for a manufacturer with 2 MCUs, let alone thousands like STM32.

I'm currently working on implementing this in regz and tooling to make the process smoother. I'd like to know the nature of your changes so that I can prioritize them.

Grazfather:
I certainly don't want to modify the SVD itself. What I mean is that the existing [patch file](https://github.com/ZigEmbeddedGroup/microzig/blob/main/port/raspberrypi/rp2xxx/patches/rp2040.zig) must be modified to work with the new SVD. As it exists, regz complains, but if I don't apply it, there are issues building rp2040 targets.

Grazfather:
This is all covered in #264, but it'll need a rebase and update to work with the rp2040 patch.

Grazfather:
> I believe that this will require also updating the rp2040 SVD

What I meant here is that this PR should also update the rp2040 SVD, but not modify it from what is in the pico-sdk.

PR #307, opened 2024-12-06
Get esp32 examples building again

Grazfather:


nosuz:
@Grazfather Thank you for making the esp32 examples available.

I read your patch and found this line in build.zig.

```
@setEvalBranchQuota(2000); // DELETEME
```

I think this built-in method can remove as you commented.
There is the same line at a few lines above this line, and I could build successfully without this line under `examples/espressif/esp` and `examples/wch/ch32v`.

Will you check this for me?

Grazfather:
Seems I don't need it anymore. That said, I don't think this branch will work for you, it still needs work.

mattnite (approved):

PR #310, opened 2024-12-09
Fix NVIC level IRQ control for RP2040

haydenridd:
- Reworked the IRQ API to be a little easier to use + made it actually work :)


haydenridd:
Converting to draft so I can get the RP2350 version working as well

haydenridd:
All ready to go. Even tested getting a GPIO rising edge IRQ to fire on RP2350! Note that this only makes it so the top level NVIC ISR enables/disables function now. More work still needs to be done on the RP2350 side (and probably the RP2040 side too) to have each peripheral instance have a method that will enable/disable the peripheral level interrupt.

vesim987 (approved):

PR #320, opened 2024-12-19
rp2xxx: Add support for assembling RP2350-specific PIO

Grazfather:
RP2350 adds the following new instruction features:
* [x] Adds PINCTRL_JMP_PIN as a source for the WAIT instruction, plus an offset in the range 0-3.
    * This gives WAIT pin arguments a per-SM mapping that is independent of the IN-mapped pins.
* [x] Adds PINDIRS as a destination for MOV.
    * This allows changing the direction of all OUT-mapped pins with a single instruction: MOV PINDIRS, NULL or MOV
PINDIRS, ~NULL
* ~[ ] Adds SM IRQ flags as a source for MOV x, STATUS~
    * This allows branching (as well as blocking) on the assertion of SM IRQ flags.
* [x] Extends IRQ instruction encoding to allow state machines to set, clear and observe IRQ flags from different PIO
blocks.
    * There is no delay penalty for cross-PIO IRQ flags: an IRQ on one state machine is observable to all state
machines on the next cycle.
* [x] Adds the FJOIN_RX_GET FIFO mode.
    * A new MOV encoding reads any of the four RX FIFO storage registers into OSR.
    * This instruction permits random reads of the four FIFO entries, indexed either by instruction bits or the Y
scratch register.
* [x] Adds the FJOIN_RX_PUT FIFO mode.
    * A new MOV encoding writes the ISR into any of the four RX FIFO storage registers.
    * The registers are indexed either by instruction bits or the Y scratch register.
RP2350 adds the following security features:
* Limits Non-secure PIOs (set to via ACCESSCTRL) to observation of only Non-secure GPIOs. Attempting to read a
Secure GPIO returns a 0.
* Disables cross-PIO functionality (IRQs, CTRL_NEXTPREV operations) between Non-secure PIO blocks (those which
permit Non-secure access according to ACCESSCTRL) and Secure-only blocks (those which do not).
RP2350 includes the following general improvements:
* [x] Increased the number of PIO blocks from two to three (8 → 12 state machines).
* Improved GPIO input/output delay and skew.
* Reduced DMA request (DREQ) latency by one cycle vs RP2040


port/raspberrypi/rp2xxx/src/hal/pio/assembler.zig
```diff
@@ -121,9 +123,25 @@ fn format_compile_error(comptime message: []const u8, comptime source: []const u
...
+};
+
+fn cpuToFormat(cpu: CPU) Format {
+    return switch (cpu) {
+        .RP2040 => .RP2040,
+        .RP2350 => .RP2350,
+    };
+}
+
+pub fn assemble(comptime format: Format, comptime source: []const u8, comptime options: AssembleOptions) Output {
```
Grazfather, line 139: We don't want to make the user provide the format

port/raspberrypi/rp2xxx/src/hal/pio/assembler/encoder.zig
```diff
@@ -624,22 +681,22 @@ const expect = std.testing.expect;
...
-    const tokens = try tokenizer.tokenize(source, diags, .{});
-    var encoder = Encoder(.{}).init(tokens.slice());
+fn encode_bounded_output_impl(comptime chip: Chip, source: []const u8, diags: *?assembler.Diagnostics) !Encoder(chip, .{}).Output {
+    const tokens = try tokenizer.tokenize(chip, source, diags, .{});
+    var encoder = Encoder(chip, .{}).init(tokens.slice());
     return try encoder.encode_output(diags);
 }
 
-fn encode_bounded_output(source: []const u8) !Encoder(.{}).Output {
+fn encode_bounded_output(comptime chip: Chip, source: []const u8) !Encoder(chip, .{}).Output {
```
Grazfather, line 690: I would like to run this for both chips, but the return type depends on the type.

port/raspberrypi/rp2xxx/src/hal/pio/assembler/tokenizer.zig
```diff
@@ -62,1082 +64,1305 @@ pub const Value = union(enum) {
...
+            };
         }
 
-        const num = Value{
-            .expression = if (has_mode)
-                (try self.get_arg(diags)) orelse {
-                    diags.* = Diagnostics.init(self.index, "irq (mode) <num> (rel): failed to get num argument", .{});
-                    return error.MissingArg;
+        fn get_movrx(self: *Self, diags: *?Diagnostics) TokenizeError!Token(chip).Instruction.Payload {
+            // TODO: Assuming one space after opcode
```
Grazfather, line 812: Only used for diag, not sure if anyone cares about this assumption

port/raspberrypi/rp2xxx/src/hal/pio/assembler/tokenizer.zig
```diff
@@ -62,1082 +64,1305 @@ pub const Value = union(enum) {
...
+                        .movtorx = .{
+                            .idxl = true,
+                            .idx = 0,
+                        },
+                    };
+                } else {
+                    // -- Parse out the index
+                    diags.* = Diagnostics.init(
+                        dest_idx,
+                        // @intCast(self.index - dest_str.len + 1),
```
Grazfather, line 839: x

mattnite:
@Grazfather there's one item unchecked, I take it you forgot to check it?

Grazfather:
@mattnite I don't understand that one. I don't see any difference in the encoding of the (plain) `mov` instruction in https://datasheets.raspberrypi.com/rp2350/rp2350-datasheet.pdf (page 893).

mattnite (approved):

PR #333, opened 2024-12-24
Cleanup Old/Extra Files

EliSauder:
Did a pass through of the repo removing extra license files and some files that are remaining from moving to a monorepo.

This mirrors the cleanup changes I made in the original workflows update PR (#232) but this is a separate PR to implement the changes.


mattnite:
@EliSauder The license files exist so that they are included in distributed tarballs.

EliSauder:
> @EliSauder The license files exist so that they are included in distributed tarballs.

@mattnite I see, but won't we generally be distributing it as a single tarball from the output for boxzer? In which case wouldn't the idea of just having the root one be fine (Also a note on this, I just checked the boxzer output and at the moment the root tar doesn't have a LICENSE file, not sure if this is intentional or not)?

mattnite:
@EliSauder the tarball in the release is effectively saved state of the packages for a version, it exists for the deployment for the pages site which iterates releases, grabs these, and extracts them for the downloads folder. The License files are needed for each package tarball, and and these are stored in that boxzer-out tarball

EliSauder:
@mattnite I think I understand now. But regardless of my full understanding, I have reverted those changes.

mattnite (approved):

PR #335, opened 2024-12-24
Add regz/uf2 release build option

EliSauder:
Added the ability to run `zig build release-regz` and `zig build release-uf2` which will generate the platform binaries for the respective tools.

This is planned to be used in some CI improvement where regz and uf2 will also be published.


EliSauder:
I implemented the release in the root build.zon file, but I can also move it to the per-project build file if that is what makes more sense.

mattnite (approved):

PR #338, opened 2024-12-27
Added watchdog timer API for both RP2040 and RP2350

haydenridd:
Tested on both RP2040 and RP2350! Only functionality I left out was the ability via the scratch registers to force the Pico to boot with a manually specified PC/SP. Wouldn't be that hard to implement, I just don't think there's any demand for it yet since that's a super advanced use case.


mattnite (approved):
Hell yeah, this looks great

PR #344, opened 2025-01-03
add an example of how to use MicroZig without HAL

lcp5y3:
I would try MicroZig, but doesn't have a supported board, so I try to do a *blinky* program with only register access.
I haven't found any examples of this case, so I propose you this one for an STM32L031.


mattnite (changes requested):
Revert those renames, put this under `examples/no_hal`

lcp5y3:
I'm not sure to understand the path naming you want. So, I've moved `examples/stmicro/stm32/stm32_no_hal` to `examples/no_hal/stm32l031`.

Is it OK for you ?

mattnite:
@lcp5y3 please rename stm32_hal to stm32 please!

mattnite (approved):

PR #359, opened 2025-01-23
Add A4988 Stepper driver

Grazfather:
* Add `Clock_Device` to drivers, using those types, which can be used in a driver
* (wip) Add an A4988 stepper motor driver into drivers


drivers/stepper/A4988.zig
```diff
@@ -0,0 +1,395 @@
...
+            try pin.write(if (self.enable_active_state == .high) .low else .high);
+        }
+    }
+
+    pub fn set_rpm(self: Self, rpm: f64) void {
+        self.rpm = rpm;
+    }
+
+    pub fn set_microstep(self: *Self, microsteps: u8) !u8 {
+        // If any MS pins are not defined, return an error and don't change anything
```
Grazfather, line 131: hm: I think we want to set the struct attribute either way, for when the MS pins are strapped

Grazfather: done, but a bit janky, by having `init` call `init_microstep` to 'force' the attribute to the correct value, but then calling `set_microstep` which would return an ignore error.

Basically, if the pins are not defined, we want microsteps set only in `init`, so we know how many steps to expect, but not allow them to be changed (since they are strapped).

drivers/stepper/stepper.zig
```diff
@@ -0,0 +1,539 @@
...
+                        const sqrt_candidate = t * t - 2 * a2 * d; // in √b^2-4ac
+                        if (sqrt_candidate >= 0)
+                            speed = @min(speed, (t - std.math.sqrt(sqrt_candidate)) / a2);
+                    }
+                    // How many microsteps from 0 to target speed
+                    self.steps_to_cruise = @intFromFloat(@as(f64, microstep_f * (speed * speed)) / (2 * accel_f));
+                    // How many microsteps are needed from cruise speed to a full stop
+                    self.steps_to_brake = @intFromFloat(@as(f64, @floatFromInt(self.steps_to_cruise)) * accel_f / decel_f);
+                    if (self.steps_remaining < self.steps_to_cruise + self.steps_to_brake) {
+                        // Hannot reach max speed, will need to brake early
```
Grazfather, line 220: typo

drivers/stepper/stepper.zig
```diff
@@ -0,0 +1,539 @@
...
+        }
+        pub fn next_action(self: *Self) !mdf.time.Duration {
+            if (self.steps_remaining > 0) {
+                self.delay_micros(self.next_action_interval, self.last_action_end);
+                //  DIR pin is sampled on rising STEP edge, so it is set first
+                try self.dir_pin.write(self.dir_state);
+                try self.step_pin.write(.high);
+                // Absolute time now
+                const start = self.clock.get_time_since_boot();
+                // var m: u32 = @intCast(self.clock.get_time_since_boot().to_us());
```
Grazfather, line 306: xx

drivers/framework.zig
```diff
@@ -34,6 +34,8 @@ pub const input = struct {
     };
 };
 
+pub const stepper = @import("stepper/stepper.zig");
```
Grazfather, line 37: We could make this a struct and export the specific drivers so the user doesn't have to do:

```zig
const stepper_driver = microzig.drivers.stepper;
const A4988 = stepper_driver.Stepper(stepper_driver.A4988);
```

Grazfather:
Build failure is fixed in #364

mattnite (approved):

mattnite:
@Grazfather you need to include the stepper directory in the paths for build.zig.zon

PR #360, opened 2025-01-24
Move time types (Absolute, Duration, Deadline) from rp2xxx HAL into mdf

Grazfather:
Pulled this out from #359

* `Absolute`, `Duration`, and `Deadline` into `framework.time` (not clock since it deals with time)
* rewrite `rp2xxx.hal.time` in terms of those types

### DIFF
--- a/drivers/framework.zig
+++ b/drivers/framework.zig
@@ -1,6 +1,7 @@
 //!
 //! The driver framework provides device-independent drivers for peripherials supported by MicroZig.
 //!
+const std = @import("std");
 
 pub const display = struct {
     pub const ssd1306 = @import("display/ssd1306.zig");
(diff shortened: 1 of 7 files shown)


drivers/framework.zig
```diff
@@ -42,6 +43,128 @@ pub const wireless = struct {
...
+            else
+                null);
+        }
+
+        /// Returns `true` if the deadline can be reached.
+        pub fn can_be_reached(deadline: Deadline) bool {
+            return (deadline.timeout != disabled_sentinel);
+        }
+
+        /// Since this takes 'now', rename to is_reached_by?
```
Grazfather, line 146: probably should do this, I do it above in `Absolute`.

Would be nice if the calls that use this didn't have to call `time.get_time_since_boot()` (or equivalent), but I am can't think of a nice way to do that. Maybe the rp2xxx hal can make a wrapping type that adds methods?

mattnite (changes requested):
Please move this to the `core` module please!

mattnite (approved):

PR #367, opened 2025-01-27
Fix usb_cdc on rp2350

patryk4815:


port/raspberrypi/rp2xxx/src/hal/usb.zig
```diff
@@ -360,8 +360,8 @@ pub fn F(comptime config: UsbConfig) type {
             var setup_packet: [8]u8 = .{0} ** 8;
             const spl: u32 = peripherals.USB_DPRAM.SETUP_PACKET_LOW.raw;
             const sph: u32 = peripherals.USB_DPRAM.SETUP_PACKET_HIGH.raw;
-            _ = rom.memcpy(setup_packet[0..4], std.mem.asBytes(&spl));
-            _ = rom.memcpy(setup_packet[4..8], std.mem.asBytes(&sph));
+            @memcpy(setup_packet[0..4], std.mem.asBytes(&spl));
```
haydenridd, line 363: Does this crash the chip per the comment in `rom.zig`?

haydenridd:
Good work! There's a question about whether `rom.zig` should really even be used on the RP2350 given they did away with rom functions for performance purposes. It makes sense though to allow some of those aliases to still "work" on the RP2350 for easier compatibility with the old HAL.

port/raspberrypi/rp2xxx/src/hal/usb.zig
patryk4815: This part don't crash chip because they are aligned, so I could remove this changes from PR.

But this will crash, when changed to `@memcpy`:
```
_ = rom.memcpy(ep.data_buffer[0..buffer.len], buffer);
```

patryk4815:
> Good work! There's a question about whether `rom.zig` should really even be used on the RP2350 given they did away with rom functions for performance purposes. It makes sense though to allow some of those aliases to still "work" on the RP2350 for easier compatibility with the old HAL.

I have no idea, people with more knowledge must decide, some functions from `rom.zig` maybe be still useful on RP2350
image

port/raspberrypi/rp2xxx/src/hal/usb.zig
patryk4815: I think new zig 0.14.0 should work fine with `@memcpy` https://github.com/ziglang/zig/blob/183bb8b08403ab594cdc543c2db2503e7eac0052/lib/compiler_rt/memcpy.zig#L4
now they have code with alignment 🤔

haydenridd: Ahhh I understand now, the rom function doesn't obey alignment. All good!

haydenridd (approved):
Thanks for the good work! A future refactor of `rom.zig` is probably in order eventually, but this is to get USB working so let's merge!

mattnite:
Thank you @patryk4815 for the great contribution and @haydenridd for the review!

Issue #355, opened 2025-01-19
Interrupt Refactor


tact1m4n3:
Hey!
What do you guys think of these changes?

### Remove `VectorTableOptions` from start.zig

Would allow for more freedom regarding what kind of handlers can be specified by the user. Leave each cpu implementation to decide what handlers it wants.

```diff
- pub const VectorTableOptions = if (@hasDecl(microzig.chip, "VectorTable")) blk: {
-    const VectorTable = microzig.chip.VectorTable;
-    const fields_with_default = fields_with_default: {
-        var count = 0;
-        for (@typeInfo(VectorTable).Struct.fields) |field| {
-            if (field.default_value != null)
-                count += 1;
-        }
-
-        break :fields_with_default count;
-    };
-
-    var fields: [fields_with_default]std.builtin.Type.StructField = undefined;
-    var idx = 0;
-    for (@typeInfo(VectorTable).Struct.fields) |field| {
-        if (field.default_value == null)
-            continue;
-
-        fields[idx] = field;
-        idx += 1;
-    }
-
-    break :blk @Type(.{
-        .Struct = .{
-            .fields = &fields,
-            .layout = .auto,
-            .decls = &.{},
-            .is_tuple = false,
-        },
-    });
-} else struct {};
```

### Use `app.interrupts` for the user to specify interrupts instead of `app.microzig_options.interrupts`

```diff
+    pub const vector_table: VectorTable = blk: {
+        var tmp: VectorTable = .{
+            .initial_stack_pointer = microzig.config.end_of_stack,
+            .Reset = .{ .C = microzig.cpu.startup_logic._start },
+        };
+
+        if (@hasDecl(app, "interrupts")) {
+            for (@typeInfo(@TypeOf(app.interrupts)).Struct.fields) |field| {
+                if (@hasField(Interrupt, field.name)) {
+                    @field(tmp, field.name) = @field(app.interrupts, field.name);
+                }
+            }
+        }
+        break :blk tmp;
+    };

-pub const vector_table: VectorTable = blk: {
-    var tmp: VectorTable = .{
-        .initial_stack_pointer = microzig.config.end_of_stack,
-        .Reset = .{ .C = microzig.cpu.startup_logic._start },
-    };
-
-    if (@hasDecl(root, "microzig_options")) {
-        for (@typeInfo(root.VectorTableOptions).Struct.fields) |field|
-            @field(tmp, field.name) = @field(root.microzig_options.interrupts, field.name);
[...]

PR #369, opened 2025-01-29
Interrupt refactor

tact1m4n3:
* interrupt refactor
* implement more functionality regarding interrupts on cortex m
* implement more functionality regarding interrupts on hazard3 (rp2350 riscv)
* csrs on hazard3
* add interrupt list to regz

#355


examples/raspberrypi/rp2xxx/src/interrupts.zig
```diff
@@ -0,0 +1,67 @@
+const std = @import("std");
+const microzig = @import("microzig");
+const rp2xxx = microzig.hal;
+const time = rp2xxx.time;
+
+const led = rp2xxx.gpio.num(0);
+const uart = rp2xxx.uart.instance.num(0);
+const baud_rate = 115200;
+const uart_tx_pin = rp2xxx.gpio.num(12);
+const uart_rx_pin = rp2xxx.gpio.num(13);
```
Grazfather, line 10: We don't see this since we are just using the uart for logging

examples/raspberrypi/rp2xxx/src/interrupts.zig
```diff
@@ -0,0 +1,67 @@
...
+    const Duration = microzig.drivers.time.Duration;
+    const current = time.get_time_since_boot();
+    const target = current.add_duration(Duration.from_us(us));
+    microzig.chip.peripherals.TIMER0.ALARM0.write_raw(@intCast(@intFromEnum(target) & 0xffffffff));
+}
+
+pub fn main() !void {
+    // init uart logging
+    inline for (&.{ uart_tx_pin, uart_rx_pin }) |pin| {
+        pin.set_function(.uart_first);
```
Grazfather, line 42: Isn't this enum RP2350 only?

core/src/cpus/cortex_m.zig
```diff
@@ -15,6 +20,42 @@ pub fn disable_interrupts() void {
...
+            @compileError("can't enable exception: " ++ interrupt_name);
+        }
+    } else {
+        @compileError("interrupt not found: " ++ interrupt_name);
+    }
+}
+
+pub fn disable(comptime interrupt: anytype) void {
+    const interrupt_name = @tagName(interrupt);
+    if (@hasField(Interrupt, interrupt_name)) {
```
Grazfather, line 39: could probably abstract these checks out to some sort of 'interrupt is valid' comptime function to reduce duplicated logic.

core/src/interrupt.zig
```diff
@@ -60,17 +58,9 @@ const CriticalSection = struct {
     }
 };
 
-// TODO: update with arch specifics
-pub const Handler = extern union {
-    C: *const fn () callconv(.C) void,
-    Naked: *const fn () callconv(.Naked) void,
-    // Interrupt is not supported on arm
-};
+// TODO: remove this once the vector table uses it's own implementation
```
Grazfather, line 61: typo `it's` -> `its`

port/raspberrypi/rp2xxx/src/cpus/hazard3.zig
```diff
@@ -57,69 +123,411 @@ pub const startup_logic = struct {
...
+    pub const dpc = CSR(0x7B1, u32);
+    pub const dscratch = CSR(0x7B2, u32);
+
+    // hazard3 specific
+    pub const meiea = CSR(0xbe0, u32);
+
+    // TODO: maybe this should be relocated somewhere else
+    pub fn CSR(addr: u24, T: type) type {
+        const size = @bitSizeOf(T);
+        if (size != 32)
```
Grazfather, line 430: Why do we have this? Are there cases where it might be e.g. i32?

Grazfather (commented):
I just read this over to see if I understood it. It looks good. I don't know riscv or hazard3 very well. Added a few comments/questions.

examples/raspberrypi/rp2xxx/src/interrupts.zig
tact1m4n3: Yes it is. I will fix that :))

port/raspberrypi/rp2xxx/src/cpus/hazard3.zig
tact1m4n3: It can be any type of size 32 bits, including packed structs

examples/raspberrypi/rp2xxx/src/interrupts.zig
tact1m4n3: Could you clarify a bit what you mean?

core/src/start.zig
```diff
@@ -59,6 +28,7 @@ pub const Options = struct {
             _ = args;
         }
     }.log,
+    interrupts: microzig.cpu.InterruptOptions = .{},
```
Grazfather, line 31: shouldn't this just be `InterruptOptions` because of your comptime `if hasDecl` above?

core/src/utilities.zig
```diff
@@ -188,6 +189,60 @@ pub fn Slice_Vector(comptime Slice: type) type {
     };
 }
 
+pub fn max_enum_tag(T: type) @typeInfo(T).Enum.tag_type {
+    var max_tag: @typeInfo(T).Enum.tag_type = 0;
```
Grazfather, line 193: Could maybe do something like
```zig
        if (@typeInfo(T) != .Enum) {
            @compileError("max_enum_tag expects an enum type");
        }
```
Also, is it possible that the enum has a max value lower than 0? If so, then this would incorrectly think it's 0. We could check and iensure that there is at least 1 enum field, and initialize the max tag to the value of the first.

core/src/utilities.zig
```diff
@@ -188,6 +189,60 @@ pub fn Slice_Vector(comptime Slice: type) type {
...
+    for (@typeInfo(T).Enum.fields) |field| {
+        if (field.value > max_tag) {
+            max_tag = field.value;
+        }
+    }
+    return max_tag;
+}
+
+pub fn GenerateInterruptEnum() type {
+    var fields: [microzig.chip.INTERRUPTS.len]std.builtin.Type.EnumField = undefined;
```
Grazfather, line 203: Similar to above, maybe we bail out if INTERRUPTS is empty. But maybe this works fine with 0 fields?

core/src/utilities.zig
```diff
@@ -188,6 +189,60 @@ pub fn Slice_Vector(comptime Slice: type) type {
...
+
+    for (&fields, microzig.chip.INTERRUPTS) |*field, interrupt| {
+        field.* = .{
+            .name = interrupt.name,
```
Grazfather, line 207: Could there accidentally be duplicated names? I am not sure what would happen in that case.

core/src/utilities.zig
```diff
@@ -188,6 +189,60 @@ pub fn Slice_Vector(comptime Slice: type) type {
...
+    return @Type(.{ .Enum = .{
+        .tag_type = i16,
+        .fields = &fields,
+        .decls = &.{},
+        .is_exhaustive = true,
+    } });
+}
+
+pub fn GenerateInterruptOptions(HandlerType: type, sources: anytype) type {
+    const sources_type_info = @typeInfo(@TypeOf(sources)).Struct;
```
Grazfather, line 221: could add check to ensure sources is a struct

Grazfather: could also add a check that `HandlerType` is a function/pointer

core/src/utilities.zig
Grazfather: Could use `ComptimeStringMap` to check for dupes

core/src/utilities.zig
```diff
@@ -188,6 +189,60 @@ pub fn Slice_Vector(comptime Slice: type) type {
...
+    var ret_fields: []const std.builtin.Type.StructField = &.{};
+
+    for (sources_type_info.fields) |sources_field| {
+        for (@typeInfo(@field(sources, sources_field.name)).Enum.fields) |enum_field| {
+            ret_fields = ret_fields ++ .{.{
+                .name = enum_field.name,
```
Grazfather, line 227: Similar potential issue with dupes here.

examples/raspberrypi/rp2xxx/build.zig
```diff
@@ -67,6 +68,12 @@ pub fn build(b: *std.Build) void {
             .name = b.fmt("pico2_{s}", .{example.name}),
             .file = example.file,
         }) catch @panic("out of memory");
+
+        available_examples.append(.{
+            .target = mb.ports.rp2xxx.boards.raspberrypi.pico2_riscv,
+            .name = b.fmt("pico2_{s}", .{example.name}),
```
Grazfather, line 74: this will collide with the pico2 arm examples

Grazfather (commented):
Have a few suggestions in utilities, but those could probably be a followup. I think I found a bug in the rp2xxx build.zig though, with name collisions.

core/src/start.zig
tact1m4n3: Oh... of course :))

examples/raspberrypi/rp2xxx/src/interrupts.zig
tact1m4n3: I think we should take care of this in another pr since most examples do this rn

examples/raspberrypi/rp2xxx/build.zig
tact1m4n3: nice catch :))

examples/raspberrypi/rp2xxx/src/interrupts.zig
Grazfather: works for me

Grazfather:
Looks good to me but I am not able to approve on my own (plus all the builds are failing still)

core/src/utilities.zig
tact1m4n3: It's a sort of bug in regz where some interrupts appear multiple times in the database. I've made a temporary solution for this in regz but it should be revisited by someone with more knowledge of the code.

mattnite (approved):
High quality patch, love this direction

Issue #370, opened 2025-01-31
Zig 0.14 support: Add .ARM.extab to generated ARM linker.ld scripts


marnix:
Per https://github.com/ziglang/zig/issues/22685#issuecomment-2627678031, Zig 0.14 has enabled 'unwind tables' for ARM embedded (and most other platforms).  Therefore it looks like we need to update `tools/generate_linker_script.zig`, as part of the upgrade to Zig 0.14 (being done on the current [zig-master](https://github.com/ZigEmbeddedGroup/microzig/tree/zig-master) branch).

```diff
diff --git a/tools/generate_linker_script.zig b/tools/generate_linker_script.zig
index 308b055e51..775892c0f6 100644
--- a/tools/generate_linker_script.zig
+++ b/tools/generate_linker_script.zig
@@ -107,6 +107,10 @@
 
         switch (program_args.cpu_arch) {
             .arm, .thumb => try writer.writeAll(
+                \\  .ARM.extab : {
+                \\      *(.ARM.extab* .gnu.linkonce.armextab.*)
+                \\  } >flash0
+                \\  
                 \\  .ARM.exidx : {
                 \\      *(.ARM.exidx* .gnu.linkonce.armexidx.*)
                 \\  } >flash0
```

I have no idea whether the above is actually correct, but it seems to work for me.  (At the time of writing, I did not try to actually debug and look at any stack traces yet...)

I have no idea whether something similar would be needed for RISCV, and/or for other embedded platforms.

marnix:
Actually this change can of course already be made on current master, but it's needed for Zig 0.14.

Grazfather:
Done in #394 and #373

PR #373, opened 2025-02-01
Enable 'unwind_tables' for all platforms (for #370)

marnix:
For ARM-based embedded firmware this requires
explicit declaration of `ARM.extab` sections in linker.d, otherwise the exception tables can incorrectly end up in RAM. (See https://github.com/ziglang/zig/issues/22685.)

(The `.unwind_tables = true` line must be removed for Zig 0.14, which already has this enabled for basically all platforms.)

Implements part of #370 already on Zig 0.13.0.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -462,6 +462,7 @@ pub fn MicroBuild(port_select: PortSelect) type {
                     .linkage = .static,
                     .root_source_file = mb.core_dep.path("src/start.zig"),
                     .strip = options.strip,
+                    .unwind_tables = true,
                 }),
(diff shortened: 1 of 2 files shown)


mattnite (approved):

ghostiam:
Hello, this commit breaks the build for RISC-V, resulting in a 400MB firmware file for the CH32V003. The only workaround is to override with:  
```
<firmware>.artifact.root_module.unwind_tables = false
```

Could you please fix this issue or add this override to the RISC-V example file so that others don’t have to deal with this problem?

PR #377, opened 2025-02-03
Add Support for non-constant stride field arrays. (STM32)

taylorh140:
Just trying to button up as many corner cases as possible on the stm32 register generations.


port/stmicro/stm32/src/chips/all.zig
```diff
@@ -350862,30 +351112,102 @@ pub const types = struct {
...
-                    EEPOL: u1,
+                    @"EEPOL[0]": u1,
                     /// External Event X Sensitivity
-                    EESNS: u2,
+                    @"EESNS[0]": u2,
                     /// External Event X Fast Mode
-                    EEFAST: u2,
-                    padding: u25,
+                    @"EEFAST[0]": u2,
+                    // skipped overlapping field EESRC[1] at offset 6 bits
```
taylorh140, line 351122: I notice a new issue here but i think its an upstream error. you can see the high resolution timer datasheet here: https://www.st.com/resource/en/reference_manual/rm0364-stm32f334xx-advanced-armbased-32bit-mcus-stmicroelectronics.pdf

mattnite (approved):
Generated code looks good to me!

PR #378, opened 2025-02-03
regz+rp2xxx: `add_interrupt` patch

tact1m4n3:
- `add_interrupt` patch
- add PROCx and SPAREx irqs rp2350 (using patch)
- fix bug in regz where interrupts for derived peripherals are skipped
- use hard float implementation for rp2350-arm since it has a fpu


mattnite (approved):
I love to see it

PR #390, opened 2025-02-17
Update gpio.zig

henryrgithub:
embassy-rs/stm32-data-generated from V1 -> V2 renames GPIO mode registers from len2 array of 4bit registers (CR[0],CR[1]) to two seperate 4bit registers (MODER, OTYPER).

I'm working on porting the F103 HAL to a G432 and came across this change when trying to get a blinky compiling. It compiles with this change, I'm hoping to get it running on the G432 today or tomorrow. I'll try and get it running on an F103 bluepill sometime by the middle of next week.


henryrgithub:
Well, whoops. I thought V1 vs V2 was a HAL update, but it's instead STM spec update. F103 uses V1, G431 uses V2. Makes sense why F103 wasn't whining about CR not existing in V2, it doesn't use V2.

PR #391, opened 2025-02-17
rp2040: Add spi slave example

Grazfather:
Add a SPI slave example.

This one behaves like the example linked from the spi master example, but doesn't work with the spi master example.

That example switches between 12 and 8 bit mode, which is cool, but in practice is hard to synchronize.

I'd propose dropping that part from the example so that spi master and spi slave can actually behave with each other.

Tested by removing the 12-bit part of the spi master code, as well as against the spi master pico-examples code.

TODO:
* [x] Test spi slave on rp2350 (spi master already tested on this chip)
* [x] Move files from rp2040-only directory


examples/raspberrypi/rp2xxx/src/rp2040_only/spi_slave.zig
```diff
@@ -0,0 +1,76 @@
...
+pub fn main() !void {
+    // Set pin functions for CS, SCK, RX
+    const csn = gpio.num(CS_PIN);
+    const mosi = gpio.num(RX_PIN);
+    const sck = gpio.num(SCK_PIN);
+    inline for (&.{ csn, mosi, sck }) |pin| {
+        pin.set_function(.spi);
+    }
+
+    inline for (&.{uart_tx_pin}) |pin| {
```
Grazfather, line 38: `for` was for when I was setting multiple pins.

examples/raspberrypi/rp2xxx/src/rp2040_only/spi_slave.zig
```diff
@@ -0,0 +1,76 @@
...
+
+    // Back to 8 bit mode
+    try spi.apply(.{
+        .clock_config = rp2xxx.clock_config,
+        .data_width = .eight,
+    });
+
+    while (true) {
+        spi.read_blocking(u8, 0, &in_buf_eight);
+        std.log.info("Got: {s}", .{in_buf_eight});
```
Grazfather, line 73: I changed master to send ascii. If we want it to send bytes with values 00..ff (like the pico-examples) then we should add a hex dump function.

examples/raspberrypi/rp2xxx/src/rp2040_only/spi_host.zig
```diff
@@ -10,57 +10,67 @@ const peripherals = microzig.chip.peripherals;
...
-    try spi.apply(.{
-        .clock_config = rp2xxx.clock_config,
-        .data_width = .twelve,
-    });
-    var out_buf_twelve: [BUF_LEN]u12 = .{ 0xAA, 0xBB, 0xCC, 0xDD } ** (BUF_LEN / 4);
-    var in_buf_twelve: [BUF_LEN]u12 = undefined;
-    csn.put(0);
-    spi.transceive_blocking(u12, &out_buf_twelve, &in_buf_twelve);
-    csn.put(1);
+    // // 12 bit data words
```
Grazfather, line 52: I think that we should remove this so that our master example is compatible with our slave example.

examples/raspberrypi/rp2xxx/src/rp2040_only/spi_host.zig
```diff
@@ -10,57 +10,67 @@ const peripherals = microzig.chip.peripherals;
...
     csn.set_function(.sio);
     csn.set_direction(.out);
     csn.put(1);
 
     // These will change depending on which GPIO pins you have your SPI device routed to
-    const hodi = rp2xxx.gpio.num(3);
-    const hido = rp2xxx.gpio.num(4);
-    const sck = rp2xxx.gpio.num(2);
-    inline for (&.{ hodi, hido, sck }) |pin| {
+    const miso = rp2xxx.gpio.num(MISO_PIN);
```
Grazfather, line 34: don't need miso if we are not transceiving.

examples/raspberrypi/rp2xxx/src/spi_master.zig
```diff
@@ -0,0 +1,42 @@
...
+const SCK_PIN = 18;
+// NOTE: rp2xxx doesn't label pins as MOSI/MISO. Instead a pin is always for
+// either receiving or transmitting SPI data, no matter whether the chip is in
+// master or slave mode.
+const RX_PIN = 16;
+
+// Communicate with another RP2040 over spi
+pub fn main() !void {
+    // Set pin functions for CS, SCK, RX
+    const csn = rp2xxx.gpio.num(CS_PIN);
```
Grazfather, line 22: I noticed that a rp2040 slave does NOT like when the master manually manages the CS pin, so I actually set its function here (vs. the old example)

examples/raspberrypi/rp2xxx/src/spi_master.zig
```diff
@@ -0,0 +1,42 @@
...
+    const sck = rp2xxx.gpio.num(SCK_PIN);
+    inline for (&.{ csn, miso, sck }) |pin| {
+        pin.set_function(.spi);
+    }
+
+    // 8 bit data words
+    try spi.apply(.{
+        .clock_config = rp2xxx.clock_config,
+        .data_width = .eight,
+        .baud_rate = 500_000,
```
Grazfather, line 33: can drop

port/raspberrypi/rp2xxx/src/hal/spi.zig
```diff
@@ -162,6 +161,17 @@ pub const SPI = enum(u1) {
         spi_regs.SSPCPSR.modify(.{ .CPSDVSR = 0 });
     }
 
+    pub fn set_slave(spi: SPI, slave: bool) void {
+        const regs = spi.get_regs();
+        // Disable SPI
+        regs.SSPCR1.modify(.{ .SSE = 0 });
```
Grazfather, line 167: I wonder whether we really need this, since we are writing to the same register, we can do this atomically.

mattnite: I imagine you could probably set MS and enable at the same time. But disabling and enabling doesn't make sense as something you could do atomically.

mattnite (approved):

PR #392, opened 2025-02-17
Merge main into zig-master and fix rp2350 build

tact1m4n3:


mattnite:
@tact1m4n3 still not quite there history wise, I'm still seeing commits that aren't yours in this PR, maybe try restarting and cherry picking commits

PR #398, opened 2025-03-01
Rename CSELR fields match ST's documentation

ParkerHitch:
Hi all!

Been trying out (and loving) microzig, but when working with DMA on my STM32F091RC, I got a little mixed up when modifying the CSELR DMA peripheral register.

In the [RM0091 reference](https://www.st.com/resource/en/reference_manual/rm0091-stm32f0x1stm32f0x2stm32f0x8-advanced-armbased-32bit-mcus-stmicroelectronics.pdf) section 10.6.7 (pg. 212) shows the fields available in the CSELR register.
The fields of that register are labeled C1S through C7S, with four bytes of padding in the MSB, and the fields in the microzig struct are CS[0] through CS[7], which leaves no padding, and (in my opinion) is slightly confusing, since if you just use the numbers you see in the manual you'd be off by 1, since they are 1-indexed. Since the microzig struct allows you to write to padding bits, and the naming of all other registers and fields I have used is very consistent with that in the manual, I figured this was an oversight.

This pull request proposes renaming the fields of the CLSER register in the microzig bdma_v2 struct, from CS[0]-CS[7] to CS[1]-CS[7], and, of course, adding a u4 padding field.

Let me know your thoughts!

P.S. I was kinda worried the naming might have been intentional, so I double checked, and microzig's bdma_v2 struct is used in the following families from st. All these reference manuals show 7 fields in DMA_CSLER:
- F0 (see previous link)
- [L0x0](https://www.st.com/resource/en/reference_manual/rm0451-ultralowpower-stm32l0x0-advanced-armbased-32bit-mcus-stmicroelectronics.pdf)
- [L0x1](https://www.st.com/resource/en/reference_manual/rm0377-ultralowpower-stm32l0x1-advanced-armbased-32bit-mcus-stmicroelectronics.pdf)
- [L0x2](https://www.st.com/resource/en/reference_manual/rm0376-ultralowpower-stm32l0x2-advanced-armbased-32bit-mcus-stmicroelectronics.pdf)
- [L0x3](https://www.st.com/resource/en/reference_manual/rm0367-ultralowpower-stm32l0x3-advanced-armbased-32bit-mcus-stmicroelectronics.pdf)
- LM4 series which has [RM0351](https://www.st.com/resource/en/reference_manual/rm0351-stm32l47xxx-stm32l48xxx-stm32l49xxx-and-stm32l4axxx-advanced-armbased-32bit-mcus-stmicroelectronics.pdf) and [RM0394](https://www.st.com/resource/en/reference_manual/rm0394-stm32l41xxx42xxx43xxx44xxx45xxx46xxx-advanced-armbased-32bit-mcus-stmicroelectronics.pdf)

Lowkey probably didn't need to do all that but here we are.


marnix:
Unfortunately, this is a generated file, by a tool under `tools` (which uses regz, also in this repository), from a Rust projects's JSON files, which are in turn generated from manually maintained YAML files, which (I think) initially were based on SVD files from ST.

(Can't add links now.)

We chose zero-based indices for simplicity -- as far as I know the JSON files have no information that a specific "array" field is intended for (e.g) peripherals 1,2,3,... as opposed to (e.g.) pins 0,1,2,....

mattnite:
Hey! @marnix is right, we cannot accept this patch because the code is generated, and the data is from an upstream project.

This is a problem I've thought about before, and I think we could provide a nice API with an integer backed, non-exhaustive enum. In short we could create a type that could be used as an index, but is only valid for a set of values. You'd either have to manifest the 0 index value in memory, or wrap access with a function.

PR #401, opened 2025-03-06
avr: update example to use core.experimental

oliverpool:
This PR updates the avr example to use `core.experimental`.

In order for the code to compile and run, it needed minor adjustments.


mattnite (approved):
Thanks!

PR #403, opened 2025-03-07
rp2350: Fix GlobalConfig for uart, rename func to match rp2040

Grazfather:
The RP2350 adds an extra value for `FUNCSEL` for the GPIOs. This value, for some GPIOs, is a second uart use. We named the first `.uart_first` and the second `.uart_second`.

This naming made it so that we had to select either `.uart` (on rp2040) or `.uart_first` (on rp2350), which made the code messier than it needed to be.

Additionally, if using the `GlobalConfiguration` with a gpio set to a uart function, it would not work since our code naively set it to `.uart`, which didn't exist on rp2350.

Making the name the same for the functionality fixes the GlobalConfiguration.

Additionally, I cleaned up a bunch of examples and tested them on rp2350. The ones that work, I moved to the chip-agnostic-examples.

The GlobalConfiguration still will not work if trying to set a pin for a function that needs the FUNCSEL set to the second uart functionality. That can be done in a follow up, and I added TODOs appropriately.


port/raspberrypi/rp2xxx/src/hal/pins.zig
```diff
@@ -509,6 +512,7 @@ pub const GlobalConfiguration = struct {
                     pin.set_pull(.disabled);
                     pin.set_input_enabled(false);
                 } else if (comptime func.is_uart_tx() or func.is_uart_rx()) {
+                    // TODO: Handle pins that can be used as uart on their secondary pinsel
```
Grazfather, line 515: s/pinsel/funcsel

port/raspberrypi/rp2xxx/src/hal/pins.zig
```diff
@@ -509,6 +512,7 @@ pub const GlobalConfiguration = struct {
                     pin.set_pull(.disabled);
                     pin.set_input_enabled(false);
                 } else if (comptime func.is_uart_tx() or func.is_uart_rx()) {
+                    // TODO: Handle pins that can be used as with a alternate uart function
                     pin.set_function(.uart);
```
Grazfather, line 516: This would fail on rp2350 before since it only had `.uart_first` and `.uart_second`

mattnite (approved):

PR #414, opened 2025-03-22
examples: rp2xxx: interrupts: Cleanup and clarify cleaning behaviour

Grazfather:
I was confused how a toggle with 0 worked, since it should just keep the value the same.

Turns out that this register is 'write-to-clear' so toggle with 0 just rewrites the same value back, which happens to work.

This PR simply clarifies what's happening by using `modify`


tact1m4n3 (approved):
LGTM

Issue #315, opened 2024-12-15
Adding proper support for RISC-V


tact1m4n3:
Just some thoughts to improve the riscv support in microzig. :grinning: Maybe we could discuss this further in this issue.

### Problem:
Currently, it seems that for each riscv chip we are creating a new cpu implementation from scratch which creates a lot of code duplication.

### Solution:
All riscv cpus should depend on a generic riscv implementation so that csrs, interrupt related functions, generic vector table, etc don't have to be duplicated. Each cpu module can select what it wants from this generic implementation and reexport the definitions (if they are aplicable for that cpu).

tact1m4n3:
Solved by #503

PR #415, opened 2025-03-22
ESP32-C3 better support

tact1m4n3:
- elf2image in zig
- esp port refactor (very basic)
- add esp image format for the esp32_c3 target
- add esp image unit tests to ci
- esp image release step

NOTE: someone should update the deploy gitHub workflow to include the esp image release step


tools/esp_image/build.zig.zon
```diff
@@ -0,0 +1,18 @@
+.{
+    .name = .tools_esp_image,
+    .fingerprint = 0x8c3a8bf1bd71f47d,
+    .version = "0.0.0",
+    .dependencies = .{
+        .clap = .{
+            .url = "git+https://github.com/Hejsil/zig-clap.git#e47028deaefc2fb396d3d9e9f7bd776ae0b2a43a",
```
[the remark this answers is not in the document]
tact1m4n3: I noticed that we don't do that for the other tools but sure I can put something like 0.1.0 there

tools/esp_image/README.md
```diff
@@ -0,0 +1 @@
+# esp image
```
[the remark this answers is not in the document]
tact1m4n3: Will do

port/espressif/esp/src/cpus/espressif-riscv.zig
```diff
@@ -2,115 +2,339 @@ const std = @import("std");
...
+    pub const tdata1 = CSR(0x7A1, u32);
+    pub const tdata2 = CSR(0x7A2, u32);
+    pub const tdata3 = CSR(0x7A3, u32);
+
+    pub const dcsr = CSR(0x7B0, u32);
+    pub const dpc = CSR(0x7B1, u32);
+    pub const dscratch = CSR(0x7B2, u32);
+
+    // TODO: maybe this should be relocated somewhere else
+    pub fn CSR(addr: u24, T: type) type {
```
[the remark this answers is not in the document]
tact1m4n3: This is a problem with the way we currently implement riscv cpus :) and I think it should be tackled in another pr (#315). For now I'll remove the duplicated code.

port/espressif/esp/src/cpus/espressif-riscv.zig
```diff
@@ -2,115 +2,339 @@ const std = @import("std");
...
 
-pub const startup_logic = struct {
-    comptime {
-        // See this:
-        // https://github.com/espressif/esp32c3-direct-boot-example
-
-        // Direct Boot: does not support Security Boot and programs run directly in flash. To enable this mode, make
-        // sure that the first two words of the bin file downloading to flash (address: 0x42000000) are 0xaedb041d.
-
-        // In this case, the ROM bootloader sets up Flash MMU to map 4 MB of Flash to
```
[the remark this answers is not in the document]
tact1m4n3: I added back the direct boot mode. For now there are two targets. I'll also add some info in the readme regarding how to flash the esp image.

examples/espressif/esp/README.md
```diff
@@ -1,4 +1,20 @@
...
-- [Blinky](src/blinky.zig) on [ESP32-C3-32S-Kit](https://www.waveshare.com/wiki/ESP-C3-32S-Kit)  
+- [Blinky](src/blinky.zig) on [ESP32-C3-32S-Kit](https://www.waveshare.com/wiki/ESP-C3-32S-Kit)
   Showcases how to do a simple RGB cycling.
+
+## How to flash the image onto the device
+
+- esp image
+
+```sh
+esptool --chip esp32c3 -b 460800 --before default_reset --after hard_reset write_flash \
```
Grazfather, line 11: nitpicking: Let's switch `-b` to `--baud` since we use the other long forms and I had to check to see what it meant.

examples/espressif/esp/README.md
```diff
@@ -1,4 +1,20 @@
...
+        0x0 bootloader.bin 0x8000 partition_table.bin 0x10000 zig-out/firmware/esp32_c3_blinky.bin
+```
+NOTE: you have to provide a [bootloader](https://docs.espressif.com/projects/esp-idf/en/stable/esp32c3/api-guides/bootloader.html) and a [partition table](https://docs.espressif.com/projects/esp-idf/en/stable/esp32c3/api-guides/partition-tables.html).
+
+- direct boot image
+
+```sh
+esptool.py --chip esp32c3 -b 460800 --before default_reset --after hard_reset write_flash 0x0 zig-out/firmware/esp32_c3_direct_boot_blinky.bin
```
Grazfather, line 19: baud and break up line please

Grazfather: also this is esptool.py but above is not.

port/espressif/esp/build.zig
```diff
@@ -26,32 +28,61 @@ pub fn init(dep: *std.Build.Dependency) Self {
...
                 // sram 1, data bus
-                .{ .kind = .ram, .offset = 0x3FC8_0000, .length = 0x0005_0000 },
+                .{ .kind = .ram, .offset = 0x3FC8_0000, .length = 0x0006_0000 },
             },
         },
         .hal = hal,
         .linker_script = b.path("esp32_c3.ld"),
     };
 
+    const chip_esp32_c3_direct_boot: microzig.Target = .{
```
Grazfather, line 54: Couldn't we just derive this from the esp format one? Seems that that doesn't support overriding the cpu_module_file, but that should be an easy fix

examples/espressif/esp/README.md
```diff
@@ -1,4 +1,20 @@
...
+- esp image
+
+```sh
+esptool --chip esp32c3 -b 460800 --before default_reset --after hard_reset write_flash \
+        0x0 bootloader.bin 0x8000 partition_table.bin 0x10000 zig-out/firmware/esp32_c3_blinky.bin
+```
+NOTE: you have to provide a [bootloader](https://docs.espressif.com/projects/esp-idf/en/stable/esp32c3/api-guides/bootloader.html) and a [partition table](https://docs.espressif.com/projects/esp-idf/en/stable/esp32c3/api-guides/partition-tables.html).
+
+- direct boot image
+
```
Grazfather, line 17: Maybe add link to https://github.com/espressif/esp32c3-direct-boot-example

tools/esp_image/README.md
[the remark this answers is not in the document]
Grazfather: Thank you

tools/esp_image/build.zig.zon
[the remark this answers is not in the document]
Grazfather: Yes I think that's probably a bad practice of ours. We could use the git url with a tag, or just the archive link.

Grazfather:
Tested both output binaries plus the elf files in a debugger and everything works great.

port/espressif/esp/build.zig
tact1m4n3: I thought about that, the problem is that derive already takes in a chip (to which the `cpu_module_file`) belongs to.

tact1m4n3: > A better solution imho would be to move `cpu_module_file` to be a property of a `Target` instead. And then add it to `derive`. But this would require updating multiple build.zigs :).

Does this sound good?

Grazfather: Maybe we should just file and issue and do it later :)

tact1m4n3: Sure

Grazfather: I think it makes sense -- It isn't CPU specific, since it depends on the bootloader, the load address, etc.

tact1m4n3: Agreed. In this case there are two cpu module files required because of the two different boot methods. In the direct boot one you have to copy both data and zero out bss, but in the esp image format case the bootloader copies the data at the right location so you only have to zero out bss.

Grazfather: #420

Grazfather (approved):

PR #417, opened 2025-03-25
(Re)Add support for multiple interrupt calling conventions

tact1m4n3:
- (Re)Add support for multiple interrupt calling conventions on cortex_m and hazard3
- use `riscv_interrupt` calling conventions for core interrupts on hazard3 and keep the c calling convention for external interrupts


tact1m4n3:
> I would personally prefer if the CSR bug fix were its own PR so that it'll survive as its own commit in the git history

Done 👍

Grazfather:
in #418 for posterity.

Thank you!

PR #424, opened 2025-04-02
Spin lock management

curuvar:
I wrote this to allow managing spinlocks with the lock-defer-unlock pattern.
Also added code to clear spinlocks on soft reset as the RP2040 doesn't do that automatically.


mattnite (approved):
This is great! Thank you

Issue #188, opened 2024-05-09
RP2040 Interrupt Support


MatthiasPortzel:
[Placeholder issue documenting what needs to happen for the RP2040 HAL to support interrupts.]

MatthiasPortzel:
I just got an example program running, which I'm going to include here as an idea of what is involved.

```zig
const microzig = @import("microzig");
const hal = microzig.hal;

// TODO: We can pull this, and the other addresses, from the HAL
const IOBANK0 = 0x40014000;

// I don't actually know if this needs to be callconv C
pub fn gpio_handler () callconv(.C) void  {
    // Acknowledge/clear the interrupt. With an edge-triggered interrupt,
    //  when missing the acknowledge, the interrupt triggers repeatedly forever
    // We're "listening" for falling-edge events on pin 18
    // We need to write a 1 to bit 10 of INTR2
    // Search for "INTR0 Register" in the rp2040 datasheet
    const INTR2: *volatile u32 = @ptrFromInt(IOBANK0 + 0x0f8);
    INTR2.* = (1 << 10);

    // Turn the on-board light on the Pico on and then off
    hal.gpio.num(25).put(1);
    hal.time.sleep_ms(1000);
    hal.gpio.num(25).put(0);
    hal.time.sleep_ms(200);
}

// Insert our handler into the vector table
// Microzig looks for this "microzig_options" struct and inserts the address
// of our handler function into the vector table when linking the program
pub const microzig_options = .{
    .interrupts = .{
        .IO_IRQ_BANK0 = microzig.interrupt.Handler {
            .C = gpio_handler
        }
    }
};

const pin_config = hal.pins.GlobalConfiguration {
    // on board LED
    .GPIO25 = .{
        .name = "led_1",
        .direction = .out,
    },
    // Button
    .GPIO18 = .{
        .name = "button_1",
        .direction = .in,
        .pull = .down,
    },
};


pub fn main() noreturn {
    // The intention is that you can short this low by pressing a button,
    //  and when releasing the button, the interrupt will fire
    _ = pin_config.apply();

    // Enable the GPIO interrupt for the GPIO bank
    // 0xE000E100 is memory mapped NVIC_ISER; 13 is IO_IRQ_BANK0
    // Search for "NVIC_ISER" and "IO_IRQ_BANK0" in the rp2040 datasheet for more info
    const ISER: *volatile u32 = @ptrFromInt(0xE000E100);
    ISER.* = (1 << 13);

[...]

Issue #381, opened 2025-02-08
Stack grows towards data


jrg1k:
The generated linker script places .bss after .data:

```
ENTRY(microzig_main);

MEMORY
{
  flash0    (rx!w) : ORIGIN = 0x00000000, LENGTH = 0x0009D800
  ram0      (rw!x) : ORIGIN = 0x20000000, LENGTH = 0x00040000
  ram1      (rw!x) : ORIGIN = 0x40100000, LENGTH = 0x00004000
}

SECTIONS
{
  .text :
  {
     KEEP(*(microzig_flash_start))
     *(.text*)
  } > flash0

  .ARM.extab : {
      *(.ARM.extab* .gnu.linkonce.armextab.*)
  } >flash0

  .ARM.exidx : {
      *(.ARM.exidx* .gnu.linkonce.armexidx.*)
  } >flash0

  .data :
  {
     microzig_data_start = .;
     *(.sdata*)
     *(.data*)
     *(.rodata*)
     microzig_data_end = .;
  } > ram0 AT> flash0

  .bss (NOLOAD) :
  {
      microzig_bss_start = .;
      *(.bss*)
      *(.sbss*)
      microzig_bss_end = .;
  } > ram0

  .flash_end :
  {
      microzig_flash_end = .;
  } > flash0

  microzig_data_load_start = LOADADDR(.data);
}
```

By ensuring the stack is the lowest item in ram a stack overflow causes a busfault instead of overwriting data.

mattnite:
This is by design

jrg1k:
But the way it is currently the hardware doesn't necessarily detect it if it just silently overwrites data. It grows over .bss and .data.

mattnite:
Oh I see, we agree on how we'd want our systems to behave. Your first comment was not clear which you thought was the ideal behaviour.

jrg1k:
Yes I was a bit lazy creating the issue. The way I currently do this in my bare-metal project is to create a separate stack section:

```
    .stack (NOLOAD) :
    {
        *(.stack)
    } > RAM
```

Then I configure the stack in zig:

```zig
export var stack: [0x800]u8 align(8) linksection(".stack") = undefined;

export var vector_table: VectorTable align(4) linksection(".vector_table") = .{
    .initial_stack_pointer = @as([*]u8, &stack) + stack.len,
    .reset_handler = Reset_Handler,
};
```

Grazfather:
[...]

PR #427, opened 2025-04-03
RP2xxx RAM based vector table

Uthedris:
These changes move the vector table to RAM so `setInterruptHandler` and `setExceptionHandler` functions can be supported. (Issue #188)

~~This doesn't quite work as is.~~  
~~It works on arm but doesn't yet on rsic-v.~~

It works on RP2040, RP2350-arm, and RP2350-riscv.

~~I would also like to make it's inclusion optional, so that it can be eliminated if RAM is tight.  My thought was to add a `.ram_vtor` field to `microzig_options` so the call to `copyVectorTable` in hal.zig could be bypassed and the set functions disabled.~~

~~I didn't make that change yet since `microzig_options` is a platform independent struct, and I didn't know if it would be okay to add a more platform specific field to it.~~  

I added code to require `.platform = .{ .ram_vectors = true },` in `microzig_option` before the vectors are moved to RAM.  Currenly the value defaults to `false` (leave vectors in flash), but that is easily changed.  

Feedback welcome.

#


port/raspberrypi/rp2xxx/src/hal/irq.zig
```diff
@@ -139,3 +169,48 @@ pub fn disable(mask: Mask) void {
...
 }
+
+const VtorFunction = *const fn () callconv(.C) void;
+
+const nvic_interrupt_count = switch (chip) {
+    .RP2040 => 25,
+    .RP2350 => 51,
+};
+
+pub var ram_vtor: [16 + nvic_interrupt_count]VtorFunction align(256) = undefined;
```
Grazfather, line 180: Maybe we should let the linker decide where to place this.

With #381 we'd want to move the stack to before .data and .bss. In that case I think we'd probably want it right as the end of RAM (making it easy to peek at)

port/raspberrypi/rp2xxx/src/hal/irq.zig
```diff
@@ -139,3 +169,48 @@ pub fn disable(mask: Mask) void {
         },
     }
 }
+
+const VtorFunction = *const fn () callconv(.C) void;
```
Grazfather, line 173: We'd like it is this calling convention were not enforced, since we might want to write functions with `callconv(.naked)` for hand-written assemblers.

Grazfather, line 173: nit: I guess it still works, but zig moved these sort of enums to lower case
```suggestion
const VtorFunction = *const fn () callconv(.c) void;
```

Grazfather (commented):
> I would also like to make it's inclusion optional, so that it can be eliminated if RAM is tight. My thought was to add a .ram_vtor field to microzig_options so the call to copyVectorTable in hal.zig could be bypassed and the set functions disabled.

the `InterruptOptions` field in `microzig.Options` is taken from the CPU

curuvar:
> the `InterruptOptions` field in `microzig.Options` is taken from the CPU

Yes, but that field is automatically generated based on list is of available interrupts, and other code expects all its fields to be interrupt names.

curuvar:
I added a separate `platforms` item to `microzig_options` to allow a developer to choose whether to move vectors.

port/raspberrypi/rp2xxx/src/hal/irq.zig
curuvar: I had thought about that and didn't see much benefit.   One thing to consider is that the vector table needs to be on a 256 byte boundry.  I don't know if Zig is smart enough to put other variables in any space needed for padding; but if Zig is smart enough, putting the vectors in a separate section might be more wasteful.

Grazfather: Because it would waste space between the end of the table and the end of flash (if we put it add flash_end - 256)?

curuvar: Perhaps it's silly for me to worry about wasting 100 bytes on a chip with 264K.  I started my programming journey on machines like the PDP-8 (where you only had 4K) and developed multi-user interactive systems for the PDP-11 (a whopping 64K), so every byte was precious.

curuvar:
Improved the code.  Tested on RP2350 on both arm and riscv cores.   Supports optionally placing ram vectors in their on .section.

port/raspberrypi/rp2xxx/src/hal/irq.zig
Grazfather: I don't blame you. I was actually just curious.

examples/raspberrypi/rp2xxx/src/interrupts.zig
```diff
@@ -59,22 +70,9 @@ pub fn main() !void {
...
-                .riscv => {
-                    interrupt.enable(.MachineExternal);
-                    interrupt.external.enable(.TIMER0_IRQ_0);
-                },
-            }
-        },
-    }
+    timer.INTE.toggle(.{ .ALARM_0 = 1 });
+
+    interrupt.enable(timer_irq);
```
Grazfather, line 75: riscv doesn't need `interrupt.external.enable`?

curuvar: It does internally, but it is done as part of system setup in my hazard3.zig by the `_start_c` function.  

The goal was to make handing NVIC interrupts (the thing most likely needed in user code) identical between the riscv and arm cores.

tact1m4n3:
Could we rename `PlatformOptions` to `CpuOptions`? Maybe we could add interrupts under `CpuOptions` as well? This would allow to later add `HalOptions` and `BoardOptions`. `HalOptions` are useful for instance in the rp2xxx hal for `image_def`. We could also add a `std` field for all `std.Options`.

```zig
const microzig_options: microzig.Options = .{
    .std = .{
        .log_level = .debug,
    },
    .cpu = .{
        .ram_vectors = false,
        .interrupts = .{
            .TIMER = ...,
        },
    },
    .hal = .{
        ...
    },
};
```

Uthedris:
I changed PlatformOptions to CPU options as that makes sense.   

I can see the sense of putting `.interrupts` under `.cpu` too as they are also cpu specific. I have not made that change as it may have a significant ripple effect that I didn't want to take the time to track down before we decide what's best.

mattnite:
@Grazfather are you happy with this patch?

Grazfather (approved):
Yes looks good.

PR #428, opened 2025-04-03
Fixed race condition in RP2xxx i2c_slave

curuvar:
I discovered a race condition with a write-restart-read operation where the ISR could be called with both the RESTART and READ_REQ flags set.  

The original code handled the READ_REQ and asked for the read data before handling the RESTART which sent the write data to the user.
 
Reordering handling of the flags resolves this.


Grazfather (approved):

curuvar:
This code does not work on RP2350 running on risc-v.  I'm looking into how the hazard3.zig handles interrupts.  I've got a few changes to that which I'm working on.

mattnite (approved):

PR #429, opened 2025-04-04
regz: Give default values to padding and reserved bits fields

Uthedris:
Would it be worthwhile defaulting padding and reserved bits to zero so we don't have to add a  `.padding = 0,` and the like every time we write to a register?

### DIFF
--- a/examples/raspberrypi/rp2xxx/src/rp2040_only/tiles.zig
+++ b/examples/raspberrypi/rp2xxx/src/rp2040_only/tiles.zig
@@ -6,7 +6,7 @@ const Pio = rp2xxx.pio.Pio;
 const StateMachine = rp2xxx.pio.StateMachine;
 
 const ws2812_program = blk: {
-    @setEvalBranchQuota(5000);
+    @setEvalBranchQuota(10_000);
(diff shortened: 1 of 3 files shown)


curuvar:
Sure, I could use modify() but that does an unnecessary read of the value if I'm just initializing the values in a regeister where write() would be more appropriate.  

There are tons of examples in the codebase where you have things like `uart_regs.UARTIBRD.write(.{ .BAUD_DIVINT = baud_ibrd, .padding = 0 });`.  It would seem better, to me at least, to be able to omit the "`.padding = 0`" as that would be the expected default and including it just clutters the code.

Grazfather:
Yeah, I didn't ever see that, but I must have been playing with registers that don't have reserved or padding fields. It's indeed annoying to have to give those dummy values. This makes sense to me, then.

Grazfather (approved):

Grazfather:
Any idea why you had to bump the eval quota?

Uthedris:
Not really.  I had to do that before for an Mmio issue.  Someone had already bumped the quota from the default 2000 to 5000.  I put it up more and it worked.     I don't grok the Zig comptime in fullness yet.

PR #432, opened 2025-04-06
RP2XXX USB - replace rom.memcpy with @memcpy

arkadiuszwojcik:
In commit 68b5ad1 `rom.memcpy` was replaced by `@memcpy` by @patryk4815  but one more `rom.memcpy` left.

### DIFF
--- a/port/raspberrypi/rp2xxx/src/hal/usb.zig
+++ b/port/raspberrypi/rp2xxx/src/hal/usb.zig
@@ -17,7 +17,6 @@ pub const vendor = usb.vendor;
 pub const templates = usb.templates.DescriptorsConfigTemplates;
 pub const utils = usb.UsbUtils;
 
-const rom = @import("rom.zig");
 const resets = @import("resets.zig");
 
 pub const RP2XXX_MAX_ENDPOINTS_COUNT = 16;
@@ -256,7 +255,7 @@ pub fn F(comptime config: UsbConfig) type {
 
             const ep = hardware_endpoint_get_by_address(ep_addr);
 
-            _ = rom.memcpy(ep.data_buffer[0..buffer.len], buffer);
+            @memcpy(ep.data_buffer[0..buffer.len], buffer);
 
             // Configure the IN:
             const np: u1 = if (ep.next_pid_1) 1 else 0;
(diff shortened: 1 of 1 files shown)


arkadiuszwojcik:
Hmm intresting. @patryk4815 can you tell me if USB CDC example from fresh git clone of microzig and zig build works for you on RP2040? Because for me it doesn't but witch that change from this commit it does. Very bizarre.

patryk4815:
I have only rp2350, soo I will test this today or tomorrow

arkadiuszwojcik:
> Why do we want to do this? Don't we think that the rom version will be optimized for the device?

In other place it was already replaced from `rom.memcpy` to `@memcpy`, so this change is to keep consistency. But, probably better solution would be to utilze comptime and use `rom.memcpy` for rp2040 and `@memcpy` for rp2350

Grazfather:
Oh I did not know that they are not in the rp2350 rom, though I just checked (they are there but not exported)

KaroLaunonen:
At least I needed this change for RP2040 usb-hid example on zig 0.14.0. It would crash without the change.

Grazfather:
Can someone test this on rp2350? If it works, we can make `rom.memcpy` for rp2350 make a compile error or just `@memset`.

I don't think we should have `rom.<anything>` secretly call builtins, I'd rather have these functions in the HAL, and have those functions determine whether they call a builtin or a rom function.

arkadiuszwojcik:
@Grazfather I am very confused right now. Without this change RP2350 is working and RP2040 is not, but with this change it is other way around: RP2350 stops working and RP2040 starts working :D Just tested on both devices. This is very odd issue.

Grazfather:
If we move the memcpy to some `chip.mempy`, we can have a dispatch there that does `@memcpy` on rp2040 and 
`std.mem.copyForwards` on rp2350. This could be done in rom but that would be misleading.

arkadiuszwojcik:
I can change that PR and create such utility dispatch function form memcpy and limit it just to USB module with proper TODO comment, that this issue have to be resulved in future. Is it ok?

arkadiuszwojcik:
@Grazfather I changed solution to use `std.mem.copyForwards` and created issue for this problem, leaving that in code comment. Tested on both platforms RP2040 and RP2350 - it works.

Grazfather (approved):

PR #435, opened 2025-04-09
esp32-c3: interrupts and clock configuration

tact1m4n3:


port/espressif/esp/src/hal/rom.zig
```diff
@@ -0,0 +1,18 @@
...
+/// Microsecond delay using rom function.
+pub fn delay_us(us: u32) void {
+    switch (compatibility.chip) {
+        .esp32_c3 => functions.ets_delay_us(us),
+    }
+}
+
+pub const functions = switch (compatibility.chip) {
+    .esp32_c3 => struct {
+        pub const ets_delay_us: *const fn (us: u32) callconv(.c) void = @ptrFromInt(0x40000050);
```
Grazfather, line 13: Could we maybe define these addresses with a const somewhere, ideally with a name matching the datasheet?

mattnite (approved):

PR #436, opened 2025-04-10
esp32c3: Add usb serial jtag logger

tact1m4n3:


port/espressif/esp/esp32_c3_direct_boot.ld
```diff
@@ -39,7 +39,9 @@ SECTIONS
       microzig_bss_end = .;
   } > ram0
 
-  .flash_end :
+  /* NOTE: fixes large binaries issue for direct boot. if this should be NOLOAD then we should also
```
Grazfather, line 42: nit: capitalize `if`

Grazfather (approved):
Tested all 6 example variants, they work great.

Let's wait for someone to opine on the .flash_end q

mattnite (approved):

port/espressif/esp/esp32_c3_direct_boot.ld
tact1m4n3: I've went with another version that also seems to fix the problem. The advantage of this one is that if you add another section to flash you won't have to change it.

Grazfather: Yeah that is nicer, but do you need the `.flash_end` section at all?

port/espressif/esp/esp32_c3_direct_boot.ld
```diff
@@ -41,7 +41,7 @@ SECTIONS
 
   .flash_end :
   {
-      microzig_flash_end = .;
+      PROVIDE(microzig_flash_end = .);
   } > flash0
```
Grazfather, line 45:
```suggestion
 PROVIDE(microzig_flash_end = .);
```

port/espressif/esp/esp32_c3_direct_boot.ld
```diff
@@ -41,7 +41,7 @@ SECTIONS
 
   .flash_end :
   {
```
Grazfather, line 43:
```suggestion
```

port/espressif/esp/esp32_c3_direct_boot.ld
tact1m4n3: I don't think this works. The logic behind the .flash_end section is that you append the symbol to `flash0`. How would the linker know otherwise what `.` means?

Grazfather: Ah, I think you're right. If you want to use `.` you need it in a section in a memory region. We could avoid the section by using the math suggested above but this is fine.

PR #448, opened 2025-04-16
Fix PR website building gh action

Grazfather:
In the PR workflow we build the website, but we use zig master. Seems that zig master changed underneath us to make it no longer compatible with the version of zine we pin.

Simply use the 0.14.0 release of zig for this workflow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -183,7 +183,7 @@ jobs:
       - name: Setup Zig
         uses: mlugg/setup-zig@v1
         with:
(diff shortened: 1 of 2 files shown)


tact1m4n3 (approved):
Nice catch. I guess `${{ env.ZIG_VERSION }}` was using the master branch?

Grazfather:
Yes exactly, which I guess we want for the other workflows... that could be fragile of course, but we can/should adjust the pinned versions for tags, otherwise e.g. re-running the 0.14.0 release in the future will fail since `master` will change.

PR #426, opened 2025-04-02
HID: report descriptor for boot keyboard

hryx:
- Add new usage constants to support keyboard report descriptor
- Add usage min/max helpers to support keyboard report descriptor
- Add report descriptor for a keyboard report conforming to the boot protocol

Reference: https://usb.org/sites/default/files/hid1_11.pdf appendix B.1

PR #449, opened 2025-04-16
HID: basic keyboard support

hryx:
This PR includes enough changes to facilitate a working USB HID keyboard implementation using MicroZig.

- Fix critical typo from #426
- Make fields pub that are required for app code to define its own report descriptors
- ACK requests that are required for keyboard; some of these are stubs which need further improvements

I used this code, along with my own application code, to create a USB HID keyboard that connects and types successfully. Tested on macOS Sequoia and Arch Linux. I tried both the included `ReportDescriptorKeyboard` and a custom report format, both work well.

I have done my best to document the areas that were added as stubs and which need improvement.

### DIFF
--- a/core/src/core/usb.zig
+++ b/core/src/core/usb.zig
@@ -238,7 +238,7 @@ pub fn Usb(comptime f: anytype) type {
                                     if (feature) |feat| {
                                         switch (feat) {
                                             .DeviceRemoteWakeup, .EndpointHalt => CmdEndpoint.send_cmd_ack(),
-                                            // TODO
+                                            // TODO: https://github.com/ZigEmbeddedGroup/microzig/issues/453
                                             .TestMode => {},
                                         }
                                     }
--- a/core/src/core/usb.zig
+++ b/core/src/core/usb.zig
@@ -23,6 +23,7 @@ pub const utils = @import("usb/utils.zig");
 pub const templates = @import("usb/templates.zig");
 
 const DescType = types.DescType;
+const FeatureSelector = types.FeatureSelector;
 const Dir = types.Dir;
 const Endpoint = types.Endpoint;
 const SetupRequest = types.SetupRequest;
@@ -202,12 +203,12 @@ pub fn Usb(comptime f: anytype) type {
                             const req = SetupRequest.from_u8(setup.request);
                             if (req == null) return;
                             switch (req.?) {
-                                SetupRequest.SetAddress => {
+                                .SetAddress => {
                                     S.new_address = @as(u8, @intCast(setup.value & 0xff));
                                     CmdEndpoint.send_cmd_ack();
                                     if (S.debug_mode) std.log.info("    SetAddress: {}", .{S.new_address.?});
                                 },
-                                SetupRequest.SetConfiguration => {
+                                .SetConfiguration => {
                                     if (S.debug_mode) std.log.info("    SetConfiguration", .{});
                                     const cfg_num = setup.value;
                                     if (S.cfg_num != cfg_num) {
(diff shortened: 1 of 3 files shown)


core/src/core/usb.zig
```diff
@@ -226,12 +227,22 @@ pub fn Usb(comptime f: anytype) type {
...
                                     if (descriptor_type) |dt| {
                                         try process_get_descriptor(setup, dt);
                                     }
                                 },
+                                .SetFeature => {
+                                    const feature = FeatureSelector.from_u8(@intCast(setup.value >> 8));
+                                    if (feature) |feat| {
+                                        switch (feat) {
+                                            .DeviceRemoteWakeup, .EndpointHalt => CmdEndpoint.send_cmd_ack(),
+                                            // TODO
```
Grazfather, line 241: Let's file an issue here for any follow up and put the issue number in the comment

core/src/core/usb/hid.zig
```diff
@@ -605,6 +595,32 @@ pub const HidClassDriver = struct {
...
+                            // For now, this request is ACKed without doing anything; in practice,
+                            // the OS will reuqest the report protocol anyway, so usually only one format is needed.
+                            // Unless the report format matches the boot protocol exactly (see ReportDescriptorKeyboard),
+                            // our device might not work in a limited BIOS environment.
+                            self.device.?.control_ack(setup);
+                        }
+                    },
+                    .SetReport => {
+                        if (stage == .Setup) {
+                            // TODO: This request sends a feature or ouptut report to the device,
```
Grazfather, line 621: typo output

core/src/core/usb/types.zig
```diff
@@ -52,15 +52,27 @@ pub const ControlStage = enum {
...
     }
 };
 
+pub const FeatureSelector = enum(u8) {
+    EndpointHalt = 0x00,
+    DeviceRemoteWakeup = 0x01,
+    TestMode = 0x02,
+    // The remaining features only apply to OTG devices.
+
+    pub fn from_u8(v: u8) ?@This() {
```
Grazfather, line 71: Can't we just use `@enumFromInt`?

hryx: Sort of. At first, I only did it like this to match the style of other decls in this file (`DescType`, `TransferType`, `SetupRequest`) as well as a few other flies. Those are all thin wrappers around `std.meta.intToEnum(...) catch null`.

But the main reason _not_ to use `@emunFromInt` directly is that the use case is decoding bytes from the host. With raw `@enumFromInt`, invalid bytes from the host would trigger illegal behavior rather than an error or null.

How would you like me to proceed? Happy to take it in any direction.

hryx:
> What was the critical typo?

That would be changing `hid_output` to `hid_input` in the boot keyboard report descriptor. Previously, keycodes could not be reported from the device.

https://github.com/ZigEmbeddedGroup/microzig/pull/449/files#diff-5cbae23b28bf92266f69cd560f2f6214a056626b41132a84185c51c94c82c798L531-R521

> Also, can you please rebase

✅ Done

core/src/core/usb.zig
hryx: Upon further investigation, "test mode" seems to be a way to test the integrity of the USB hardware itself and is only sent by testing devices, not typical hosts. I can't tell if this is something that could even be implemented in firmware on a RP2040 (for example).

I can still make an issue to investigate more, regardless.

Grazfather (approved):

mrusme:
@hryx thank you for this! 

> USB HID keyboard that connects and types successfully.

Do you happen to have a brief example of how you did it? I cannot seem to get anything being typed.

PR #457, opened 2025-04-18
drivers: Add 28byj/uln2003 stepper motor driver

Grazfather:


examples/raspberrypi/rp2xxx/build.zig
```diff
@@ -52,6 +52,7 @@ pub fn build(b: *std.Build) void {
         .{ .name = "watchdog-timer", .file = "src/watchdog_timer.zig" },
         .{ .name = "interrupts", .file = "src/interrupts.zig" },
         .{ .name = "stepper", .file = "src/stepper.zig" },
+        .{ .name = "new_stepper", .file = "src/new_stepper.zig" },
```
Grazfather, line 55: Not sure if we want to add this example. wdyt? If so, I'll rename this and cleanup the example, otherwise it's just for testing and I'll remove before merge.

mattnite (approved):
Yes I do like having this example, please rename it

mattnite (approved):

PR #463, opened 2025-04-19
Update boxzer to use new hashing format

mattnite:


mattnite:
Leaving for now, it looks like the hash contains the right id and size, but the raw digest is incorrect, will need to compare with the compiler implementation of that hash to see if anything has changed.

PR #469, opened 2025-04-21
Interrupt and multicore safe allocator

Uthedris:
This PR supplies an implementation of a std.mem.Allocator.

It can be use to allocate memory from a specified buffer, or from the otherwise unused ram that follows the .bss section. 

Unlike std.heap.FixedBufferAllocator it supports re-use of freed memory, coalescing adjacent freed blocks to lessen fragmentation. 

It can be safely used from multiple cores or from an interrupt service routine.

It has been tested on RP2040, RP2350 (both amd and riscv) and on an esp-c3.

### DIFF
--- a/core/src/allocator.zig
+++ b/core/src/allocator.zig
@@ -3,47 +3,123 @@ const microzig = @import("microzig.zig");
 
 const Alignment = std.mem.Alignment;
 
+extern var microzig_heap_start: u8;
+extern var microzig_heap_end: u8;
+
 /// The number of free lists we maintain.
 const free_list_count: usize = 6;
 
-/// The free lists
-var free_lists: [free_list_count]?*Chunk = undefined;
+const Alloc = @This();
 
-var low_boundary: usize = undefined;
-var high_boundary: usize = undefined;
+/// The lists of free chunks.
+free_lists: [free_list_count]?*Chunk = @splat(null),
 
-var mutex: microzig.hal.mutex.Mutex = .{};
+/// The beginning address of the memory region that can be allocated from.
+low_boundary: usize,
 
-/// Set up the initial heap computing a low and high boundary and adding all
-/// available memory to the free list as a single chunk.
+/// The next address beyond the highest address of the memory region can allocated from.
+high_boundary: usize,
+
+/// An optional fallback allocator that is used when the allocator runs out of memory.
+/// This allows allocation from multiple disjoint memory regions.
+fallback: ?*Alloc = null,
+
+/// A mutex used to protect access to the allocator.
+mutex: microzig.interrupt.Mutex = .{},
+
+/// Return a []u8 slice that contains the memory located between the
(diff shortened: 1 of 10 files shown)


core/src/allocator.zig
```diff
@@ -0,0 +1,614 @@
...
+/// var heap_allocator = microzig.core.Allocator.setupWithHeap(1024);
+///
+/// // Get the std.mem.Allocator from the heap allocator.
+/// const allocator : std.mem.Allocator = heap_allocator.allocator();
+/// ```
+///
+/// aParameters:
+/// - `reserve`: The number of bytes to omit at the end of the heap.
+///
+pub fn setupWithHeap(reserve: usize) Alloc {
```
mattnite, line 41: Use snake case for functions, also zig conventions to use the word init, so something like `init_with_heap`

mattnite (changes requested):
This is really great, but an allocator is going to need unit tests, just a few for this PR and we can follow up later as well.

Uthedris:
Looking closer, I'm not sure how to proceed with adding tests.   I'm adding a zig test case to the allocator itself.  That would run on the machine doing the compilation, but should validate the allocator itself.  Do you also want some sort of test program to run under microzig on the microcontrollers. I'd guess that would go in the test direcory, but I'm unsure of the details of how I should do that.

Grazfather:
I don't think we have a place for that now, but we do want to get HIL testing in place, so at the very least I'd say hold on to it. We could commit here, but I'd be afraid it would bit rot with nothing running it.

core/src/allocator.zig
Uthedris: Done

Uthedris:
There is an example program in `examples/raspberrypi/rp2xxx/src/allocator.zig` that tests the allocator.

mattnite (commented):
Could we have some unit tests that run on a host? 

And do we really want to make a global heap in the linkerscript? Why not just make a static buffer?

Uthedris:
If by "run on a host," you mean to run on the machine where the code is compiled, the problem is that the mutex code suspends interrupts and, since that accesses harware registers, those routines won't compile on the host.   I suppose I could do some comptime magick to exclude the mutex from the host build, but that would prevent testing allocation from different exectuion threads.  

The code does allow accessing from a static buffer, it also has a function to make such a buffer from the remaining heap. On a lot of systems I've used that was standard practice, the stack grows down from the top of the heap and dynamic allocations grow up from the bottom.  It just seemed reasonable provide a method to be able to use whatever memory a particular chip had avaiable.

mattnite:
Good point, in that  case we are good to go, just fix the conflicts and then we're good to go!

mattnite:
@Uthedris ping, there are conflicts still

mattnite:
@Uthedris ping

Grazfather:
@Uthedris this is looking good. If you can just add a few unit tests for the allocator then this should be good to go.

mattnite (approved):

PR #482, opened 2025-04-21
update package naming scheme

mattnite:


tools/package-test/build.zig.zon
```diff
@@ -1,6 +1,6 @@
 .{
-    .name = .package_test,
-    .fingerprint = 0x63321bc42367f374,
+    .name = .microzig_package_test,
```
tact1m4n3, line 2: shouldn't this be microzig_tools_packagetest?

Grazfather (commented):
we can probably get rid of the 'extra' gitignore in the aviron dir

Grazfather (approved):
Yeah we can clean that up whenever. It's inconsequential.

PR #439, opened 2025-04-11
Add time hal for esp32c3

Grazfather:

PR #483, opened 2025-04-21
esp: Add GPIO device to hal

Grazfather:
* Fix a few bugs with GPIO-related registers.
* Add a GPIO device interface type
* Port the stepper driver example to run on it (tested on hardware)

Merge after #439


port/espressif/esp/src/hal/gpio.zig
```diff
@@ -135,7 +136,7 @@ pub const Pin = struct {
     }
 
     pub fn read(self: Pin) Level {
-        std.debug.assert(IO_MUX.GPIO[self.number].FUN_IE == 1);
+        std.debug.assert(IO_MUX.GPIO[self.number].read().FUN_IE == 1);
```
Grazfather, line 139: This was a bug

examples/espressif/esp/src/stepper_driver.zig
```diff
@@ -0,0 +1,66 @@
...
+        ms1: GPIO_Device,
+        ms2: GPIO_Device,
+        ms3: GPIO_Device,
+        dir: GPIO_Device,
+        step: GPIO_Device,
+    } = undefined;
+    inline for (std.meta.fields(@TypeOf(pins)), .{ 0, 1, 2, 20, 10 }) |field, num| {
+        const pin = gpio.instance.num(num);
+        // Give the pin a sane default config
+        pin.apply(.{});
```
Grazfather, line 31: Interesting that I had to do this. I guess the default values for a bunch of registers don't work. If that's the case, then maybe the set direction method should do more than just set outp.t

mattnite (approved):

PR #486, opened 2025-04-22
padding and reserve field cleanup

Uthedris:
Added code to default padding and reserve in some packed sturct definitions where they were missing.

Removed un-needed padding = 0 and reserver = 0 in register writes.


tact1m4n3:
Should those reserved fields in cortex_m be set to zero by default? I think this isn't recommended to be done.

Grazfather:
What should they be set to? The problem is that if you use `.write` you have to provide a value for the padding fields anyway, and if you use `.modify`, it's just going to read and write back whatever it read. I don't know of a system (that isn't buggy) where writing a 1 or a 0 to a padding or reserved field actually does anything.

Uthedris:
I suppose the code that parsed the .svd files could set the default for padding and filler by parsing the `<reseValue>`, but since I don't know of a case where writing zero is harmful there seems little point.

mattnite:
I'm open to this change for padding and reserved bits, we do need to change the default based off the reset value in the SVD, but also consider that across the entire repo this change is only saving 100 lines of code, so I don't think it's as big of a convenience as it may seem.

@ikskuh and I did discuss this in the past and we decided that the verbosity was worth it over potential hardware bugs, but @Grazfather does make a good point: I've also never seen hardware that misbehaves when writing 0 to padding/reserved bits. Felix what do you think of this change?

Uthedris:
The reason for my orignal PR (#429) suggesting setting these defaults wasn't to save space in the repo, but rather to not make the users of Microzig have to worry about setting unused bits every time they do an Mmio write.

I had wondered whether the write should work like modify, but instead of reading the register to get the base values the write would start with the reset value. That way the user would only have to specify the values that were not the default.  

I didn't pursue that because there is value in code being clear and explicitly defining the values of every actual field when initializing a register, though I do think that having to define values for unused parts of the register is a bit much.

mattnite:
Okay I'm happy with that. Writes must be explicit for all real bitfields -- way too easy to shoot yourself in the foot if we use reset values as defaults, and padding and reserved will be defaulted to zero. I'll make this change in regz as well.

mattnite (approved):

mattnite:
Okay just need to fix merge conflicts and we're good to go

Uthedris:
I fixed the conflicts, but there was some wierdness with `examples/espressif/esp/src/systimer.zig`.  I commented it out but someone more familiar with the expressif stuff might want to look at it.

examples/espressif/esp/src/systimer.zig
```diff
@@ -32,6 +32,10 @@ pub fn main() !void {
     // sets the period to one second.
     alarm.set_period(@intCast(1_000_000 * systimer.ticks_per_us()));
 
+    // SYSTIMER.COMP0_LOAD.write(.{
+    //     .TIMER_COMP0_LOAD = 1,
+    // });
+
```
tact1m4n3, line 38: You can remove this. It is done inside the hal

port/espressif/esp/src/hal/clocks/esp32_c3.zig
```diff
@@ -258,10 +258,10 @@ fn bbpll_configure(pll_freq: CpuClockSource.PllClock.PllFreq) void {
 
 const I2C_ANA_MST_TYPE = extern struct {
     ANA_CONF0: microzig.mmio.Mmio(packed struct {
-        reserved0: u2,
+        reserved0: u2 = 0,
         BBPLL_STOP_FORCE_HIGH: u1,
         BBPLL_STOP_FORCE_LOW: u1,
-        padding: u28,
+        padding: u28 = 0,
```
tact1m4n3, line 264: Could you please remove the default values here? That struct's definition is not complete, so padding isn't really padding, but other fields that I ignored for now.

port/raspberrypi/rp2xxx/src/cpus/hazard3.zig
```diff
@@ -481,29 +481,29 @@ pub const csr = struct {
...
     });
     pub const meinext = Csr(0xbe4, packed struct {
         update: u1,
-        reserved0: u1,
+        reserved0: u1 = 0,
         irq: u9,
-        reserved1: u20,
+        reserved1: u20 = 0,
         noirq: u1,
     });
```
tact1m4n3, line 508: Could you also please undo the changes to csrs (in hazard3) for now? Riscv has some strict rules regarding reserved bits. Don't know how much of that affects hazard3 custom csrs, but for know let's leave them as they are.

port/espressif/esp/src/hal/clocks/esp32_c3.zig
Uthedris: Done

port/raspberrypi/rp2xxx/src/cpus/hazard3.zig
Uthedris: done

tact1m4n3 (approved):

port/espressif/esp/src/hal/clocks/esp32_c3.zig
tact1m4n3: Thank you!

PR #494, opened 2025-04-24
zig fmt everything/test in ci

Grazfather:
I noticed a few files getting changed automatically when I opened them, because my editor is configured to automatically `zig fmt` on save. I imagined most of us have this setting, and it can be annoying when our diffs are larger than they should be. `zig fmt .` to fix.

Added a workflow to run `zig fmt --check` and fail if anything shows up, to ensure that no one commits improperly-formatted files in the future.


Grazfather:
Confirmed it'll fail:

```
Run zig fmt --check .
./examples/espressif/esp/hello.zig
Error: Process completed with exit code 1.
```

arkadiuszwojcik:
What about code that often looks better as multiline, but fmt reduces it to a single line? Is this something that can be configured?

Grazfather:
If you add a trailing comma to the list the formatter will keep it on separate lines

Grazfather:
For example:

```zig
const x = .{1,2,3};
const y = .{4,5,6,};
```
becomes
```zig
const x = .{ 1, 2, 3 };
const y = .{
    4,
    5,
    6,
};
```

arkadiuszwojcik:
It’s really bizarre to me that it works that way—new lines because of a trailing comma. And how about constructions like this:

```
const val: u16 = consts.IRQ_F2_F3_FIFO_RD_UNDERFLOW | 
            consts.IRQ_F2_F3_FIFO_WR_OVERFLOW | 
            consts.IRQ_COMMAND_ERROR | 
            consts.IRQ_DATA_ERROR | 
            consts.IRQ_F2_PACKET_AVAILABLE | 
            consts.IRQ_F1_OVERFLOW;
```

`fmt` flattened that for me, but it looks much better vertically.

Grazfather:
that does not flatten for me

mattnite:
@arkadiuszwojcik There are certainly times where formatting the code a certain way is better than whatever zig fmt gives you, I do this a lot for tables for example. In those cases you can disable formatting with: `// zig fmt: off` and then set it to on when you're done.

PR #498, opened 2025-04-26
[hdsc] HC32L110 bring up

vesim987:
As a disclaimer:
- this thing is some weird mix of CortexM0+ and peripherals from random 8051 MCU
- the documentation is lacking
- the official SDK is barely helping
- function configuration for pins is a mess, so most likely i will need to do it outside the standard `gpio` part of the HAL.

Don't yell at me for random numbers in the code, a lot of it is from poking the MCU with a stick and seeing what happens :D
And also, examples are taken from rp2xxx i plan to keep similar interface where applicable, most of configuration is most likely going to be comptime-only due to the limited space.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -148,7 +148,8 @@ jobs:
           [
             espressif/esp,
             gigadevice/gd32,
-          #microchip/avr,
+            hdsc/hc32l110,
+            #microchip/avr,
             microchip/atsam,
             nordic/nrf5x,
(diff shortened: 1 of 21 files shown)


arkadiuszwojcik:
I think you can remove all examples for this HAL except `blinky` and some other simple ones that might work. BTW, a very unorthodox selection of HAL to work with ;)

mattnite:
@vesim987 are you going to finish this?

PR #501, opened 2025-04-26
Add SH1106 driver

hryx:
This adds support for the SH1106 OLED display driver. Tested and working on hardware (SPI only).

This device is very similar to the already supported SSD1306. They are so close that another approach is to create some abstraction over both, but I felt that duplicating code was a cheaper abstraction to start off. Be aware that some of the code and comments are in fact copied directly from the SSD1306 implementation.

Main differences between SH1106 and SSD1306:

- SH1106 only has page addressing mode, no "horizontal" or "vertical" mode.
- Some command values are different, though most of the commands are the same.

Excluded:

- Tests. Should these be included? Are unit tests valuable for these drivers (as opposed to e.g. end-to-end tests on hardware)?
- Framebuffer; would be 100% same as ssd1306.zig's Framebuffer. Could be moved to shared file.

Non-critical changes from SSD1306 code:

- Fix any obvious style inconsistencies.
- Remove stuttering from decl naming.
- Minor command name changes for clarity.

Etcetera:

- Fix typo in SSD1306 code.

### DIFF
--- a/drivers/display/sh1106.zig
+++ b/drivers/display/sh1106.zig
@@ -10,6 +10,9 @@
 
 const std = @import("std");
 const mdf = @import("../framework.zig");
+const common = @import("common.zig");
+
+const I2C_ControlByte = common.I2C_ControlByte;
 
 pub const Driver_Mode = enum {
     /// The driver operates in I²C mode, which requires a 8 bit datagram device.
@@ -38,6 +41,12 @@ pub const Options = struct {
     /// Which digital i/o interface should be used.
     Digital_IO: type = mdf.base.Digital_IO,
 };
+
+pub const DisplayOnMode = enum(u8) { resume_to_ram = 0xA4, ignore_ram = 0xA5 };
+pub const NormalOrInverseDisplay = enum(u8) { normal = 0xA6, inverse = 0xA7 };
+pub const DisplayMode = enum(u8) { off = 0xAE, on = 0xAF };
+
+pub const InputError = error{InvalidEntry};
 
 const cols = 132;
 const rows = 64;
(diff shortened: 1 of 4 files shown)


drivers/display/sh1106.zig
```diff
@@ -0,0 +1,371 @@
...
+
+            try self.dc_pin.write(switch (mode) {
+                .command => .low,
+                .data => .high,
+            });
+        }
+    };
+}
+
+pub const I2C_ControlByte = @import("ssd1106.zig");
```
Grazfather, line 359: could put this into a common file?

drivers/display/sh1106.zig
```diff
@@ -0,0 +1,371 @@
...
+
+pub const DisplayOnMode = enum(u8) { resume_to_ram = 0xA4, ignore_ram = 0xA5 };
+pub const NormalOrInverseDisplay = enum(u8) { normal = 0xA6, inverse = 0xA7 };
+pub const DisplayMode = enum(u8) { off = 0xAE, on = 0xAF };
+
+pub const InputError = error{InvalidEntry};
```
Grazfather, line 365: why not put these at the top with the other types?

Grazfather (approved):

drivers/display/sh1106.zig
hryx: Sounds good - not sure where it should go though, because this type doesn't seem specific to display drivers. I won't overthink it for now and will put it in drivers/display/common.zig unless you have another preference.

drivers/display/sh1106.zig
hryx: Because of copy/paste :P

I'll put them up there!

mattnite (approved):

PR #502, opened 2025-04-26
[drivers] More user friendly SSD1306 init function

vesim987:
With this it can be used as:
```
    const  display = try ssd1306.init(.i2c, i2c_device, null);
```

### DIFF
--- a/drivers/display/ssd1306.zig
+++ b/drivers/display/ssd1306.zig
@@ -44,6 +44,24 @@ pub const SSD1306_Options = struct {
     Digital_IO: type = mdf.base.Digital_IO,
 };
 
+pub fn init(comptime mode: Driver_Mode, device: anytype, digital_io: anytype) !SSD1306_Generic(.{
+    .mode = mode,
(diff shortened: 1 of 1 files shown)


drivers/display/ssd1306.zig
```diff
@@ -854,7 +872,7 @@ pub const Framebuffer = struct {
 
     // Tests:
 
-    test init {
+    test "Frmebuffer.init" {
```
Grazfather, line 875: typo

drivers/display/ssd1306.zig
```diff
@@ -47,6 +47,24 @@ pub const SSD1306_Options = struct {
     Digital_IO: type = mdf.base.Digital_IO,
 };
 
+pub fn init(comptime mode: Driver_Mode, device: anytype, digital_io: anytype) !SSD1306_Generic(.{
```
Grazfather, line 50: Does this need an error type in the return?

Grazfather: I see it does

PR #505, opened 2025-04-28
Cortex m33/m55 system control block updates.

Uthedris:
Updated the `SystemControlBlock` definition in m33.zig and m55.zig to provide register field definitions where appropriate (like M0-plus and the other do).

Due to the fact that the ARMv8-M SCB starts at 0xE000ECFC instead of 0xE000ED00.  This change also defines an scb_base_offset in each of the chip model files instead of using a constant in the overall cortex_m.zig file.

Should these files default reserved bits to zero too?


mattnite (approved):
Yeah we could make the reserved and padding default to 0 here.

Issue #220, opened 2024-07-14
dimIndex does not get parsed correctly in .svd files


DNedic:
While updating the ESP32-C3 `.svd` file, I noticed that register groups that had  `dimIndex` fields added were suddenly missing in `chip.zig` and were replaced with `reserved` fields.

To reproduce, replace the ESP32-C3 `.svd` file with one found in the [esp-pacs repo](https://github.com/esp-rs/esp-pacs/blob/main/esp32c3/svd/esp32c3.base.svd), or [patch it](https://github.com/esp-rs/esp-pacs?tab=readme-ov-file#patching-the-svds) and use the patched version and re-build and check the `GPIO.PINn` registers.

mattnite:
Checked and we are now generating this correctly:

```
    /// offset: 0x68
    reserved104: [12]u8,
    /// GPIO pin configuration register
    /// offset: 0x74
    PIN: [26]mmio.Mmio(packed struct(u32) {
        /// set GPIO input_sync2 signal mode. :disable. 1:trigger at negedge. 2or3:trigger at posedge.
        PIN_SYNC2_BYPASS: u2 = 0x0,
        /// set this bit to select pad driver. 1:open-drain. :normal.
        PIN_PAD_DRIVER: u1 = 0x0,
        /// set GPIO input_sync1 signal mode. :disable. 1:trigger at negedge. 2or3:trigger at posedge.
        PIN_SYNC1_BYPASS: u2 = 0x0,
        reserved7: u2 = 0,
        /// set this value to choose interrupt mode. :disable GPIO interrupt. 1:trigger at posedge. 2:trigger at negedge. 3:trigger at any edge. 4:valid at low level. 5:valid at high level
        PIN_INT_TYPE: u3 = 0x0,
        /// set this bit to enable GPIO wakeup.(can only wakeup CPU from Light-sleep Mode)
        PIN_WAKEUP_ENABLE: u1 = 0x0,
        /// reserved
        PIN_CONFIG: u2 = 0x0,
        /// set bit 13 to enable CPU interrupt. set bit 14 to enable CPU(not shielded) interrupt.
        PIN_INT_ENA: u5 = 0x0,
        padding: u14 = 0,
    }),
```

PR #508, opened 2025-04-28
Add support for SVD dimIndex parsing

zovko:
DimElementGroup support for peripheral, register and field types (without `dimName`)

New features:
- List type (`%s` parsing) for DimElementGroup
  - Supported by `register` and `field` types. [register dim description](https://arm-software.github.io/CMSIS_5/SVD/html/elem_registers.html#elem_register) [field dim description](https://arm-software.github.io/CMSIS_5/SVD/html/elem_registers.html#elem_fields)
  - `%s` can be anywhere in string, `dimIndex` must be defined.
- Array type (`[%s]` parsing) for `DimElement` group
  - Supported by `peripheral` and `register` [peripheral dim description](https://arm-software.github.io/CMSIS_5/SVD/html/elem_peripherals.html#elem_peripheral)
  - `[%s]` must be at the end of the string
- List and array type described [here](https://arm-software.github.io/CMSIS_5/SVD/html/elem_special.html#dimElementGroup_gr)

Changed implementation:
- Fields cannot have `DimElementGroup` array type.
- `elements.dim_increment * 8) != size` removed. This is not always true, specifically for list types.

**Note:** This heavily breaks esp examples, rasperrypi seems to still compile.

Partially solves #220

Still to implement:
- `dimName`, `displayName`
- CSV parsing for `dimIndex`


tools/regz/src/svd.zig
```diff
@@ -386,19 +389,78 @@ fn load_register(
...
+fn load_register_with_dim_element_group(ctx: *Context, node: xml.Node, parent: StructID, dim_elements: DimElements) !void {
     const db = ctx.db;
-    const register_props = try ctx.derive_register_properties_from(node, .{ .@"struct" = parent });
     const name = node.get_value("name") orelse return error.MissingRegisterName;
+    const register_props = try ctx.derive_register_properties_from(node, .{ .@"struct" = parent });
     const size = register_props.size orelse return error.MissingRegisterSize;
-    const count: ?u64 = if (try DimElements.parse(node)) |elements| count: {
-        if (elements.dim_index != null or elements.dim_name != null)
-            return error.TodoDimElementsExtended;
+    const address_offset_string = node.get_value("addressOffset") orelse return error.MissinRegisterOffset;
```
Grazfather, line 421: typo in the error name

tools/regz/src/svd.zig
```diff
@@ -424,32 +486,17 @@ fn load_register(
...
 fn load_field(ctx: *Context, node: xml.Node, register_id: RegisterID) !void {
     const db = ctx.db;
 
     const bit_range = try BitRange.parse(node);
-    const count: ?u16 = if (try DimElements.parse(node)) |elements| count: {
-        if (elements.dim_index != null or elements.dim_name != null)
+    const dim_elements = try DimElements.parse(ctx, node);
+    const count: ?u16 = if (dim_elements) |elements| count: {
+        if (elements.dim_name != null)
             return error.TodoDimElementsExtended;
```
Grazfather, line 498: Does this error name still make sense?

tools/regz/src/svd.zig
```diff
@@ -629,24 +693,95 @@ const DimElements = struct {
...
+            @memset(buf, 0);
+            // dim index range is with
+            dim_index = try std.fmt.bufPrintZ(buf, "0-{d}", .{dim.? - 1});
+        }
+
+        return DimElements{
+            .type = if (is_array_type) DimType.array else DimType.list,
+            .dim = if (dim) |d| blk: {
+                break :blk d;
+            } else return null,
```
Grazfather, line 736:
```suggestion
            .dim = dim.? orelse return null;
```
This should work, if not I think you can create a block for only t

Grazfather: e.g. `.dim = if (dim) |d| d else return null,`

tools/regz/src/svd.zig
zovko: It does,dim_name is not yet handled

mattnite (approved):
Looks good to me, just get the fixes that @Grazfather has suggested in. It looks like this changes the generated code for esp which will need updating, when I have time I could try to add that to the PR for you tomorrow.

For the stm32 CI failure all you need to do is run `zig build -Dgenerate` in the `port/stmicro/stm32` directory and commit the changes.

zovko:
@mattnite The changes do not look good for stm32 now, for example SQR1 is definitely different now. What to do about that specifically? I added the commit here but I believe the stm32 builds will be broken now.

tools/regz/src/Database.zig
```diff
@@ -1669,29 +1669,27 @@ pub fn add_struct_field(db: *Database, parent: StructID, opts: AddStructFieldOpt
 
     try db.exec(
         \\INSERT INTO struct_fields
-        \\  (struct_id, name, description, size_bits, offset_bits, enum_id, count, stride)
```
mattnite, line 1672: I think I understand the problem now, what you want to do is revert your changes in `Database.zig` and `gen.zig`. Ultimately your change is about improving how we interpret SVD, not the datamodel. The datamodel sits between SVD, ATDF, a specific one for STM32, and more in the future, so just because a certain field may not be populated by the SVD code, doesn't mean it's not used elsewhere. Thank you!

mattnite (changes requested):

tools/regz/src/Database.zig
zovko: `Database.zig` changes reverted. `gen.zig` did not have implementation changes, only tests. I ran the `zig build -Dgenerate` in the `port/stmicro/stm32` directory and there are no changes anymore, so implementation for stm32 is unaffected.

tools/regz/src/svd.zig
```diff
@@ -629,24 +693,93 @@ const DimElements = struct {
...
+        if (dim_index == null and !is_array_type and is_list_type) {
+            const buf = try ctx.arena.allocator().allocSentinel(u8, 16, 0);
+            @memset(buf, 0);
+            // dim index range is with
+            dim_index = try std.fmt.bufPrintZ(buf, "0-{d}", .{dim.? - 1});
+        }
+
+        return DimElements{
+            .type = if (is_array_type) DimType.array else DimType.list,
+            .dim = dim orelse return null,
```
Grazfather, line 734: You handle this above (line 708), so you should just be able to do `dim.?` here.

mattnite:
@zovko apply this patch and it should fix the last CI failure. It looks like the old code was trying its best to make repeated registers into arrays, and so I changed some registers to use `[%s]` instead of `<name>%s` to create arrays instead of creating distinct registers.

```
diff --git a/port/espressif/esp/src/chips/ESP32-C3.svd b/port/espressif/esp/src/chips/ESP32-C3.svd
index ab064fa..ff363b2 100644
--- a/port/espressif/esp/src/chips/ESP32-C3.svd
+++ b/port/espressif/esp/src/chips/ESP32-C3.svd
@@ -11190,7 +11190,7 @@
         <register>
           <dim>26</dim>
           <dimIncrement>0x4</dimIncrement>
-          <name>PIN%s</name>
+          <name>PIN[%s]</name>
           <description>GPIO pin configuration register</description>
           <addressOffset>0x74</addressOffset>
           <size>0x20</size>
@@ -11264,7 +11264,7 @@
         <register>
           <dim>128</dim>
           <dimIncrement>0x4</dimIncrement>
-          <name>FUNC%s_IN_SEL_CFG</name>
+          <name>FUNC_IN_SEL_CFG[%s]</name>
           <description>GPIO input function configuration register</description>
           <addressOffset>0x154</addressOffset>
           <size>0x20</size>
@@ -11295,7 +11295,7 @@
         <register>
           <dim>26</dim>
           <dimIncrement>0x4</dimIncrement>
-          <name>FUNC%s_OUT_SEL_CFG</name>
+          <name>FUNC_OUT_SEL_CFG[%s]</name>
           <description>GPIO output function select register</description>
           <addressOffset>0x554</addressOffset>
           <size>0x20</size>
@@ -12846,7 +12846,7 @@
         <register>
           <dim>8</dim>
           <dimIncrement>0x4</dimIncrement>
-          <name>COMD%s</name>
+          <name>COMD[%s]</name>
           <description>I2C_COMD%s_REG</description>
           <addressOffset>0x58</addressOffset>
           <size>0x20</size>
@@ -15904,7 +15904,7 @@
         <register>
           <dim>22</dim>
           <dimIncrement>0x4</dimIncrement>
-          <name>GPIO%s</name>
+          <name>GPIO[%s]</name>
           <description>IO MUX Configure Register for pad XTAL_32K_P</description>
           <addressOffset>0x4</addressOffset>
           <size>0x20</size>
@@ -36929,4 +36929,4 @@
       </registers>
     </peripheral>
   </peripherals>
-</device>
\ No newline at end of file
+</device>
```

zovko:
@mattnite Should that patch even be done? It doesn't sound right to update the vendor SVD with local changes. What would be the reason to change the registers from sequence to array? This can also easily introduce bugs if the `dimIncrement` is different than register size (not the case for suggested patch though). But regardless, this practice can introduce more bugs in future.

I would even advise to update the SVD file itself for C3. Latest version is version 18 [(link)](https://github.com/espressif/svd/blob/main/svd/esp32c3.svd) while the current SVD file in the repo is version 8. I would rather update the examples than the SVD itself to make the examples compile. 

Also, with SVD modification it will also break the direct link to the documentation, for example:
-FUNC%s_IN_SEL_CFG
+FUNC_IN_SEL_CFG[%s]

[Documentation](https://www.espressif.com/sites/default/files/documentation/esp32-c3_technical_reference_manual_en.pdf) can be easily searchable for `FUNC0_IN_SEL_CFG`, while `FUNC_IN_SEL_CFG` will not be found

mattnite:
Yes it's okay to apply this patch. I want to get this functionality into Regz, and we'll eventually move over to directly fetching those SVD files from that repo.

zovko:
@mattnite applies the patch to C3 SVD

mattnite (approved):

PR #522, opened 2025-05-02
WCH CH32V: fix cpu startup

ghostiam:
~Depends on: https://github.com/ZigEmbeddedGroup/microzig/pull/520
Please review only this commit: https://github.com/ZigEmbeddedGroup/microzig/pull/522/commits/2281a878cbf53fc56c45bfe414eccc0bf462e5d4~

Because `_start` is called during an interrupt, calling external functions causes the microcontroller to hang.
This is not noticeable during debugging, since everything works after flashing, but if the microcontroller is powered off, it won't start on the next power-up, not even after pressing the reset button.
This doesn't affect the `ch32v003`, as it has no reset interrupt, but I decided to make everything consistent.

Tested on: CH32V003 and CH32V305

### DIFF
--- a/port/wch/ch32v/src/cpus/qingkev2-rv32ec.zig
+++ b/port/wch/ch32v/src/cpus/qingkev2-rv32ec.zig
@@ -27,7 +27,7 @@ pub inline fn wfe() void {
 pub const startup_logic = struct {
     extern fn microzig_main() noreturn;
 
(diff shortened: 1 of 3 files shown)


mattnite (approved):
Thanks for all these contributions!

ghostiam:
Thank you for the project.
I’ve resolved the conflicts.

PR #532, opened 2025-05-08
WCH CH32V: fix build 🤯

ghostiam:
It seems that in Docker the error disappeared after renaming the files, and locally the error doesn't happen at all...


mattnite (approved):
Yeah this is funky, I think zig 0.14.0 has some cache bugs

Issue #491, opened 2025-04-23
Add ram image (no flash) target for rp2xxx


Grazfather:
The bootrom on rp2xxx chips support UF2 images that run directly from flash. This should allow for faster dev work, and avoid wearing out the flash prematurely.

It seems that the pico-sdk already supports this, by setting `pico_set_binary_type(blink_simple no_flash)` in the _CMakeLists.txt_.

Trying to figure out how to do this, I figure it entails:

1. A custom linker, that puts .text, .rodata, .data, .bss all in RAM. No `NOLOAD` at all obviously.
2. Replace the stage 2 bootloader, though this might not be necessary. I believe this bootloader only configures flash for XIP which we would not care about.
3. Replace `startup_logic._start`, since we don't want to clear `.bss` and copy `.data` from flash.
  * Maybe we could set some configs that would avoid these steps. This would probably simplify things.
4. Make sure the image fits in ram. This would probably be detected while linking.
5. Make sure that VTOR gets setup correctly.
6. Make it so that we don't need `microzig_flash_start`, `microzig_data_start`, `microzig_bss_start`, `microzig_data_load_start` to be defined, or give them dummy unused values.

iirc, the bootrom just copies the uf2 to wherever the address says (whether that be flash or ram) and then jumps to the stage2 bootloader, which should respect the reset vector in the vector table.

I think that this means that we will need a custom stage2 bootloader which knows where the (ram) vector table is, and sets it. This implies that we would need to flash (to flash) a ram-image-aware bootloader, which would mean that we would want to hard-code a known location for the vector table in ram. We might need for this target to produce separate binaries then: One for the bootloader for flash, and another that is ram only.

Additionally, we'd like to be able to flash this in openocd/gdb. I am not sure if this copies a stub into memory, but if it does, we'd also need to make sure that we don't trample the stub.

Grazfather:
Have a POC working:

Makefile
```Makefile
SOURCES = blink.c
OBJECTS = $(SOURCES:.c=.o)
TARGET = rpram_blink

CC = arm-none-eabi-gcc
LD = arm-none-eabi-ld
CFLAGS = -mcpu=cortex-m0plus
LDFLAGS = -nostdlib -T rp2040_ram.ld -Map=output.map

.PHONY: all clean load

all: $(TARGET).uf2

rpblink.o: rpblink.c
	$(CC) $(CFLAGS) -c $< -o $@

$(TARGET).elf: $(OBJECTS)
[...]

PR #535, opened 2025-05-09
Add rp2040 ram image (flashless) target board

Grazfather:
This tries to implement #491.

Basically:

* Load everything starting at SRAM
* Skip copying `.data` and clearing `.bss`
* Set `VTOR` to point to the vector table in sram

This counts on the behaviour of the bootloader, which, after loading a uf2 will jump to the lowest address it loaded, where I place the stub that sets VTOR and jumps to the entry point.

Note that because of a bug (ziglang/zig#23869), if you're building this target, you should do a clean build to properly pick up updates.


examples/raspberrypi/rp2xxx/build.zig
```diff
@@ -23,6 +23,8 @@ pub fn build(b: *std.Build) void {
         .{ .target = mb.ports.rp2xxx.boards.raspberrypi.pico, .name = "pico_hd44780", .file = "src/rp2040_only/hd44780.zig" },
         .{ .target = mb.ports.rp2xxx.boards.raspberrypi.pico, .name = "pico_pcf8574", .file = "src/rp2040_only/pcf8574.zig" },
         .{ .target = mb.ports.rp2xxx.boards.raspberrypi.pico, .name = "pico_i2c_slave", .file = "src/rp2040_only/i2c_slave.zig" },
+        // TODO: Fix
```
Grazfather, line 26: tbd: Do we want to build a bunch of the examples with the ram image target?

port/raspberrypi/rp2xxx/rp2040_ram_image.ld
```diff
@@ -0,0 +1,52 @@
+/*
+ * Target CPU:  ARM Cortex-M0+
+ * Target Chip: RP2040
+ */
+
+ENTRY(_entry_point);
```
Grazfather, line 6: This doesn't work! The elf entry point is incorrect, but the bootloader doesn't care and jumps to the lowest address flashed, which is 0x20000000

build.zig
```diff
@@ -332,6 +332,9 @@ pub fn MicroBuild(port_select: PortSelect) type {
             /// If set, overrides the `linker_script` property of the target.
             linker_script: ?LazyPath = null,
 
+            /// If set, overrides the default `entry` property of the arget.
+            entry: ?Build.Step.Compile.Entry = null,
```
Grazfather, line 336: This is in the firmware options... maybe this doesn't make sense to be here, but we probably at least want it in the target.

port/raspberrypi/rp2xxx/rp2040_ram_image.ld
Grazfather: Fixed this by setting it in the target via 8d07e064 and 69733398

core/src/cpus/cortex_m.zig
```diff
@@ -655,14 +674,28 @@ pub const startup_logic = struct {
     };
 };
 
+fn is_ramimage() bool {
+    // HACK
+    // TODO: Use microzig_options?
+    if (microzig.config.board_name) |board_name|
+        return std.mem.containsAtLeast(u8, board_name, 1, "ram image");
+    return false;
```
tact1m4n3, line 682: I guess you can use `CPU_Options` for this

mattnite (approved):
I think entry point is okay there for now, we can refactor later

core/src/cpus/cortex_m.zig
mattnite: I think this is something that warrants a different PR since it's something we can add as a general development method.

Grazfather: I have a branch experimenting with adding similar functionality to the esp32c3, so yeah.
I agree it isn't CPU specific.
I'll look into adding it as a field in `Target` and exposing it via `config`.

Issue #405, opened 2025-03-09
rp2350: Support extra gpios


Grazfather:
We hardcode 30 GPIOs in e.g. pins.zig, but there are more GPIOs on the rp2350 in the QFN80 package.

curuvar:
I suppose the proper way to do this would be to add .RP2350_QNF80 to the list of RP chip types.  Then select between one of two Pin enums at comptime based on this value.

Doing a quick look, there are a lot of places where the code switches on this value that would have to be updated though.

Issue #311, opened 2024-12-10
[RP2xxx] Configure and access ADC from pins helper


MatthiasPortzel:
The ADC integration that was in the pins helper had completely bit-rotted, so in #303 I replaced it with FIXME comments. I think the change is trivial, I'm just not at all familiar with the ADC on the RP2040. We need to replace `.init()` and then write a quick test program.

MatthiasPortzel:
Sounds like this was closed by #540.

PR #540, opened 2025-05-13
Rp2350 support improvement

Uthedris:
Addresses #405, #311 
Supersedes PR #531.

Improves support for RP2350, especially the RP2350B.  Some of the changes:

- Adds support for GPIO pins 30 through 47.
- Adds support for extra ADC pins and the change in location of the ADCs (pin 40-47 vs 26-30)
- Adds basic support for external PSRAM (R/W memory located at 0x1100_0000)
- Adds UART alternate RX an TX pins (also for RP2350A)
- Adds support in `pins.zig` for PI02. (it looks like `pio.zig` already has needed changes) 
- Adds support in `pins.zig` for PWM8 through PWM11 (changes to `pwm.zig` may be needed)
- Adds support for the Adafruit Meto RP2350 board (that uses the RP2350B chip).

Some things that the RP2350 supports that are not (yet) addressed:
- HSTX support (pin type is set in `pins.zig`)
- TRNG support
- Always-On Timer changes from Real Time Clock.
- Repurposing USB and QSPI pins
- Probably others.

@Grazfather was right about the added complexity in adding a new chip type.  This version currently just looks for a specific declaration in the board file to switch over to the rp2340B.  There is probably a better way to do that, but as it's just a flag that is used in relatively few place, it should be easy to change.

It does supply unique `function_table` entries for each chip version, as that was needed due to the change in pin location of the ADCs.  This version supplies unique lists of compatible pins for RP2040, RP2350A and RP2350B.

### DIFF
--- a/port/raspberrypi/rp2xxx/build.zig
+++ b/port/raspberrypi/rp2xxx/build.zig
@@ -10,6 +10,9 @@ chips: struct {
 },
 
 boards: struct {
+    adafruit: struct {
+        metro_rp2350: *const microzig.Target,
+    },
     raspberrypi: struct {
         pico: *const microzig.Target,
         pico2_arm: *const microzig.Target,
@@ -142,6 +145,15 @@ pub fn init(dep: *std.Build.Dependency) Self {
             .rp2350_riscv = chip_rp2350_riscv.derive(.{}),
         },
         .boards = .{
+            .adafruit = .{
+                .metro_rp2350 = chip_rp2350_arm.derive(.{
+                    .board = .{
+                        .name = "Adafruit Metro RP2350",
+                        .url = "https://www.adafruit.com/product/5542",
+                        .root_source_file = b.path("src/boards/adafruit_metro_rp2350.zig"),
+                    },
+                }),
+            },
             .raspberrypi = .{
                 .pico = chip_rp2040.derive(.{
                     .board = .{
--- a/port/raspberrypi/rp2xxx/build.zig
+++ b/port/raspberrypi/rp2xxx/build.zig
@@ -149,7 +149,7 @@ pub fn init(dep: *std.Build.Dependency) Self {
                 .metro_rp2350 = chip_rp2350_arm.derive(.{
                     .board = .{
                         .name = "Adafruit Metro RP2350",
-                        .url = "https://www.adafruit.com/product/5542",
+                        .url = "https://www.adafruit.com/product/6267",
                         .root_source_file = b.path("src/boards/adafruit_metro_rp2350.zig"),
                     },
                 }),
--- a/port/raspberrypi/rp2xxx/src/hal/adc.zig
+++ b/port/raspberrypi/rp2xxx/src/hal/adc.zig
@@ -33,8 +33,11 @@ pub const Config = struct {
     /// each input. So for 4 inputs in round-robin mode you'd see 1/4 sample
     /// rate for a given put vs what is set here.
     sample_frequency: ?u32 = null,
+    /// Configure pins to be used in round-robin mode.
     round_robin: ?InputMask = null,
+    /// Configure the ADC FIFO.
     fifo: ?fifo.Config = null,
+    /// Enable the temperature sensor.
     temp_sensor_enabled: bool = false,
 };
 
@@ -42,7 +45,7 @@ pub const Config = struct {
 /// CS.EN = 1. The global clock configuration is not needed to configure the
 /// sample rate because the ADC hardware block requires a 48MHz clock.
 pub fn apply(config: Config) void {
-   ADC.CS.write(.{
+    ADC.CS.write(.{
         .EN = 0,
         .TS_EN = @intFromBool(config.temp_sensor_enabled),
         .START_ONCE = 0,
--- a/port/raspberrypi/rp2xxx/src/hal/adc.zig
+++ b/port/raspberrypi/rp2xxx/src/hal/adc.zig
@@ -10,6 +10,8 @@ const resets = @import("resets.zig");
 const clocks = @import("clocks.zig");
 const chip = @import("compatibility.zig").chip;
 
+const has_rp2350b = chip == .RP2350 and @hasDecl(microzig.board, "has_rp2350b") and microzig.board.has_rp2350b;
+
 pub const Error = error{
     /// ADC conversion failed, one such reason is that the controller failed to
     /// converge on a result.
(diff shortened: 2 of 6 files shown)


port/raspberrypi/rp2xxx/build.zig
```diff
@@ -141,6 +144,15 @@ pub fn init(dep: *std.Build.Dependency) Self {
             .rp2350_riscv = chip_rp2350_riscv.derive(.{}),
         },
         .boards = .{
+            .adafruit = .{
+                .metro_rp2350 = chip_rp2350_arm.derive(.{
+                    .board = .{
+                        .name = "Adafruit Metro RP2350",
+                        .url = "https://www.adafruit.com/product/5542",
```
Grazfather, line 151: This URL points to coloured knobs for me.

port/raspberrypi/rp2xxx/src/hal/adc.zig
```diff
@@ -92,42 +77,80 @@ pub fn select_input(in: Input) void {
...
-
-    /// Get the corresponding GPIO pin for an ADC input. Panics if you give it
-    /// temp_sensor.
-    pub fn get_gpio_pin(in: Input) gpio.Pin {
-        return switch (in) {
-            else => gpio.num(@as(u5, @intFromEnum(in)) + 26),
-            .temp_sensor => @panic("temp_sensor doesn't have a pin"),
-        };
-    }
+/// For RP2040 and RP2350A  , the values are:
```
Grazfather, line 86: formatting

port/raspberrypi/rp2xxx/src/hal/pins.zig
```diff
@@ -42,19 +54,35 @@ pub const Pin = enum {
     GPIO27,
     GPIO28,
     GPIO29,
+    GPIO30, // RP2340B only
```
Grazfather, line 57: 2350b?

port/raspberrypi/rp2xxx/src/hal/pins.zig
```diff
@@ -5,12 +5,24 @@ const StructField = std.builtin.Type.StructField;
...
 
 const gpio = @import("gpio.zig");
 const pwm = @import("pwm.zig");
 const adc = @import("adc.zig");
 const resets = @import("resets.zig");
 
+const compatibility = @import("compatibility.zig");
+const chip = compatibility.chip;
+
+const has_rp2350b = chip == .RP2350 and @hasDecl(microzig.board, "has_rp2350b") and microzig.board.has_rp2350b;
```
Grazfather, line 18: Can this go into compatibility? Same with the one in adc.zig

port/raspberrypi/rp2xxx/src/hal/pins.zig
```diff
@@ -76,7 +114,7 @@ pub const Function = enum {
 
     PIO0,
     PIO1,
-    // TODO: PIO2 for RP2350
+    PIO2, // RP2340 only
```
Grazfather, line 117: 2350

port/raspberrypi/rp2xxx/src/hal/pins.zig
```diff
@@ -262,82 +459,256 @@ pub const Function = enum {
     }
 };
 
-fn all() [30]u1 {
-    var ret: [30]u1 = undefined;
-    for (&ret) |*elem|
-        elem.* = 1;
+fn all() [48]u1 {
```
Grazfather, line 462: have you consider having the type of the return here be comptime is has_rp2350b? could still use splat in that case

port/raspberrypi/rp2xxx/src/hal/pins.zig
```diff
@@ -262,82 +460,249 @@ pub const Function = enum {
...
+        single(26), // ADC0
+        single(27), // ADC1
+        single(28), // ADC2
+        single(29), // ADC3
+        none(), // ADC4
+        none(), // ADC5
+        none(), // ADC6
+        none(), // ADC7
+        none(), // QMI_CS1
+        none(), // HSTX
```
Grazfather, line 556: doesn't the normal 2350 package still have HSTX on GPIO 12-19?

port/raspberrypi/rp2xxx/src/hal/pins.zig
```diff
@@ -262,82 +460,249 @@ pub const Function = enum {
...
+        list(&.{ 2, 5, 8, 11, 14, 17, 20, 23, 26, 29 }), // USB_VBUS_EN
+        single(26), // ADC0
+        single(27), // ADC1
+        single(28), // ADC2
+        single(29), // ADC3
+        none(), // ADC4
+        none(), // ADC5
+        none(), // ADC6
+        none(), // ADC7
+        none(), // QMI_CS1
```
Grazfather, line 555: Is this not on normal 2350?

Grazfather (commented):
Nice work here. Have a few Qs

port/raspberrypi/rp2xxx/src/hal/pins.zig
Uthedris: It is on the RP2350A (see line 704) -- You're looking at the RP2040 function table.  

There are three versions of the function table.  In order RP2040, RP2350B and RP2350A.  I built separate tables for RP2040 and RP2350A so that trying to configure  nonexistent pins fails with the "pin cannot be configured for ..." message.

port/raspberrypi/rp2xxx/src/hal/pins.zig
Uthedris: Same comment.  Line 703.

port/raspberrypi/rp2xxx/src/hal/pins.zig
Grazfather: Oops yeah I misread the part we were in :)

mattnite (approved):

port/raspberrypi/rp2xxx/src/hal/gpio.zig
```diff
@@ -89,7 +94,7 @@ pub const Pull = enum {
     disabled,
 };
 
-pub fn num(n: u6) Pin {
+pub fn num(n: u9) Pin {
```
piotrfila, line 97: Was this intentional? Another type uses u9, but the maximum number of GPIOs is 48, so shouldn't a u6 suffice?
I know this is a nit pick since zig automatically widens integers, but I was confused when I saw this.

Uthedris: Nope, I have no clue why I changed it  Clearly it should be u6.

PR #545, opened 2025-05-15
Rp2040 atomics

Uthedris:
This adds the missing RP2040 atomic external definitions to support Zig's `@atomicLoad`, `@atomicRmw`, `@aromicStore`, `@cmpxchgStrong` and `@cmpxchgWeak`. operations.

### DIFF
--- a/port/raspberrypi/rp2xxx/src/hal/atomic.zig
+++ b/port/raspberrypi/rp2xxx/src/hal/atomic.zig
@@ -1,6 +1,11 @@
 // Based on: https://github.com/ziglang/zig/blob/79460d4a3eef8eb927b02a7eda8bc9999a766672/lib/compiler_rt/atomics.zig
 // and: https://github.com/raspberrypi/pico-sdk/blob/ee68c78d0afae2b69c03ae1a72bf5cc267a2d94c/src/rp2_common/pico_atomic/atomic.c
 
+//! Atomic operations for RP2040
+//!
+//! These functions should not be called directly. Instead, use the Zig atomic
+//! builtins or `std.atomic.Value`.
+
 const std = @import("std");
 const builtin = @import("builtin");
 const microzig = @import("microzig");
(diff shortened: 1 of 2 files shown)


port/raspberrypi/rp2xxx/src/hal/atomic.zig
```diff
@@ -15,60 +23,193 @@ inline fn atomic_unlock(critical_section: CriticalSection) void {
...
-// TODO: export all the rest atomics
+export fn __atomic_load_4(ptr: *u32, model: i32) callconv(.c) u32 {
+    return atomic_load(u32, ptr, model);
+}
 
-comptime {
-    // This export should only happen for the RP2040 due to the ARMv6-M ISA used by the ARM Cortex-M0+.
-    if (chip == .RP2040) {
-        @export(&__atomic_fetch_and_4, .{ .name = "__atomic_fetch_and_4", .linkage = linkage, .visibility = visibility });
-        @export(&__atomic_fetch_or_4, .{ .name = "__atomic_fetch_or_4", .linkage = linkage, .visibility = visibility });
```
mattnite, line 70: What about this note regarding the rp2040? Should these functions be exported for the rp2350?

arkadiuszwojcik: @mattnite RP2350 uses newer ARM ISA (ARMv8) where all those operations are dedicated instructions so there is no need to emulate atomics this way

arkadiuszwojcik: Yes, good question. RP2350 don't need those symbols as all those operations are made by dedicated instructions.

arkadiuszwojcik:
@Uthedris so `_ = @import("hal/atomic.zig");` from `hal.zig` can be removed right now? I had to use it so conditional comptime export would work. Now chip check in that file is removes so maybe this line is not required?

Uthedris:
@arkadiuszwojcik  the guard against being in test mode still seems to be required.  When I removed it, I could build my a program that used atomics just fine, but the GitHub test suite failed the Unit Test Ports tests.

mattnite (approved):

PR #546, opened 2025-05-15
Rp2350 hal additions

Uthedris:
This adds support for the RP2350 true random number genrator and the always-on timer.
One of the uses if the always on timer is to support a realtime clock with alarm.

This also adds a DateTime class that can be used to parse timestamp values into years, months days, etc., and to format this data as a text string.


mattnite:
This is looking good, can you add some example programs for these new modules?

PR #552, opened 2025-05-18
esp: join cpu and add rom functions

tact1m4n3:
- merge esp riscv cpu separate files for image and direct boot into one
- add rom functions
- add `cat` util to merge `rom_functions.ld` with the base linker script

The `cat` util can be placed in the root tools folder if it's needed by multiple ports.


port/espressif/esp/src/hal/rom.zig
```diff
@@ -10,9 +10,9 @@ pub fn delay_us(us: u32) void {
 
 pub const functions = switch (compatibility.chip) {
     .esp32_c3 => struct {
-        pub const ets_delay_us: *const fn (us: u32) callconv(.c) void = @ptrFromInt(0x40000050);
-        pub const ets_update_cpu_frequency: *const fn (ticks_per_us: u32) callconv(.c) void = @ptrFromInt(0x40000588);
-        pub const rom_i2c_readReg: *const fn (block: u8, block_hostid: u8, reg_add: u8) callconv(.c) u8 = @ptrFromInt(0x40001954);
-        pub const rom_i2c_writeReg: *const fn (block: u8, block_hostid: u8, reg_add: u8, indata: u8) callconv(.c) void = @ptrFromInt(0x4000195c);
+        pub extern fn ets_delay_us(us: u32) callconv(.c) void;
+        pub extern fn ets_update_cpu_frequency(ticks_per_us: u32) callconv(.c) void;
+        pub extern fn esp_rom_regi2c_read(block: u8, block_hostid: u8, reg_add: u8) callconv(.c) u8;
```
Grazfather, line 15: why do we name these with a redundant `esp_rom_` prefix? I think that the path of the linker file plus comments makes it clear what sort of functions they are.

port/espressif/esp/src/hal/rng.zig
```diff
@@ -0,0 +1,16 @@
...
+const RNG = microzig.chip.peripherals.RNG;
+
+pub fn random_u32() u32 {
+    return RNG.DATA;
+}
+
+pub fn read(buf: []u8) void {
+    var i: usize = 0;
+    while (i < buf.len) : (i += 4) {
+        const count = @max(buf.len - i, 4);
```
Grazfather, line 12: do you mean min here? If buf is, say, 7 bytes long, then the second iteration will have `count = 4` and `i = 4`, and the memcpy will write out of bounds.

port/espressif/esp/src/cpus/esp_riscv.zig
```diff
@@ -232,6 +233,130 @@ pub const interrupt = struct {
 pub const nop = riscv32_common.nop;
 pub const wfi = riscv32_common.wfi;
 
+pub const startup_logic = switch (cpu_config.boot_mode) {
+    .direct => struct {
+        comptime {
```
Grazfather, line 238: why is this a comptime block? Could we do more similar to the cortex_m.zig stuff and make this a structure (like `_vector_table`, and then export it like in `export_startup_logic`?

tact1m4n3: This is the way it was done originally. But I can change it.

Grazfather:
We can do either style, but ideally not two styles, and it can be done in a
follow up

port/espressif/esp/src/hal/rom.zig
tact1m4n3: That's the name used by esp-idf and esp-hal. Those are also global symbols so imo it's okay that they have a prefix. Plus there is an endless list of them and renaming each one would take forever (they are defined in the linker script).

port/espressif/esp/src/hal/rng.zig
tact1m4n3: Oh... right 😄

port/espressif/esp/src/cpus/esp_riscv.zig
```diff
@@ -232,6 +233,130 @@ pub const interrupt = struct {
...
+            // Direct Boot: does not support Security Boot and programs run directly in flash. To enable this mode, make
+            // sure that the first two words of the bin file downloading to flash (address: 0x42000000) are 0xaedb041d.
+
+            // In this case, the ROM bootloader sets up Flash MMU to map 4 MB of Flash to
+            // addresses 0x42000000 (for code execution) and 0x3C000000 (for read-only data
+            // access). The bootloader then jumps to address 0x42000008, i.e. to the
+            // instruction at offset 8 in flash, immediately after the magic numbers.
+
+            asm (
+                \\.extern _start
```
Grazfather, line 251: Is this needed?

tact1m4n3: Nope

mattnite (approved):

PR #558, opened 2025-05-20
nRF52xxx: Add UART hal + examples

mattnite:


port/nordic/nrf5x/src/hal/gpio.zig
```diff
@@ -0,0 +1,85 @@
+const microzig = @import("microzig");
+const peripherals = microzig.chip.peripherals;
+const P0 = peripherals.P0;
+const GPIOTE = microzig.GPIOTE;
+
+pub const Direction = enum(u1) {
+    input = 0x0,
```
Grazfather, line 7: rp2xxx uses `.in` and `.out`

port/nordic/nrf5x/src/hal/gpio.zig
```diff
@@ -0,0 +1,85 @@
...
+pub const Sense = enum(u2) {
+    disabled = 0x0,
+    /// sense high level
+    high = 0x2,
+    /// sense low level
+    low = 0x3,
+    _,
+};
+
+pub const Drive = @FieldType(P0.PIN_CNF[0].underlaying_type, "DRIVE");
```
Grazfather, line 20: typo with underlying type, but also, I never got this working. I had to expose the type via a patch.

If we want this to look like rp2xxx we might want to call it DriveStrength... but then again, it's not really the same.

port/nordic/nrf5x/src/hal/gpio.zig
```diff
@@ -0,0 +1,85 @@
...
+
+pub const InputBuffer = enum(u1) {
+    connect = 0x0,
+    disconnect = 0x1,
+};
+
+pub const Pull = enum(u2) {
+    disabled = 0x0,
+    pulldown = 0x1,
```
Grazfather, line 29: rp2xxx uses up, down, disabled. better if we unify imo

port/nordic/nrf5x/src/hal/gpio.zig
```diff
@@ -0,0 +1,85 @@
...
+    pullup = 0x3,
+    _,
+};
+
+pub fn num(n: u5) Pin {
```
Grazfather, line 34: this would have to be chip specific for chips with other banks. I wrote mine to take two args, maybe we could write two functions.

port/nordic/nrf5x/src/hal/gpio.zig
```diff
@@ -0,0 +1,85 @@
...
+
+    /// Set GPIO pin as input or output
+    pub inline fn set_direction(pin: Pin, direction: Direction) void {
+        switch (direction) {
+            .input => P0.DIRCLR.raw = pin.mask(),
+            .output => P0.DIRSET.raw = pin.mask(),
+        }
+    }
+    pub inline fn set_sense(pin: Pin, sense: Sense) void {
+        P0.PIN_CNF[@intFromEnum(pin)].modify(.{
```
Grazfather, line 49: it'll make my PR on top cleaner if you use `get_regs` instead of hardcoding `P0` everywhere.

port/nordic/nrf5x/patches/nrf52832.zig
```diff
@@ -0,0 +1,39 @@
+const Patch = @import("microzig/build-internals").Patch;
```
Grazfather, line 1: (so far) we could use the same patch file as 52840

PR #561, opened 2025-05-23
nRF5x: Add i2c hal + datagram device

Grazfather:
This adds a HAL for the i2c master peripheral on the nRF5x, both for the byte-banged and the DMA version.

It adds an example of a bus scan and reading from a TMP117 temperature scanner, both using both HALs


port/nordic/nrf5x/src/hal/drivers.zig
```diff
@@ -0,0 +1,111 @@
+//!
```
Grazfather, line 1: This file is literally identical to the rp2xxx one! That's the benefit of copying its interface.

port/raspberrypi/rp2xxx/src/hal/i2c.zig
```diff
@@ -195,6 +195,7 @@ test "i2c.translate_baudrate" {
     try std.testing.expectError(ConfigError.InputFreqTooLow, translate_baudrate(1_000_000, 31_900_000));
 }
 
+// TODO: Why create this empty struct?
```
Grazfather, line 198: I will remove this, but I am wondering why we do this. My hal does not. I don't think it's necessary?

mattnite (approved):

PR #562, opened 2025-05-24
Fixes to interrupt handling

Uthedris:
Fixes:
  - riscv32_common.zig `is_pending` function need a compare to return a `bool`.
  - hazard3.zig interrupt priority registers were incorrect.
  - hazard3.zig removed an interrupt enable that tact1m4n3 didn't like.


tact1m4n3 (changes requested):
Nice catch! Don't know why I used `meifa` to set priorities :laughing:. Also, some riscv examples are broken, please change the examples using external interrupts to enable the `MachineExternal` interrupt on riscv.

tact1m4n3 (approved):

PR #570, opened 2025-05-28
esp32 Trap Dispatch Improvement

Uthedris:
This makes the esp32 trap dispatch a bit more efficient.

The current version initially allocates a 160 byte stack frame (4*40) even though only 140 bytes (4*35) are used.   It then allocates an additional 16 bytes which are only used to preserve various values over function calls.

This change reduces the initial allocation to the required 140 bytes.  It eliminates the second 16 byte allocation entirely, using "s" registers to preserve values.  The "s" registers are themselves preserved by callconv(.c) functions.


tact1m4n3 (approved):
Great patch!

PR #572, opened 2025-05-28
drivers: Add ICM-20948 9dof sensor driver

Grazfather:


Grazfather:
Hah! Coincidence! But so am I! I was just searching for an i2c device that 1. wasn't too expensive 2. had a QWIIC breakout board 3. Could be useful to me.

Sounds like you'll be able to help me test :D

mattnite (approved):

PR #576, opened 2025-05-30
Clock_Device: Allow providing sleep function to avoid polling

Grazfather:
Also update the test device to crucially provide this (as a no-op) so that we can sleep call driver methods that sleep in tests without hanging.


tact1m4n3 (approved):
Cool!

mattnite (approved):

PR #578, opened 2025-05-31
Driver for the mlx90640 thermal imaging camera

cswank:
There are tons of calculations that need to be done in order to turn the output of the device into actual temperature values.  Those calculations were copied from https://github.com/netzbasteln/MLX90640-Thermocam and mostly do not follow the zig naming conventions.


mattnite:
Really cool contribution! It's up to you if you'd like to fix the casing, we can always revisit it later

drivers/framework.zig
```diff
@@ -38,6 +38,7 @@ pub const input = struct {
 
 pub const sensor = struct {
     pub const TMP117 = @import("sensor/TMP117.zig").TMP117;
+    pub const MLX90640 = @import("sensor/MLX90640.zig").MLX90640;
```
Grazfather, line 41: go ahead and merge/rebase since I just merged a PR that will conflict on this line (I added another sensor driver)

drivers/sensor/MLX90640.zig
```diff
@@ -0,0 +1,882 @@
...
+            @as(u8, @truncate(address & 0xFF)),
+            @as(u8, @truncate(val >> 8)),
+            @as(u8, @truncate(val & 0xFF)),
+        };
+
+        try self.i2c.write(&req);
+    }
+
+    fn writeThenRead(self: *Self, address: u16, buf: []u16) !void {
+        var req = [2]u8{ @as(u8, @truncate(address >> 8)), @as(u8, @truncate(address & 0xFF)) };
```
Grazfather, line 105: you can do `std.mem.nativeToBig` here, then do `std.mem.asBytes` to convert it to a byte slice

drivers/sensor/MLX90640.zig
```diff
@@ -0,0 +1,882 @@
...
+            @as(u8, @truncate(address >> 8)),
+            @as(u8, @truncate(address & 0xFF)),
+            @as(u8, @truncate(val >> 8)),
+            @as(u8, @truncate(val & 0xFF)),
+        };
+
+        try self.i2c.write(&req);
+    }
+
+    fn writeThenRead(self: *Self, address: u16, buf: []u16) !void {
```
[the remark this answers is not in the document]
Grazfather: You could make an enum for the address here to ensure you don't pass another value type

drivers/sensor/MLX90640.zig
```diff
@@ -0,0 +1,882 @@
...
+
+    pub fn setRefreshRate(self: *Self, rate: u16) !void {
+        try self.writeThenRead(self.control_register_address, self.frame[0..1]);
+        const val: u16 = (self.frame[0] & ~self.refresh_rate_mask) | ((rate << 7) & self.refresh_rate_mask);
+        try self.write(self.control_register_address, val);
+    }
+
+    pub fn refreshRate(self: *Self) !u3 {
+        try self.writeThenRead(self.control_register_address, self.frame[0..1]);
+        const val = self.frame[0] >> 7 & 0b111;
```
Grazfather, line 127: Could just define a type for this register read, bit cast it, then return the relevant field

drivers/sensor/MLX90640.zig
```diff
@@ -0,0 +1,882 @@
...
+        if (mode == cmee) {
+            irDataCP[1] = irDataCP[1] - cpo * (1 + self.params.cpKta * (ta - 25)) * (1 + self.params.cpKv * (vdd - 3.3));
+        } else {
+            irDataCP[1] = irDataCP[1] - (cpo + self.params.ilChessC[0]) * (1 + self.params.cpKta * (ta - 25)) * (1 + self.params.cpKv * (vdd - 3.3));
+        }
+
+        var range: u8 = 0;
+        var pattern: i32 = 0;
+        var pixelNumber: i32 = 0;
+        for (0..768) |i| {
```
Grazfather, line 201: What is 768? Would like to avoid magic numbers

drivers/sensor/MLX90640.zig
```diff
@@ -0,0 +1,882 @@
...
+                    }
+                }
+            }
+
+            for (0..outlierPixCnt) |x| {
+                pixCnt = @intCast(x);
+                for (pixCnt + 1..outlierPixCnt) |i| {
+                    warn = self.checkAdjacentPixels(self.params.outlierPixels[pixCnt], self.params.outlierPixels[i]);
+                    if (warn != 0) {
+                        //Serial.println("Outlier pixel has adjacent outlier pixel");
```
Grazfather, line 837: remove or convert to `std.log.err`.

drivers/sensor/MLX90640.zig
```diff
@@ -0,0 +1,882 @@
...
+                brokenPixCnt = brokenPixCnt + 1;
+            } else if ((self.eeprom[pixCnt + 64] & 0x0001) != 0) {
+                self.params.outlierPixels[outlierPixCnt] = @intCast(pixCnt);
+                outlierPixCnt = outlierPixCnt + 1;
+            }
+
+            pixCnt = pixCnt + 1;
+        }
+
+        var warn: i16 = 0;
```
Grazfather, line 807: We don't need to use numbers here... this isn't C :)

drivers/sensor/MLX90640.zig
cswank: Cool, I was hoping to learn better ways of doing some of this stuff.

drivers/sensor/MLX90640.zig
cswank: Oh yeah, I was copying so much c++ this was bound to happen.

drivers/sensor/MLX90640.zig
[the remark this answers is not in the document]
Grazfather: Try throwing claude at it. Once I got my driver working I used Claude to do the error handling and it (almost) nailed it.

drivers/sensor/MLX90640.zig
```diff
@@ -0,0 +1,896 @@
...
+        }
+        if (pixPosDif > 30 and pixPosDif < 34) {
+            return -6;
+        }
+
+        return 0;
+    }
+};
+
+test "refresh rate" {
```
cswank: When I run zig test I get 

`drivers/sensor/MLX90640.zig:12:22: error: import of file outside module path: '../framework.zig'
const base = @import("../framework.zig").base;`

I can run other tests, I'm not sure why mine is different.

cswank: Aha!  The tests should be in the MLX90640 struct.  I don't really understand why, but I got them working.

cswank: Nope, I was wrong.  Zero tests are running when I moved this, but it compiles.  I'll keep digging.

examples/raspberrypi/rp2xxx/build.zig
```diff
@@ -25,6 +25,7 @@ pub fn build(b: *std.Build) void {
         .{ .target = raspberrypi.pico, .name = "pico_hd44780", .file = "src/rp2040_only/hd44780.zig" },
         .{ .target = raspberrypi.pico, .name = "pico_pcf8574", .file = "src/rp2040_only/pcf8574.zig" },
         .{ .target = raspberrypi.pico, .name = "pico_i2c_slave", .file = "src/rp2040_only/i2c_slave.zig" },
+        .{ .target = raspberrypi.pico, .name = "pico_mlx90640", .file = "src/mlx90640.zig" },
```
Grazfather: You moved the file but you also have to move this line into `chip_agnostic_examples`

drivers/sensor/MLX90640.zig
```diff
@@ -0,0 +1,922 @@
...
+    scratch_data: [768]f32 = undefined,
+    i2c: mdf.base.Datagram_Device,
+    clock_device: mdf.base.Clock_Device,
+    params: parameters,
+    emissivity: f32,
+    open_air_shift: f32,
+
+    scale_alpha: f32 = 0.000001,
+
+    frame_loop: [2]u1 = [2]u1{ 0, 1 },
```
Grazfather:
```suggestion
    frame_loop: [2]u1 = .{ 0, 1 },
```

drivers/sensor/MLX90640.zig
```diff
@@ -0,0 +1,922 @@
...
+    }
+
+    fn write(self: *Self, reg: Self.registers, val: u16) !void {
+        const addr: u16 = @intFromEnum(reg);
+        var req = [4]u8{
+            @as(u8, @truncate(addr >> 8)),
+            @as(u8, @truncate(addr & 0xFF)),
+            @as(u8, @truncate(val >> 8)),
+            @as(u8, @truncate(val & 0xFF)),
+        };
```
Grazfather:
```suggestion
        var req = struct{addr_be: u16, val_be: u16}{
           .addr_be = std.mem.nativeToBig(addr),
           .var_be = std.mem.nativeToBig(val),
       };
```

Then `self.i2c_write(std.mem.asBytes(&req));` below

drivers/sensor/MLX90640.zig
```diff
@@ -0,0 +1,922 @@
...
+        try self.i2c.write_then_read(req, self.frame_data[0 .. buf.len * 2]);
+        for (0.., buf) |i, _| {
+            buf[i] = std.mem.readInt(u16, self.frame_data[i * 2 .. (i * 2) + 2][0..2], .big);
+        }
+    }
+
+    fn read_control_register(self: *Self) !control_register {
+        try self.write_then_read(Self.registers.control, self.frame[0..1]);
+        const val: control_register = @bitCast(self.frame[0]);
+        return val;
```
Grazfather:
```suggestion
        return @bitCast(self.frame[0]);
```

drivers/sensor/MLX90640.zig
Grazfather: This is the resolution? 24*32? Might want to put these into constants.

drivers/sensor/MLX90640.zig
```diff
@@ -0,0 +1,922 @@
...
+            const conversionPattern: i32 = (@divTrunc((pixelNumber + 2), 4) - @divTrunc((pixelNumber + 3), 4) + @divTrunc((pixelNumber + 1), 4) - @divTrunc(pixelNumber, 4)) * (1 - 2 * ilPattern);
+
+            if (mode == 0) {
+                pattern = ilPattern;
+            } else {
+                pattern = chessPattern;
+            }
+
+            var irData: f32 = @floatFromInt(self.frame[i]);
+            if (irData > 32767) {
```
Grazfather, line 244: Doesn't this just mean that frame should have `i16` instead of `u16`?

drivers/sensor/MLX90640.zig
```diff
@@ -0,0 +1,922 @@
...
+            }
+
+            try self.write_then_read(Self.registers.ram, self.frame[0..832]);
+            ready = false;
+        }
+
+        try self.write_then_read(Self.registers.control, self.frame[832..833]);
+    }
+
+    fn getVdd(self: *Self) f32 {
```
Grazfather: get_vdd

drivers/sensor/MLX90640.zig
```diff
@@ -0,0 +1,922 @@
...
+        const resolutionRAM: f32 = @floatFromInt((self.frame[832] & 0x0C00) >> 10);
+        const resolutionCorrection = std.math.pow(f32, 2, @floatFromInt(self.params.resolutionEE)) / std.math.pow(f32, 2, resolutionRAM);
+        const vdd25: f32 = @floatFromInt(self.params.vdd25);
+        const kvdd: f32 = @floatFromInt(self.params.kVdd);
+        vdd = (resolutionCorrection * vdd - vdd25) / kvdd + 3.3;
+
+        return vdd;
+    }
+
+    fn getTa(self: *Self, vdd: f32) f32 {
```
Grazfather: get_ta

drivers/sensor/MLX90640.zig
cswank: Yes, this camera has 24x32 pixels.  I do plan on putting them as constants, just trying to think of a name (which is the hard part).

drivers/sensor/MLX90640.zig
cswank: Yes, it does seem like that.  The conversions that I copied have several instances where this type of logic happens.  I'll try changing to signed ints to see what happens after I get all this refactoring in a good place.

cswank: This logic, actually, is straight from the datasheet.  Throughout all of the datasheet's calculations there are examples of checking to see if a value is larger than some number then subtracting to make it right.  Due to the extreme complexity of these calculations I think I'd rather leave things as they are.

drivers/sensor/MLX90640.zig
Grazfather: WIDTH, HEIGHT, NUM_PIXELS = WIDTH * HEIGHT?

drivers/sensor/MLX90640.zig
Grazfather: You figured this out, right?

drivers/sensor/MLX90640.zig
Grazfather: Yeah. Maybe add a comment so the next guy who reads this doesn't think the same thing.

drivers/sensor/MLX90640.zig
cswank: Yes, I can run the tests now.

drivers/sensor/MLX90640.zig
cswank: I think this is the last thing I need to tackle.

drivers/sensor/MLX90640.zig
cswank: I'll add it above a bit where the first such instance occurs.

drivers/sensor/MLX90640.zig
```diff
@@ -0,0 +1,910 @@
...
+    clock_device: mdf.base.Clock_Device,
+    params: parameters,
+    emissivity: f32,
+    open_air_shift: f32,
+
+    scale_alpha: f32 = 0.000001,
+    pixel_count: u10 = 768,
+
+    frame_loop: [2]u1 = .{ 0, 1 },
+    refresh_rate_mask: u16 = 0b111 << 7,
```
Grazfather: Make this sort of stuff const so it doesn't take up space in the struct.

drivers/sensor/MLX90640.zig
```diff
@@ -0,0 +1,910 @@
...
+
+    pub fn init(cfg: MLX90640_Config) !Self {
+        return Self{
+            .i2c = cfg.i2c,
+            .clock_device = cfg.clock,
+            .open_air_shift = cfg.open_air_shift,
+            .emissivity = cfg.emissivity,
+            .params = parameters{},
+        };
```
Grazfather:
```suggestion
        return .{
            .i2c = cfg.i2c,
            .clock_device = cfg.clock,
            .open_air_shift = cfg.open_air_shift,
            .emissivity = cfg.emissivity,
            .params = .{},
        };
```
Shorthand!

drivers/sensor/MLX90640.zig
```diff
@@ -0,0 +1,910 @@
...
+        const val = try self.read_control_register();
+        return val.refresh_rate;
+    }
+
+    pub fn resolution(self: *Self) !u2 {
+        const val = try self.read_control_register();
+        return val.resolution;
+    }
+
+    pub fn temperature(self: *Self, result: []f32) !void {
```
Grazfather, line 157: why doesn't this retutrn anything?

drivers/sensor/MLX90640.zig
```diff
@@ -0,0 +1,910 @@
...
+
+        var kvScale: u8 = @intCast((self.eeprom[56] & 0x0F00) >> 8);
+
+        for (0..rows) |i| {
+            for (0..columns) |j| {
+                const p = 32 * i + j;
+                const split = 2 * (p / 32 - (p / 64) * 2) + p % 2;
+                self.scratch_data[p] = @floatFromInt(KvT[split]);
+                self.scratch_data[p] = self.scratch_data[p] / std.math.pow(f32, 2, @floatFromInt(kvScale));
+                //self.scratch_data[p] = self.scratch_data[p] * mlx90640->offset[p];
```
Grazfather: delete?

cswank: I'll delete it.  I think it was a commented out cpp line that I partially converted to zig and left it there to see if I could understand why it was there.

drivers/sensor/MLX90640.zig
cswank: It populates the 'result' array rather than returning an array.

cswank:
> I can't test this unfortunately, but it's looking good for now. We could do another cleanup to make it more zig like later on

I have one more test I've been working on this morning that tests the temperature() fn.  It has a few large arrays ([1668]u8 and another that's similar in size).  Any concerns about putting arrays that large in the test fn?  

This test will be very nice if someone ever makes the changes to make all those calculations more zig like.

EDIT:  I added it.  Let me know if you want to keep it.

Grazfather:
No worries about adding large arrays to tests -- They run on the host machine.

Issue #470, opened 2025-04-21
Generic Read-Write descriptor for driver framework/SPI


ikskuh:
Just dumping this so it isn't lost in the Discord:

An idea we could use for an abstract SPI transfer system:

```zig
fn transceivev(actions: []const Transaction) !usize;

const Transaction = struct {
  /// buffer to receive
  input: []u8 = &.{},
  /// index of first received byte
  input_offset: usize = 0,
  /// buffer to send
  output: []const u8 = &.{},
  /// index of first sent byte
  output_offset: usize = 0,
  /// transceive at least this amount of items
  min_len: usize = 0,
  /// output when total len bigger than output.len
  pad: u8 = 0x00,

  pub fn length(t: @This()) usize {
    return @max(
      t.min_length,
      t.input_offset + t.input.len,
      t.output_offset + t.output.len,
    );
  }
};
```

tact1m4n3:
This hasn't been added yet

PR #581, opened 2025-06-02
ESP32-C3 spi and WS2812 driver

tact1m4n3:
- add spi driver for esp32-c3
- add more gpio functions
- add helper function for enabling peripherals
- add ws2812 generic driver

I chose a slightly different approach when writing the spi hal (I stored the registers as a field and the type is no longer an enum). Part of the motivation was the calling of `get_regs` in every single function. Please let me know what you think about it.


drivers/led/ws2812.zig
```diff
@@ -0,0 +1,75 @@
+const std = @import("std");
+const mdf = @import("../framework.zig");
+
+// inspired by https://github.com/smart-leds-rs/ws2812-spi-rs/blob/68d98bf9c7f10bc852bace0f3fd5af2a1cef51a7/src/lib.rs
+
+/// WS2812 leds driver. The device should output at a frequency between 2MHz and 3.8MHz. While no
+/// bytes are written the output pin **must** be low.
+pub fn WS2812(options: struct {
+    max_led_count: usize = 1,
+    Datagram_Device: type = mdf.base.Datagram_Device,
```
Grazfather, line 10: (for later) I wonder if we create `Spi_Device` and similar,  to make it so that different hals can provide a unified Spi interface. In the case like this, it's clear that not just any Datagram device would work, for example. For one, the data line would stay high.

drivers/led/ws2812.zig
```diff
@@ -0,0 +1,75 @@
...
+        /// Writes the given colors in order to the device. The slice len **must** be less than the
+        /// maximum led count.
+        pub fn write(self: *Self, colors: []const Color) !void {
+            std.debug.assert(colors.len <= options.max_led_count);
+
+            // ensures that a reset takes place between writes
+            if (self.next_write_time) |next_write_time| {
+                const now = self.clock_dev.get_time_since_boot();
+                if (!next_write_time.is_reached_by(now)) {
+                    self.clock_dev.sleep_us(next_write_time.diff(now).to_us());
```
Grazfather, line 41: another todo for later: If our clock device is based on time since boot, we could easily add a `sleep_until` method.

port/espressif/esp/src/hal/system.zig
```diff
@@ -1,3 +1,4 @@
+const std = @import("std");
```
Grazfather: was this missing?

port/espressif/esp/src/hal/drivers.zig
```diff
@@ -110,6 +110,97 @@ pub const I2C_Device = struct {
...
+    pub fn readv(dev: SPI_Device, datagrams: []const []u8) ReadError!usize {
+        dev.bus.readv_blocking(datagrams, dev.bit_mode);
+        return microzig.utilities.Slice_Vector([]u8).init(datagrams).size();
+    }
+
+    const vtable = Datagram_Device.VTable{
+        .connect_fn = connect_fn,
+        .disconnect_fn = disconnect_fn,
+        .writev_fn = writev_fn,
+        .readv_fn = readv_fn,
```
Grazfather, line 212: We have since added `writev_then_readv_fn`. Can you add this?

port/espressif/esp/src/hal/drivers.zig
```diff
@@ -110,6 +110,97 @@ pub const I2C_Device = struct {
...
+        dev.bus.write_blocking(datagram, dev.bit_mode);
+    }
+
+    pub fn writev(dev: SPI_Device, datagrams: []const []const u8) WriteError!void {
+        dev.bus.writev_blocking(datagrams, dev.bit_mode);
+    }
+
+    pub fn read(dev: SPI_Device, datagram: []u8) ReadError!usize {
+        dev.bus.read_blocking(datagram, dev.bit_mode);
+        return datagram.len;
```
Grazfather: It might be useful to actually return this from read_blocking

port/espressif/esp/src/hal/drivers.zig
```diff
@@ -110,6 +110,97 @@ pub const I2C_Device = struct {
...
+    }
+
+    pub fn readv(dev: SPI_Device, datagrams: []const []u8) ReadError!usize {
+        dev.bus.readv_blocking(datagrams, dev.bit_mode);
+        return microzig.utilities.Slice_Vector([]u8).init(datagrams).size();
```
Grazfather: Same idea: Make readv_blocking return this.

port/espressif/esp/src/hal/spi.zig
```diff
@@ -0,0 +1,369 @@
...
+                    bestpre = pre;
+                }
+            }
+
+            const l: i32 = bestn;
+            var h: i32 = @divFloor(128 * bestn + 127, 256);
+            if (h <= 0) h = 1;
+
+            return @as(u32, @intCast(l - 1)) | (@as(u32, @intCast(h - 1)) << 6) |
+                (@as(u32, @intCast(bestn - 1)) << 12) | (@as(u32, @intCast(bestpre - 1)) << 18);
```
Grazfather: could you create an anonymous struct or something hear to make it more clear what's happening?
maybe even something like
```
return @bitCast(struct{x: uXX, y: uYY...}{x = @intCAst(l-1)...});
```

It looks like these are all u6 but we don't do anything to ensure e.g. that `l - 1` fits in 6 bits. Is this intentional? We could constrain the types above to make it easier and more clear.

port/espressif/esp/src/hal/spi.zig
```diff
@@ -0,0 +1,369 @@
...
+
+    pub fn readv_blocking(
+        self: SPI_Bus,
+        buffer_vec: []const []u8,
+        bit_mode: BitMode,
+    ) void {
+        const vec: Slice_Vector([]const u8) = .init(buffer_vec);
+        var iter = vec.iterator();
+
+        var remaining = vec.size();
```
Grazfather: could store and return this original value

port/espressif/esp/src/hal/spi.zig
```diff
@@ -0,0 +1,369 @@
...
+const hal = microzig.hal;
+const system = hal.system;
+
+const SPI_Regs = microzig.chip.types.peripherals.SPI2;
+
+pub const fifo_word_len = 16;
+pub const fifo_byte_len = fifo_word_len * @sizeOf(u32);
+
+pub const SPI_Instance = enum {
+    spi2,
```
Grazfather, line 20: Why only spi2?

port/espressif/esp/src/hal/system.zig
tact1m4n3: Nah! I did some changes that I reverted and didn't remove the imports. We should make the linter detect unused imports :))

port/espressif/esp/src/hal/drivers.zig
tact1m4n3: Sure

port/espressif/esp/src/hal/spi.zig
tact1m4n3: > could you create an anonymous struct or something hear to make it more clear what's happening? maybe even something like

Yep.

> It looks like these are all u6 but we don't do anything to ensure e.g. that `l - 1` fits in 6 bits. Is this intentional? 

Yeah. It's not possible to overflow here.

port/espressif/esp/src/hal/spi.zig
tact1m4n3: It's the only one available for user use in esp32-c3. There is also spi0 and spi1, but those are used internally (idk for what, maybe flash or smth). At some point, I disabled their clocks on bootup and I didn't know why my program froze :)

Grazfather: OK thanks

port/espressif/esp/src/hal/system.zig
Grazfather: That's a good idea

Grazfather: #583

port/espressif/esp/src/hal/drivers.zig
tact1m4n3: Why do we return the transfer length only for read and not for write?

drivers/led/ws2812.zig
tact1m4n3: Imo it's necessary to add something like this. For instance we can't write drivers rn for spi full duplex devices.

tact1m4n3: Also checkout #470

tact1m4n3:
Should be ready!

drivers/led/ws2812.zig
Grazfather: Exactly. Glad we have a ticket for it.

drivers/led/ws2812.zig
```diff
@@ -0,0 +1,75 @@
...
+/// bytes are written the output pin **must** be low.
+pub fn WS2812(options: struct {
+    max_led_count: usize = 1,
+    Datagram_Device: type = mdf.base.Datagram_Device,
+    Clock_Device: type = mdf.base.Clock_Device,
+}) type {
+    return struct {
+        const Self = @This();
+
+        // we can fit two bytes into one bit
```
Grazfather: two LED colours into one byte?

port/espressif/esp/src/hal/drivers.zig
```diff
@@ -110,6 +110,123 @@ pub const I2C_Device = struct {
...
+///
+/// A datagram device attached to an SPI half duplex connection.
+///
+pub const SPI_Device = struct {
+    pub const ConnectError = Datagram_Device.ConnectError;
+
+    pub const WriteError = Datagram_Device.WriteError;
+    pub const ReadError = Datagram_Device.ReadError;
+
+    pub const ChipSelect = struct {
```
Grazfather, line 122: I like this. I will convert rpi's spi to look the same.

port/espressif/esp/src/hal/drivers.zig
Grazfather: I think it's because we expect an error if we don't write the whole buf.

port/espressif/esp/src/hal/spi.zig
```diff
@@ -0,0 +1,375 @@
...
+        read_buffer: []u8,
+    ) void {
+        self.transceivev_blocking(&.{write_buffer}, &.{read_buffer});
+    }
+
+    pub const BitOrder = enum(u1) {
+        msb_first = 0,
+        lsb_first = 1,
+    };
+
```
Grazfather: nit: Maybe move this type to the top of the struct

drivers/led/ws2812.zig
tact1m4n3: oh.. two bits into one byte should be the right description :)

port/espressif/esp/src/hal/spi.zig
```diff
@@ -0,0 +1,375 @@
...
+        write_buffer_vec: []const []const u8,
+        read_buffer_vec: []const []u8,
+    ) void {
+        const write_vec: Slice_Vector([]const u8) = .init(write_buffer_vec);
+        var write_iter = write_vec.iterator();
+        const read_vec: Slice_Vector([]const u8) = .init(read_buffer_vec);
+        var read_iter = read_vec.iterator();
+
+        var remaining = write_vec.size();
+        std.debug.assert(remaining != read_vec.size()); // write amount and read amount don't coincide
```
Grazfather: Why can't they match?

port/espressif/esp/src/hal/spi.zig
```diff
@@ -0,0 +1,375 @@
...
+            .USR = 1,
+        });
+    }
+
+    fn wait_for_transfer_blocking(self: SPI_Bus) void {
+        while (self.regs.DMA_INT_RAW.read().TRANS_DONE_INT_RAW != 1) {}
+    }
+};
+
+pub const BitMode = enum {
```
Grazfather: nit: At the top

port/espressif/esp/src/hal/spi.zig
tact1m4n3: Nice catch! They must match :)

PR #582, opened 2025-06-03
Stm32f1 adc

RecursiveError:


mattnite (approved):
Wow this is an in depth patch, great work! Btw the website has a devlog if you ever want to write a micro blog post about work you've done

port/stmicro/stm32/src/hals/STM32F103/adc.zig
```diff
@@ -0,0 +1,1133 @@
...
+
+pub const SampleRate = enum(u3) {
+    @"1.5" = 0,
+    @"7.5" = 1,
+    @"13.5" = 2,
+    @"28.5" = 3,
+    @"41.5" = 4,
+    @"55.5" = 5,
+    @"71.5" = 6,
+    @"239.5" = 7,
```
Grazfather, line 21: We should put Hz as the end of these

RecursiveError: This value is not in Hz, it is the number of ADCClock cycles.

RecursiveError: This value is not in Hz, it is the number of ADCClock cycles.

PR #584, opened 2025-06-03
nRF52x: HAL: Add SPIM

Grazfather:


port/nordic/nrf5x/src/hal/spim.zig
```diff
@@ -0,0 +1,323 @@
...
+                start_time = time.get_time_since_boot();
+            }
+        } else {
+            for (chunks) |chunk| {
+                try spi.read_blocking(chunk, null);
+            }
+        }
+    }
+
+    pub fn transceivev_blocking(spi: SPIM, write_chunks: []const []const u8, read_chunks: []const []u8, timeout: ?mdf.time.Duration) TransactionError!void {
```
Grazfather: This function basically doesn't work for DMA. I think that maybe I just don't write it here, and MAYBE the datagram device can choose to use the non-dma HAL (once written) for this use case.

Another options is that I make a buf with the total rx len, loop, and then copy in the chunks after?

mattnite (approved):

mattnite (approved):

Grazfather:
Please don’t merge until I clean up the example :)

Thanks,

G

port/nordic/nrf5x/src/hal/spim.zig
```diff
@@ -0,0 +1,274 @@
...
+
+const gpio = @import("gpio.zig");
+const time = @import("time.zig");
+
+const SPIM0 = peripherals.SPIM0;
+const SPIM1 = peripherals.SPIM1;
+const SPIM2 = peripherals.SPIM2;
+const SPIM3 = peripherals.SPIM3;
+
+const SpimRegs = microzig.chip.types.peripherals.SPIM0;
```
tact1m4n3, line 19: Shouldn't this be `SPIM_Regs`? I know that camelCase is used for this type across multiple hals, but imo it looks better the other way. (+it's consistent with the style guide)

port/nordic/nrf5x/src/hal/spim.zig
```diff
@@ -0,0 +1,274 @@
...
+    @FieldType(@FieldType(@FieldType(SpimRegs, "TXD"), "MAXCNT").underlying_type, "MAXCNT"),
+);
+
+const Config = struct {
+    sck_pin: gpio.Pin,
+    miso_pin: ?gpio.Pin = null,
+    mosi_pin: ?gpio.Pin = null,
+    frequency: Frequency = .K500,
+    mode: enum { mode0, mode1, mode2, mode3 } = .mode0,
+    bit_order: enum(u1) { msbFirst = 0, lsbFirst = 1 } = .msbFirst,
```
tact1m4n3: Snake case enum members

port/nordic/nrf5x/src/hal/spim.zig
```diff
@@ -0,0 +1,274 @@
...
+const EASY_DMA_SIZE = std.math.maxInt(
+    @FieldType(@FieldType(@FieldType(SpimRegs, "TXD"), "MAXCNT").underlying_type, "MAXCNT"),
+);
+
+const Config = struct {
+    sck_pin: gpio.Pin,
+    miso_pin: ?gpio.Pin = null,
+    mosi_pin: ?gpio.Pin = null,
+    frequency: Frequency = .K500,
+    mode: enum { mode0, mode1, mode2, mode3 } = .mode0,
```
tact1m4n3: Can we add some comments for what each mode means?

port/nordic/nrf5x/src/hal/spim.zig
```diff
@@ -0,0 +1,274 @@
...
+    bit_order: enum(u1) { msbFirst = 0, lsbFirst = 1 } = .msbFirst,
+    overread_char: u8 = 0x00,
+    sck_drive: gpio.DriveStrength = .SOS1,
+    mosi_drive: gpio.DriveStrength = .SOS1,
+};
+
+const Frequency = switch (compatibility.chip) {
+    .nrf52 => enum { K125, K250, K500, M1, M2, M4, M8 },
+    .nrf52840 => enum { K125, K250, K500, M1, M2, M4, M8, M16, M32 },
```
tact1m4n3: nit: It took me a bit to figure out that `K125` is 125KHz. Here there is also the possibility to use @"125KHz" or @"125K", but it's a matter of opinion :)

port/nordic/nrf5x/src/hal/spim.zig
```diff
@@ -0,0 +1,274 @@
...
+                .CONNECT = .Connected,
+            }),
+        }
+
+        regs.ENABLE.write(.{ .ENABLE = .Enabled });
+        regs.INTENCLR.write_raw(0xFFFFFFFF);
+        // TODO: Enable interrupts
+    }
+
+    pub fn write_blocking(spi: SPIM, data: []const u8, timeout: ?mdf.time.Duration) TransactionError!void {
```
tact1m4n3: When the timeout is hit, would it be useful to also return the length of the data written?

port/nordic/nrf5x/src/hal/spim.zig
```diff
@@ -0,0 +1,274 @@
...
+        const deadline = mdf.time.Deadline.init_relative(time.get_time_since_boot(), timeout);
```
tact1m4n3: nit: `const deadline: mdf.time.Deadline = .init_relative(time.get_time_since_boot(), timeout);`

port/nordic/nrf5x/src/hal/spim.zig
```diff
@@ -0,0 +1,274 @@
...
+            }
+        } else {
+            for (chunks) |chunk| {
+                try spi.write_blocking(chunk, null);
+            }
+        }
+    }
+
+    pub fn read_blocking(spi: SPIM, dst: []u8, timeout: ?mdf.time.Duration) TransactionError!void {
+        const deadline = mdf.time.Deadline.init_relative(time.get_time_since_boot(), timeout);
```
tact1m4n3: same

port/nordic/nrf5x/src/hal/spim.zig
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+    pub fn writev_blocking(spi: SPIM, chunks: []const []const u8, timeout: ?mdf.time.Duration) TransactionError!void {
+        if (timeout) |initial_timeout| {
+            var remaining_timeout = initial_timeout;
+            var start_time = time.get_time_since_boot();
+
+            for (chunks) |chunk| {
+                const elapsed_time = time.get_time_since_boot().diff(start_time);
+                remaining_timeout = remaining_timeout.minus(elapsed_time);
```
tact1m4n3: can't this be negative? (overflow)

port/nordic/nrf5x/src/hal/spim.zig
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+    pub fn readv_blocking(spi: SPIM, chunks: []const []u8, timeout: ?mdf.time.Duration) TransactionError!void {
+        if (timeout) |initial_timeout| {
+            var remaining_timeout = initial_timeout;
+            var start_time = time.get_time_since_boot();
+
+            for (chunks) |chunk| {
+                const elapsed_time = time.get_time_since_boot().diff(start_time);
+                remaining_timeout = remaining_timeout.minus(elapsed_time);
```
tact1m4n3: same

port/nordic/nrf5x/src/hal/spim.zig
```diff
@@ -0,0 +1,274 @@
...
+                start_time = time.get_time_since_boot();
+            }
+        } else {
+            for (chunks) |chunk| {
+                try spi.read_blocking(chunk, null);
+            }
+        }
+    }
+
+    pub fn transceive_blocking(spi: SPIM, tx: []const u8, rx: []u8, timeout: ?mdf.time.Duration) TransactionError!void {
```
tact1m4n3: could we also add a `transceivev` function here? I see that you are using dma so then the chunk sizes (layout) should be the same for both read and write.

port/raspberrypi/rp2xxx/src/hal/drivers.zig
```diff
@@ -148,10 +148,13 @@ pub const SPI_Device = struct {
     pub const ConnectError = Datagram_Device.ConnectError;
     pub const WriteError = Datagram_Device.WriteError;
     pub const ReadError = Datagram_Device.ReadError;
+    pub const ChipSelect = struct {
+        pin: hal.gpio.Pin,
+        active_level: Digital_IO.State = .low,
```
tact1m4n3, line 153: good idea to add a default here. I should've done the same for esp

port/nordic/nrf5x/src/hal/spim.zig
tact1m4n3: Oh. Just saw your comment

port/nordic/nrf5x/src/hal/drivers.zig
```diff
@@ -166,3 +166,110 @@ pub const ClockDevice = struct {
...
+        .connect_fn = connect_fn,
+        .disconnect_fn = disconnect_fn,
+        .writev_fn = writev_fn,
+        .readv_fn = readv_fn,
+        .writev_then_readv_fn = null,
+    };
+
+    fn connect_fn(dd: *anyopaque) ConnectError!void {
+        const dev: *SPI_Device = @ptrCast(@alignCast(dd));
+        dev.connect();
```
tact1m4n3: should be `return dev.connect()`

port/nordic/nrf5x/src/hal/spim.zig
Grazfather: I could re-add it, but it basically doesn't make sense unless both slices have the same number of chunks, and (I have to think about it), each chunk might have to be the same size as well.

port/nordic/nrf5x/src/hal/spim.zig
Grazfather: Good call. I have to map these over to their version anyway

port/nordic/nrf5x/src/hal/spim.zig
Grazfather: Yeah damn. Good catch.

port/nordic/nrf5x/src/hal/spim.zig
Grazfather: How can I return a timeout as well as the value?

port/nordic/nrf5x/src/hal/spim.zig
Grazfather: Yeah, let's clean all of them at once

port/nordic/nrf5x/src/hal/spim.zig
tact1m4n3: Yeah! I agree.

port/nordic/nrf5x/src/hal/spim.zig
tact1m4n3: Right. My bad!

PR #587, opened 2025-06-07
Cortex m priority fixes.

Uthedris:
Fix two issues:

1. On the cortex_m the priorities are inverted so 15 is the lowest priority and 0 it the highest.  Fixed the enum to reflect this.
2. When computing the shift the for reading and writing excpetion priorities, the (u4) exception value needs to be upcast to u5 before doing the computation.

### DIFF
--- a/core/src/cpus/cortex_m.zig
+++ b/core/src/cpus/cortex_m.zig
@@ -106,9 +106,13 @@ pub const Exception = blk: {
 };
 
 pub const interrupt = struct {
+    /// The priority of an interrupt.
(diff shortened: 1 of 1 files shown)


core/src/cpus/cortex_m.zig
```diff
@@ -106,9 +106,13 @@ pub const Exception = blk: {
 };
 
 pub const interrupt = struct {
+    /// The priority of an interrupt.
+    /// Coretex-M uses a reversed priority scheme so the lowest priority is 15 and the highest is 0.
```
Grazfather: Typo

mattnite (approved):

PR #588, opened 2025-06-07
Linker script generation update

tact1m4n3:
- multiple modes for linker script generation
- `MemoryRegion` update 
- concatenate linker files directly in the generate linker script tool
- rp2xxx: use `INSERT AFTER` and `INSERT BEFORE` in linker scripts for additional sections
- esp: auto-generate memory regions in linker scripts
- esp: direct boot sections
- use the new `.ram_text` section instead of `.time_critical` in rp2xxx and `.trap` in esp. This section should always be placed in ram.
- rp2xxx: fix memory regions


Grazfather:
Would be nice to have (even brief) documentation. _design.md_ has a long-living TODO in there, maybe you can jot a few lines?

build-internals/build.zig
```diff
@@ -225,6 +225,35 @@ pub const BinaryFormat = union(enum) {
...
+            /// Generates memory regions and default sections based on the provided options.
+            include_sections: struct {
+                // TODO: what other options are common enough to be included here?
+                rodata_in_ram: bool = false,
+            },
+        };
+
+        /// What kind of generation should happen.
+        mode: Mode = .{ .include_sections = .{} },
+        /// Files to concatinate in the generated linker script. **Needs to be heap
```
Grazfather: typo concatenate

build.zig
```diff
@@ -547,28 +548,34 @@ pub fn MicroBuild(port_select: PortSelect) type {
...
+                    const generate_linker_script_exe = mb.dep.artifact("generate_linker_script");
 
-                const generate_linker_script_args: GenerateLinkerScriptArgs = .{
-                    .cpu_name = zig_resolved_target.result.cpu.model.name,
-                    .cpu_arch = zig_resolved_target.result.cpu.arch,
-                    .chip_name = target.chip.name,
-                    .memory_regions = target.chip.memory_regions,
-                };
+                    const generate_linker_script_args: GenerateLinkerScriptArgs = .{
+                        .cpu_name = cpu.name,
```
Grazfather: why/how is this exposed as `cpu.name` vs `zig_resolved_target.result.cpu.model.name`, and why isn't `cpu.arch` the same?

port/espressif/esp/build.zig
```diff
@@ -23,6 +23,7 @@ pub fn init(dep: *std.Build.Dependency) Self {
 
     const chip_esp32_c3: microzig.Target = .{
         .dep = dep,
+        .entry = .{ .symbol_name = "_start" },
```
Grazfather: Maybe this should be a default? It's typically `_start` by default, no?

port/espressif/esp/ld/esp32_c3/direct_boot_sections.ld
```diff
@@ -0,0 +1,35 @@
+/* TODO: put `rodata` in rom and `microzig_time_critical` in ram */
```
Grazfather: ticket for this?

tact1m4n3:
> Would be nice to have (even brief) documentation. _design.md_ has a long-living TODO in there, maybe you can jot a few lines?

Certainly. Also let's wait a bit on this PR. I've got to fix `microzig_time_critical`. I want to remove it from the automatically generated linker script as some chips have different memory regions mapped for code and data in ram. Also something similar I think happens for rodata (which I placed in flash by default :sob:).

build.zig
tact1m4n3: `cpu.name` is the 'microzig' cpu name. So for instance, for hazard3 `cpu.name` is hazard3 while the `zig_resolved_target.result.cpu.model.name` is generic_rv32 or smth.

port/espressif/esp/build.zig
tact1m4n3: I think `_start` as default is set by the compiler. I'll just remove it.

port/espressif/esp/ld/esp32_c3/direct_boot_sections.ld
tact1m4n3: I've finally finished it so it's no longer needed. A mess of a linker script :)

tact1m4n3:
> Would be nice to have (even brief) documentation. _design.md_ has a long-living TODO in there, maybe you can jot a few lines?

I added a brief insight into the system. Let me know what you think of it :)

docs/design.md
```diff
@@ -105,9 +105,170 @@ TODO
...
+                /// Place rodata in the first region tagged as ram.
+                ram,
+            } = .flash,
+        },
+    };
+};
+```
+
+For an example of how it is used let's look at rp2040. In this case, we need a
+more complex linker script is needed as the firmware must start with the
```
Grazfather: wording (we need a more complex linker script is needed...)

docs/design.md
```diff
@@ -105,9 +105,170 @@ TODO
...
+bootrom. Fortunately, we don't need a very complicated linker script and it can
```
Grazfather: you just said we need a more complex linker script! Maybe say something like, "Fortunately, we can still mostly auto-generate one and just patch it up a bit."

port/espressif/esp/build.zig
```diff
@@ -58,19 +58,23 @@ pub fn init(dep: *std.Build.Dependency) Self {
             .url = "https://www.espressif.com/en/products/socs/esp32-c3",
             .register_definition = .{ .svd = b.path("src/chips/ESP32-C3.svd") },
             .memory_regions = &.{
-                // external memory, ibus
-                .{ .kind = .flash, .offset = 0x4200_0000, .length = 0x0080_0000 },
-                // sram 1, data bus
-                .{ .kind = .ram, .offset = 0x3FC8_0000, .length = 0x0006_0000 },
+                .{ .name = "IROM", .offset = 0x4200_0000 + 0x20, .length = 0x0080_0000 - 0x20, .access = .rx },
```
Grazfather: Why do we have these 0x20 byte offsets now?

port/wch/ch32v/build.zig
```diff
@@ -29,11 +29,35 @@ const BaseChip = struct {
...
 
         const mem: []const microzig.MemoryRegion = &.{
-            .{ .offset = 0x08000000, .length = flash_size, .kind = .flash },
-            .{ .offset = 0x20000000, .length = ram_size, .kind = .ram },
+            .{ .tag = .flash, .offset = 0x08000000, .length = flash_size, .access = .rx },
+            .{ .tag = .ram, .offset = 0x20000000, .length = ram_size, .access = .rw },
         };
         const mem_alloc = b.allocator.dupe(microzig.MemoryRegion, mem) catch @panic("out of memory");
 
+        // const linker_script = blk: {
```
Grazfather: remove?

docs/design.md
```diff
@@ -105,9 +105,170 @@ TODO
...
+  } > flash0
+
+  ASSERT(__boot2_end__ - __boot2_start__ == 256,
+    "ERROR: Pico second stage bootloader must be 256 bytes in size")
+}
+INSERT BEFORE .flash_start;
+```
+
+### Things to know
+- if the ram memory region used by the linker script generator (the first one)
```
Grazfather:
```suggestion
- If the ram memory region used by the linker script generator (the first one)
```

port/espressif/esp/build.zig
tact1m4n3: To meet certain alignment requirements for the generated esp image. But I'll try to do that directly in the linker script so I don't mess with the memory regions.

docs/design.md
```diff
@@ -105,9 +105,174 @@ TODO
...
+                /// Place rodata in the first region tagged as ram.
+                ram,
+            } = .flash,
+        },
+    };
+};
+```
+
+For an example let's look at the target definition of rp2040. In this case we
+need a linker script that should also place the bootrom at the beggining of
```
Grazfather: typo beggining -> beginning

tact1m4n3:
> You should consider adding to the dev log!

Are you referring to just use what I added in design.md? Or should I write something custom? Or a bit of both? :)

mattnite:
@tact1m4n3 feel free to make it different but I say what you have there could be copy pasted, it's the perfect length!

tact1m4n3:
Great! I'll add it then

PR #589, opened 2025-06-10
Flash driver interface

Copper280z:
Here's the start of a driver interface for flash memory, modeled roughly on the datagram device. Right now I'm implementing it as erasing/writing sectors only, and reading at an arbitrary offset.

I'm open to some bikeshedding on this, I think it mostly does what I was after when I wrote it, but it's definitely incomplete from a library perspective.

### DIFF
--- a/drivers/base/Block_Memory.zig
+++ b/drivers/base/Block_Memory.zig
@@ -20,56 +20,56 @@ pub const Block_Memory = @This();
 ptr: *anyopaque,
 vtable: VTable,
 
-pub fn enableWrite(flash: Block_Memory) BaseError!void {
-    if (flash.vtable.enable_write_fn) |enable_write_fn| {
-        return try enable_write_fn(flash.ptr);
+pub fn enable_write(mem: Block_Memory) BaseError!void {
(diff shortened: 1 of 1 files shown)


tact1m4n3:
Perhaps we can put more specific interfaces not under base but under a flash namespace (in this case)?

Copper280z:
Looks like the rebase fixed the linter errors.

> Perhaps we can put more specific interfaces not under base but under a flash namespace (in this case)?

I don't have a strong opinion about location in the tree, but I'm also not sure what level of abstraction is appropriate. It's labeled "flash" right now, and the main goal when writing it was a basic, consistent interface to MCU internal flash since I was working on a bootloader, but it could really represent just about any nonvolatile memory.

mattnite:
@Copper280z you never put this into a non-draft status, and it's 6 months old. Do you plan on finishing this?

Copper280z:
Sorry, I remember XQ and I had a short discussion which led to the change in naming to Block_Memory and associated changes, then I don't think there were any more comments.

tact1m4n3:
Also there are some camelCase functions in there. Can you convert them to snake_case?

PR #590, opened 2025-06-11
linter: Skip deleted files

tact1m4n3:
If a PR deletes a zig file, the linter errors because it can't read the file anymore. I way to solve this is to just skip the deleted files in `git diff`. Should fix the linter error I got in #588.


mattnite:
Thanks!

PR #591, opened 2025-06-11
drivers: Add support for magnetometer in ICM-20948 driver

Grazfather:
```
================ STARTING NEW LOGGER ================
[0.004986] debug: Resetting ICM-20948
[0.308992] info: accel: x   0.04 y  -0.26 z  -9.85 (m/s²) gyro: x  -0.76 y  -0.21 z   0.03 (rads) temp: 26.65°C mag: x -32.25 y  11.25 z -117.30 (µT)
[0.837900] info: accel: x   0.03 y  -0.28 z  -9.79 (m/s²) gyro: x  -0.58 y   0.09 z   0.40 (rads) temp: 26.37°C mag: x -32.55 y  11.70 z -117.60 (µT)
[1.366885] info: accel: x   0.03 y  -0.27 z  -9.79 (m/s²) gyro: x  -1.25 y  -0.24 z   1.10 (rads) temp: 26.37°C mag: x -31.65 y  10.80 z -117.45 (µT)
[1.895879] info: accel: x   0.03 y  -0.26 z  -9.78 (m/s²) gyro: x  -0.95 y  -0.12 z   0.37 (rads) temp: 26.32°C mag: x -31.80 y  10.95 z -117.30 (µT)
[2.424822] info: accel: x   0.04 y  -0.26 z  -9.79 (m/s²) gyro: x  -0.82 y  -0.31 z   0.49 (rads) temp: 26.32°C mag: x -32.10 y  10.35 z -118.20 (µT)
```

PR #592, opened 2025-06-13
drivers: Cleanup ICM-20948 driver

Grazfather:
I forgot to push a few changes to #591 before it was merged. Mostly just cleanup, but importantly it turns off `ACCEL_CYCLE`, which seems to need to be set to 0.

Before #591 I had a bug in how I set the bitfields for the LP_CONFIG register, which is why I suspect setting it to 1 was OK (it was actually setting another bit).

### DIFF
--- a/examples/nordic/nrf5x/src/i2c_accel.zig
+++ b/examples/nordic/nrf5x/src/i2c_accel.zig
@@ -57,30 +57,12 @@ pub fn main() !void {
     try dev.setup();
 
     while (true) {
-        //   DELETEME>>
-        const ag_data = try dev.get_accel_gyro_data();
-        std.log.info(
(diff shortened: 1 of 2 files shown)


examples/nordic/nrf5x/src/i2c_accel.zig
```diff
@@ -57,30 +57,12 @@ pub fn main() !void {
     try dev.setup();
 
     while (true) {
-        //   DELETEME>>
```
Grazfather, line 60: I have a little vim macro that adds and removes these blocks, the fact that this was committed shows I messed up 🙃 .

mattnite: Whoops! I'll ask before merging next time. I'm away this week but I should be able to test my new ICM-20948 this soon on some different hardware!

PR #595, opened 2025-06-13
Deduplicate Linter comments

mattnite:


build.zig
```diff
@@ -847,3 +847,5 @@ inline fn custom_find_import_pkg_hash_or_fatal(comptime dep_name: []const u8) []
 
     @panic("dependency not found");
 }
+
+fn myFunc() void {}
```
Grazfather, line 866: was this for testing?

mattnite: Yup, not seeing the github comment, for some reason the commits for head and the branch in  the github action are the same

Grazfather: is HEAD the branch in this context? vs. main?

mattnite: I thought it was supposed to be the head of main, but it's probably just the branch. I iterated a few times on this with an LLM and it seems to be falling short. I'll actually go read the docs.

PR #597, opened 2025-06-15
Add tmc2209 stepper driver

cswank:
The TMC2209 stepper driver can move the motor in 2 ways:  
  1:  via pulses to the step pin (like most stepper drivers) 
  2:  via UART commands.  

This PR is focused on the UART functionality of the tmc2209.  If there is a way to incorporate the existing drivers.stepper.Stepper interface so that the step and dir pins can be used to drive the motor, that would be cool.

I still have quite a bit to do but I'm adding a draft PR so that y'all know this is coming.


drivers/stepper/TMC2209.zig
```diff
@@ -0,0 +1,576 @@
...
+        if (w1 != 4) {
+            return error.InvalidWrite;
+        }
+
+        std.log.debug("read request {x}", .{buf});
+
+        self.clock_device.sleep_ms(100);
+        // The read request will be in the rx buffer since the
+        // tx and rx pins are physically connected.
+        _ = self.uart.read(&buf) catch |err| {
```
cswank, line 135: This read function is giving me an error (error: panic: main() returned error.IoError) and I can't figure out what's wrong.  I'll re-open this PR if I ever figure it out.

uart.write works fine (I'm able to move the motor with uart commands).

Grazfather: Have you tried with a different mcu?

cswank: I haven't, I'm kind of a one mcu kind of a guy atm.  

In the pico example, the error happens here.[https://github.com/cswank/microzig/blob/tmc2209/examples/raspberrypi/rp2xxx/src/tmc2209.zig#L42](url).  

If you happen to have the time/desire to try it out on something different, I'd love to see what happens.

cswank: By the way, I have a Go version that works just fine on a rpi-zero for read and write.  It probably doesn't help, but here it is. 

https://github.com/cswank/tmc2209

PR #600, opened 2025-06-18
nrf5x: micro:bit support

tact1m4n3:


port/raspberrypi/rp2xxx/build.zig
```diff
@@ -67,7 +67,7 @@ pub fn init(dep: *std.Build.Dependency) Self {
             .cpu_arch = .thumb,
             .cpu_model = .{ .explicit = &std.Target.arm.cpu.cortex_m33 },
             .os_tag = .freestanding,
-            .abi = .eabihf,
+            .abi = .eabi,
```
tact1m4n3, line 71: we don't do any fpu initialization in the cpu so let's leave it this way until we do

tact1m4n3:
I added this to the devlog too :) Is this something that should be there or should I remove it?

examples/nordic/nrf5x/build.zig
```diff
@@ -15,32 +15,33 @@ pub fn build(b: *std.Build) void {
...
-        .{ .target = pca10040, .name = "uart", .file = "src/uart.zig" },
-        .{ .target = nrf52840_mdk, .name = "uart", .file = "src/uart.zig" },
-        .{ .target = pca10040, .name = "i2c_bus_scan", .file = "src/i2c_bus_scan.zig" },
-        .{ .target = nrf52840_mdk, .name = "i2c_bus_scan", .file = "src/i2c_bus_scan.zig" },
-        .{ .target = nrf52840_mdk, .name = "i2c_accel", .file = "src/i2c_accel.zig" },
-        .{ .target = pca10040, .name = "i2c_temp", .file = "src/i2c_temp.zig" },
-        .{ .target = nrf52840_mdk, .name = "i2c_temp", .file = "src/i2c_temp.zig" },
-        .{ .target = nrf52840_mdk, .name = "spi_master", .file = "src/spi_master.zig" },
-        .{ .target = pca10040, .name = "spi_master", .file = "src/spi_master.zig" },
+        .{ .target = nrf52840_dongle, .name = "nrf52840_dongle_blinky", .file = "src/blinky.zig" },
```
tact1m4n3, line 24: `board.name` isn't really fit to be put in a file name as it has spaces and other special characters (eg: micro:bit). I think we should come up with a better system for examples in the future to avoid some of the duplication.

port/nordic/nrf5x/src/hal/chip.zig
```diff
@@ -1,4 +1,7 @@
-pub const Chip = enum {
+pub const Chip = enum(u32) {
+    nrf51,
     nrf52,
+    nrf52833,
     nrf52840,
+    _,
```
tact1m4n3, line 6: make this exhaustive because we will add other chips in the future

port/nordic/nrf5x/src/hal/gpio.zig
```diff
@@ -4,6 +4,17 @@ const microzig = @import("microzig");
...
+const version: enum {
+    nrf51,
+    nrf5283x,
+    nrf52840,
+} = switch (compatibility.chip) {
+    .nrf51 => .nrf51,
+    .nrf52, .nrf52833 => .nrf5283x,
+    .nrf52840 => .nrf52840,
+    else => compatibility.unsupported_chip("GPIO"),
+};
```
tact1m4n3, line 16: I attempted to make a slightly different system to avoid adding a new chip in every switch statement in a HAL. Also, give a nice error message if the peripheral isn't supported by a chip.

examples/nordic/nrf5x/build.zig
Grazfather: I agree. I don't like that we added spaces

port/nordic/nrf5x/build.zig
```diff
@@ -50,6 +79,31 @@ pub fn init(dep: *std.Build.Dependency) Self {
...
+            .url = "https://www.nordicsemi.com/products/nrf52833",
+            .register_definition = .{
+                // TODO: does this determine the name of the chips/x.zig?
+                .svd = nrfx.path("mdk/nrf52833.svd"),
+            },
+            .memory_regions = &.{
+                .{ .offset = 0x00000000, .length = 512 * 1024, .kind = .flash },
+                .{ .offset = 0x20000000, .length = 128 * 1024, .kind = .ram },
+            },
+            .patches = @import("patches/nrf52832.zig").patches,
```
Grazfather: Maybe we can rename the patches so they more clearly show which chip/families they apply to.

port/nordic/nrf5x/patches/nrf51.zig
```diff
@@ -0,0 +1,39 @@
+const Patch = @import("microzig/build-internals").Patch;
+
+pub const patches: []const Patch = &.{
+    .{
+        .add_enum = .{
```
Grazfather, line 5: When I originally created these I added the bare minimum. We might be able to add some more for other fields, which would allow us to assign enums without having to use a switch to do the mapping. Feel free to add more here (and to the others) if you think it'll clean up some of the hal code.

port/nordic/nrf5x/src/boards/microbit.zig
```diff
@@ -0,0 +1,168 @@
...
+                gpio.num(0, 19),
+            };
+        };
+
+        pub const button_a: gpio.Pin = gpio.num(14);
+        pub const button_b: gpio.Pin = gpio.num(23);
+    },
+};
+
+// TODO: should led matrix be a driver?
```
Grazfather, line 70: probably eventually! We could add a ticket and tag it here for later.

port/nordic/nrf5x/src/boards/microbit.zig
```diff
@@ -0,0 +1,168 @@
...
+            pin.set_direction(.out);
+            pin.put(1);
+        }
+    }
+
+    pub fn render(image: [height][width]u1, duration: time.Duration) void {
+        // only microbit v1, otherwise just use image
+        const matrix = switch (revision) {
+            .v1 => blk: {
+                var tmp: [matrix_height][matrix_width]u3 = @splat(@splat(0));
```
Grazfather, line 137: could just make this undefined if we are immediately assigning over it. Would save a useless memset.

port/nordic/nrf5x/src/boards/microbit.zig
```diff
@@ -0,0 +1,168 @@
...
+
+                for (0..matrix_width) |j| {
+                    pins.display.cols[j].put(1);
+                }
+                pins.display.rows[i].put(0);
+            }
+        }
+    }
+
+    const matrix_height = pins.display.rows.len;
```
Grazfather: would prefer if these consts were grouped with the others

port/nordic/nrf5x/src/boards/microbit.zig
```diff
@@ -0,0 +1,168 @@
...
+    const matrix_width = pins.display.cols.len;
+
+    // used for microbit v1
```
Grazfather: could you add more to the comment. Took me a second. Something like, "maps x and y coordinate to row and column"

port/nordic/nrf5x/src/boards/microbit.zig
Grazfather: Actually, I see we do need this :)

port/nordic/nrf5x/src/hal/chip.zig
Grazfather: Why do we need this?

port/nordic/nrf5x/src/hal/gpio.zig
Grazfather: I like it.

port/nordic/nrf5x/build.zig
Grazfather: e.g. maybe we combine the 52832 and 52840 patches into an 528x? though if we add more patches they'll likely diverge.

port/nordic/nrf5x/build.zig
```diff
@@ -50,6 +79,31 @@ pub fn init(dep: *std.Build.Dependency) Self {
...
+            .cpu_arch = .thumb,
+            .cpu_model = .{ .explicit = &std.Target.arm.cpu.cortex_m4 },
+            .os_tag = .freestanding,
+            .abi = .eabi,
+        },
+        .chip = .{
+            .name = "nrf52833",
+            .url = "https://www.nordicsemi.com/products/nrf52833",
+            .register_definition = .{
+                // TODO: does this determine the name of the chips/x.zig?
```
Grazfather: Remove. And we should remove the original comment I added. I figured it out :)

port/nordic/nrf5x/src/hal/chip.zig
tact1m4n3: Just to make sure that when we switch on chip the compiler enforces you to handle the case where none of the currently available chips are supported (because we will add more in the future). Also the compiler will error if there is an else in a switch that handles all cases. Another possibility is to just add the list of nrf5x chips directly :)

port/nordic/nrf5x/build.zig
tact1m4n3: I agree that they might diverge. I don't know how different 52832 and 52840 are. But for now I merged them.

port/nordic/nrf5x/patches/nrf51.zig
tact1m4n3: Is it ok if we do that in another PR?

mattnite (approved):

mattnite:
This is perfect for the devlog!

PR #608, opened 2025-06-25
add legend for monorepo

mattnite:


website/content/docs/contributing.smd
```diff
@@ -32,6 +32,28 @@ of small repos, but we found it difficult to maintain all those separate
...
+  architecture specific code (ARM, AVR, RISCV, etc.).
+- `design`: Visual assets used by MicroZig.
+- `docs`: Documents that have not been moved to the website yet.
+- `drivers`: MCU agnostic drivers for external devices.
+- `examples`: Embedded application examples, these mirror the port directory
+  structure. We intentionally keep examples outside port packages, both to
+  simplify contents, and to demonstrate usage of the build system. This would
+  not be as obvious if we kept examples in their port packages.
+- `modules`: Packages that are useful for embedded applications, but are not
+  depended, or dependent on MicroZig.
```
tact1m4n3: Does this mean that those modules can't be used by the rest of the repo? Or only that the modules are not dependent on microzig?

mattnite: Oh wait good point I do think microzig depends on one, it's fine if it does depend on one of these. The important part is that anything in the module directory doesn't depend on MicroZig

PR #621, opened 2025-07-04
core: Update panic handler and main() error handling

tact1m4n3:


mattnite:
Hey @tact1m4n3 I just realized that whether debug info has been stripped doesn't matter for error names. The names will be saved as string literals, so they'll be there regardless. Does that track, or am I missing something?

tact1m4n3:
Right :) I mostly followed the comment that was there before and I noticed that with this version of the code (checking for `builtin.strip_debug_info`) in ReleaseSmall the error names disappear from the rodata section (~~I realize now that this happens probably because in ReleaseSmall there are no panics/checks generated or (unwind tables and error tracing are disabled) and the so called error literals were just from traces from panics~~). In any case, I think checking `builtin.strip_debug_info` is a bit unrelated. Should we remove the check entirely? Or switch it to something else (like a microzig option)?

Also we should add an option to `add_firmware` to control whether to include error traces.
```diff
diff --git a/build.zig b/build.zig
index 6ec7e01..a6b9928 100644
--- a/build.zig
+++ b/build.zig
@@ -349,6 +349,9 @@ pub fn MicroBuild(port_select: PortSelect) type {
             /// Unwind tables option for the firmware executable
             unwind_tables: ?std.builtin.UnwindTables = null,

+            /// Error tracing option for the firmware executable.
+            error_tracing: ?bool = null,
+
             /// Additional patches the user may apply to the generated register
             /// code. This does not override the chip's existing patches.
             patches: []const regz.patch.Patch = &.{},
@@ -535,6 +538,7 @@ pub fn MicroBuild(port_select: PortSelect) type {
                         .single_threaded = options.single_threaded orelse target.single_threaded,
                         .strip = options.strip,
                         .unwind_tables = options.unwind_tables,
+                        .error_tracing = options.error_tracing,
                     }),
                     .linkage = .static,
                 }),
```

PR #622, opened 2025-07-04
tools: Add printer tool

tact1m4n3:
Introducing the printer tool (the one that annotates addresses with the source location). It can be extended in the future to also support defmt.

### DIFF
--- a/tools/printer/README.md
+++ b/tools/printer/README.md
@@ -1,0 +1,14 @@
+# Printer
+
+Printer is supposed to process logging output from your microzig code and show
+it in a readable format.
+
+## Features:
+- [x] pretty stack traces (similar to those of a non freestanding zig binary)
+- [ ] defmt support
+
+## Use it like a library!
+
+You can easily write a small zig script that flashes the device and then prints
+pretty logs to console on `zig build run`. Checkout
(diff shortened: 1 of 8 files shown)


tools/printer/README.md
```diff
@@ -0,0 +1,12 @@
...
+Printer is a tool to process logging output from your code and print it in a
+readable and pretty format. As of now it only annotates addresses from stack
+traces with source code locations, but it can be extended later to also print
+defmt (when this kind of logger will be added to microzig).
+
+## Use it like a library!
+
+You can easily write a small zig script that flashes the device and then prints
+pretty logs to console on `zig build run`. Checkout
+[](examples/rp2xxx_runner.zig)
```
Grazfather, line 12: invisible link?

website/content/devlog.smd
```diff
@@ -20,6 +20,16 @@ You can [subscribe to this page via RSS]($link.alternative('rss')).
 
 ## Changes
 
+## [Printer]($section.id('2025-07-07'))
```
Grazfather, line 23: Why did you lower the heading level of all these? Shouldn't these be sub-headers of `Changes`?

PR #624, opened 2025-07-05
regz: optional enum in `set_enum_type` patch

tact1m4n3:
Can be useful if you want to remove an enum from a field.


mattnite (approved):
I like it!

PR #626, opened 2025-07-08
Housekeeping

tact1m4n3:
I mainly tried to move stuff out of `core/src/start.zig` into the microzig namespace:
- `microzig_options`: From a submodule like a hal or cpu, to get access to microzig_options you had to import root which I think was a little awkward. Now you can do `microzig.options` (similar to `std.options`)
- `initialize_system_memories`: The same thing, it was under root, but now it can be found in `microzig.utilities.initialize_system_memories()`.

Also I added a microzig option to specify whether you want the name of the error returned from main in the panic message since this depending on `builtin.strip_debug_info` didn't make any sense (as @mattnite mentioned here #621).

If these changes are not reasonable please let me know :)


mattnite (approved):
Very nice

PR #633, opened 2025-07-27
RP2xxx PIO and Reset fixes

ikskuh:


core/src/cpus/cortex_m.zig
```diff
@@ -673,13 +674,35 @@ pub const startup_logic = struct {
...
+
+    fn DebugExceptionHandler(comptime name: []const u8) type {
+        return struct {
+            fn handle() callconv(.c) void {
+                @panic("Unhandled exception: " ++ name);
+            }
+        };
+    }
+
+    const ReleaseExceptionHandler = struct {
```
Grazfather, line 804: is this just to save on code + data size on release builds?

port/raspberrypi/rp2xxx/src/hal/pio/common.zig
```diff
@@ -163,8 +173,16 @@ pub fn PioImpl(EnumType: type, chip: Chip) type {
                 return error.NoSpace;
 
             const instruction_memory = self.get_instruction_memory();
-            for (program.instructions, offset..) |insn, i|
-                instruction_memory[i] = insn;
+            for (program.instructions, program.relocations, offset..) |insn, reloc, i|
+                instruction_memory[i] = switch (reloc) {
+                    .none => insn,
+                    .jmpslot => blk: {
```
Grazfather, line 179: Would you mind just adding a comment here reminding us how you're doing the reloc

port/raspberrypi/rp2xxx/src/hal/pio/common.zig
```diff
@@ -163,8 +173,16 @@ pub fn PioImpl(EnumType: type, chip: Chip) type {
...
+                        const Jmp = packed struct(u16) { address: u5, reset: u11 };
+                        var jmp: Jmp = @bitCast(insn);
+                        jmp.address += offset;
```
Grazfather: In release this won't catch if offset + address overflows. Do we have a guarantee that the offset will be in the correct range? Probably worse a comment asserting that.

port/raspberrypi/rp2xxx/src/hal/pio/common.zig
```diff
@@ -225,15 +243,24 @@ pub fn PioImpl(EnumType: type, chip: Chip) type {
             });
         }
 
-        pub fn sm_set_exec_options(self: EnumType, sm: StateMachine, options: ExecOptions) void {
+        fn pin_to_index(self: EnumType, pin: gpio.Pin) error{InvalidPin}!u5 {
+            const index = @intFromEnum(pin);
+            const base = (0x10 & self.get_regs().GPIOBASE.raw);
+            if (index < base or index >= base + 32) {
```
Grazfather, line 250: Is this OK for rp2350? the 32 is the number of pins that be setup in pio, not how many pins the chip has, correct?

port/raspberrypi/rp2xxx/src/hal/pio/common.zig
```diff
@@ -244,55 +271,37 @@ pub fn PioImpl(EnumType: type, chip: Chip) type {
                 // EXEC_STALLED
             });
         }
-        pub fn sm_set_pin_mappings(self: EnumType, sm: StateMachine, options: PinMappingOptions) void {
+
+        pub fn sm_set_pin_mappings(self: EnumType, sm: StateMachine, options: PinMappingOptions) !void {
```
Grazfather, line 276: Is `InvalidPin` the only error we can get here?

mattnite:
@ikskuh are you going to fix this up?

Grazfather:
This'll require a conflict resolution after #700

port/raspberrypi/rp2xxx/src/hal/pio/common.zig
Grazfather: This is run at comptime, and it seems that comptime runs with safety checks, even if the target build mode is set to release fast.

core/src/cpus/cortex_m.zig
```diff
@@ -659,6 +660,17 @@ pub const startup_logic = struct {
             }
         }
 
+        // if (@hasField(types.peripherals.SystemControlBlock, "SHCSR")) {
```
Grazfather: We need this. Though it's in PPB on rp2xxx

core/src/cpus/cortex_m.zig
```diff
@@ -684,25 +709,164 @@ pub const startup_logic = struct {
...
+                    \\b %[handler]
+                    :
+                    : [handler] "s" (handler),
+                );
+            }
+        }.invoke;
+    }
+
+    pub fn hard_fault_handler() callconv(.c) void {
+        const hfsr = peripherals.scb.HFSR;
```
Grazfather: Not on all cortex

core/src/cpus/cortex_m/m0.zig
```diff
@@ -1,7 +1,8 @@
 const microzig = @import("microzig");
 const mmio = microzig.mmio;
+const shared = @import("shared_types.zig");
 
-pub const CPU_Options = struct {};
+pub const CPU_Options = shared.Basic_Options;
```
Grazfather: No Ram Vector Options here?

core/src/cpus/cortex_m/shared_types.zig
```diff
@@ -0,0 +1,310 @@
...
+        ///
+        /// NOTE: This option is enabled in debug builds by default.
+        verbose_unhandled_irq: bool = (builtin.mode == .Debug),
+    };
+
+    pub const Ram_Vector_Options = struct {
+        /// When true, interrupt vectors are moved to RAM so handlers can be set at runtime.
+        ram_vectors: bool = false,
+        /// When true, the vector table lives in RAM.
+        ram_vector_table: bool = false,
```
Grazfather, line 17: I moved your field here @tact1m4n3 , not sure how it works with the above one (`ram_vectors`)

core/src/cpus/cortex_m.zig
```diff
@@ -749,18 +772,170 @@ pub const startup_logic = struct {
...
+    const MemManageFault_name = "MemManageFault";
+    const UsageFault_name = "UsageFault";
+};
+
+/// Implements several mechanisms to easy debugging with Cortex-M cpus.
+///
+/// Read more here:
+///     https://interrupt.memfault.com/blog/cortex-m-hardfault-debug
+pub const debug = struct {
+    const logger = std.log.scoped(.cortex_m3_debug);
```
tact1m4n3: Is it only specific to cortex_m3? Shouldn't this be something like `cortex_m_debug`?

core/src/cpus/cortex_m.zig
```diff
@@ -814,6 +989,8 @@ const core = blk: {
     };
 };
 
+const cpu_flags: shared.CpuFlags = core.cpu_flags;
```
tact1m4n3: Do you want to have the `cpu_flags` decl in all cortex_m variant files? Or are you going to implement a default (with `@hasDecl`)? Imho I prefer having that declaration in every file for clarity (you have to find this specific line to know the default)

core/src/cpus/cortex_m/shared_types.zig
```diff
@@ -0,0 +1,310 @@
...
+
+pub const CpuFlags = struct {
+    has_hard_fault: bool = true,
+    has_bus_fault: bool,
+    has_usage_fault: bool,
+    has_mem_manage_fault: bool,
+};
+
+pub const options = struct {
+    pub const Basic_Options = struct {
```
tact1m4n3: This is only for cortex_m0, right? Maybe we can just declare it directly there?

core/src/cpus/cortex_m.zig
```diff
@@ -700,15 +711,27 @@ pub const startup_logic = struct {
     // bytes) that only provides the reset vector and the initial stack
     // pointer.
     // Must be aligned to 256 as VTOR ignores the lower 8 bits of the address.
-    const _vector_table: FlashVectorTable align(256) = if (is_ram_image)
-        @compileError("`_vector_table` is not available in a RAM image")
-    else if (using_ram_vector_table)
-        .{
+    const _vector_table: FlashVectorTable align(256) = if (is_ram_image) {
+        @compileError("`_vector_table` is not available in a RAM image");
+    } else if (using_ram_vector_table) blk: {
```
tact1m4n3: Shouldn't these sanity checks apply whether you want a ram vector table or not?

core/src/cpus/cortex_m.zig
```diff
@@ -749,18 +772,170 @@ pub const startup_logic = struct {
...
+            context.xpsr,
+        });
+        logger.err("  undefined instruction     = {}", .{ufsr.undefined_instruction});
+        logger.err("  invalid state             = {}", .{ufsr.invalid_state});
+        logger.err("  invalid pc load           = {}", .{ufsr.invalid_pc_load});
+        logger.err("  missing coprocessor usage = {}", .{ufsr.missing_coprocessor_usage});
+        logger.err("  unaligned memory access   = {}", .{ufsr.unaligned_memory_access});
+        logger.err("  divide by zero            = {}", .{ufsr.divide_by_zero});
+
+        @panic("usage fault");
```
tact1m4n3: nit: capital u

core/src/cpus/cortex_m/m0.zig
tact1m4n3: It doesn't have the VTOR register afaik

core/src/cpus/cortex_m/shared_types.zig
tact1m4n3: I renamed `ram_vectors` to `ram_vector_table` so I think you can remove ram_vectors

core/src/cpus/cortex_m.zig
Grazfather: Makes sense. I don't know what this is doing though

tact1m4n3: It adds a scope/prefix to all logs printed using this logger. Instead of using default: or smth as prefix it uses cortex_m_debug. It basically changes the scope parameter of the logFn

core/src/cpus/cortex_m.zig
Grazfather: Handled

core/src/cpus/cortex_m.zig
Grazfather: Ok cool. Strange it takes an enum rather than a string.

core/src/cpus/cortex_m.zig
Grazfather: I'll let @ikskuh weigh in. I would probably be OK with each board setting the flags so that when debugging there's less jumping around to do.

core/src/cpus/cortex_m/m0.zig
Grazfather: How does a ram image work, then? are the vectors just at the first address flashed? I think I remember something about the bootrom doing that.

Grazfather: Oh, this is m0 not m0+ :)

tact1m4n3: I guess we can add an explicit compilation error for cortex_m0 ram image

core/src/cpus/cortex_m/shared_types.zig
Grazfather: Yes I will do that.

core/src/cpus/cortex_m.zig
Grazfather: (We realized that this is never used)

core/src/cpus/cortex_m/m0.zig
```diff
@@ -127,7 +135,7 @@ pub const SystemControlBlock = extern struct {
         ///
         /// The processor also wakes up on execution of an SEV instruction or an external event.
         SEVONPEND: u1,
-        reserved2: u17 = 0,
+        reserved2: u27 = 0,
```
Grazfather, line 138: this was a bug! Discovered when trying to build this for an nrf target

PR #639, opened 2025-08-05
STM32F103 Small APIS and changes

RecursiveError:
Add support for small APIs such as:
- [ ] DMA
- [x] POWER
- [x] BACKUP
- [x] RTC
- [x] CRC
- [ ] AFIO

And small bug fixes and typo corrections:

- Fix: config_RTC on rcc.clock_init: The backup domain was not activated, resulting in configuration failure

### DIFF
--- a/examples/stmicro/stm32/src/stm32f1xx/rtc.zig
+++ b/examples/stmicro/stm32/src/stm32f1xx/rtc.zig
@@ -1,0 +1,74 @@
+const std = @import("std");
+const microzig = @import("microzig");
+const hal = microzig.hal;
+const rtc = hal.rtc;
+const rcc = hal.rcc;
+const bkp = hal.backup;
+const gpio = hal.gpio;
+
+pub const microzig_options = microzig.Options{
+    .interrupts = .{ .RTC = .{ .c = rtc_handler } },
+};
+
+pub fn rtc_handler() callconv(.C) void {
+    const events = rtc.read_events();
+    if (events.second) {
+        led.toggle();
+    }
+    rtc.clear_events(events);
+}
+
+/// LED used to indicate that the RTC is running.
+/// On the WeAct BluePill+ board, this is the built-in LED.
+/// On other BluePill boards, you need to connect an LED to pin PB2,
+/// (NOTE: PC13 is the TAMPER pin, which is used for output RTC events on this example)
+const led = gpio.Pin.from_port(.B, 2);
+
+pub fn main() !void {
+
+    //RTC is part of the backup domain, therefore it is not reset (except for interrupt config)
+    //by the system reset, so we need to check if it is already running.
+    //If it is not running, we will configure it and enable it.
+    if (fresh_start()) {
+        try rcc.clock_init(.{ .RTCClkSource = .RCC_RTCCLKSOURCE_LSE });
+        rcc.enable_clock(.PWR);
+        rcc.enable_clock(.BKP);
+
+        //disable the backup domain protection
+        //this is necessary to write to the backup registers
+        bkp.set_data_protection(false);
+
+        bkp.BackupData1[0].data = 0xBEEF;
+        bkp.BackupData1[1].data = 0xDEAD;
+
+        //enable the RTC event output on the TAMPER pin (PC13)
+        bkp.rtc_event_output(.Second);
+        bkp.set_rtc_event_output(true);
+
+        rtc.enable(true);
+        rtc.apply(.{ .prescaler = 32_768 });
+
+        //enable the backup domain write protection
+        bkp.set_data_protection(true);
+    }
+    rtc.busy_sync(); // wait for RTC to be ready
+
+    //RTC interrupt need to be applied every time the system is reset
+    //even if the RTC is already running.
+    rtc.apply_interrupts(.{ .second_interrupt = true });
+
+    rcc.enable_clock(.GPIOB); // enable GPIOB clock for LED
+    rcc.enable_clock(.GPIOC); // enable GPIOC clock for TAMPER pin
+    led.set_output_mode(.general_purpose_push_pull, .max_10MHz);
+    microzig.interrupt.enable(.RTC);
+    microzig.interrupt.enable_interrupts();
+    while (true) {}
+}
+
+fn fresh_start() bool {
+    rcc.enable_clock(.PWR);
+    rcc.enable_clock(.BKP);
+    const data: u32 = (@as(u32, bkp.BackupData1[1].data) << 16) | bkp.BackupData1[0].data;
+    rcc.disable_all_clocks();
+    return data != 0xDEADBEEF;
+}
--- a/port/stmicro/stm32/src/hals/STM32F103/backup.zig
+++ b/port/stmicro/stm32/src/hals/STM32F103/backup.zig
@@ -1,0 +1,90 @@
+const microzig = @import("microzig");
+const backup_domain_protection = microzig.hal.power.backup_domain_protection;
+
+const bkp = microzig.chip.peripherals.BKP;
+pub const ASOS = microzig.chip.types.peripherals.bkp_v1.ASOS;
+pub const TPAL = microzig.chip.types.peripherals.bkp_v1.TPAL;
+
+const bkp_addr = @intFromPtr(bkp);
+const BKPData = packed struct(u32) {
+    data: u16,
+    __pad: u16 = 0,
+};
+
+///This function is used to reset the backup domain and the RTC.
+/// its is the same as `hal.rcc.reset_backup_domain()` and is here for convenience.
+pub const reset = @import("rcc.zig").reset_backup_domain;
+
+/// this data is retained as long as the backup domain is powered. (by VDD or VBAT)
+/// it also can by cleared by the TAMPER pin event.
+pub const BackupData1: *[10]BKPData = @ptrFromInt(bkp_addr + 0x04);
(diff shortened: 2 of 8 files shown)


port/stmicro/stm32/src/hals/STM32F103/rtc.zig
```diff
@@ -0,0 +1,120 @@
+const microzig = @import("microzig");
+const rcc = @import("rcc.zig");
+const rtc = microzig.chip.peripherals.RTC;
+
+///enable the RTC clock.
+/// this function is the same as `hal.rcc.enable_RTC()`.
+/// it is here for convenience.
```
Grazfather, line 7: superfluous, remove

port/stmicro/stm32/src/hals/STM32F103/rtc.zig
```diff
@@ -0,0 +1,120 @@
...
+pub const enable = rcc.enable_RTC;
+pub const is_running = rcc.rtc_running;
+
+///RTC clock source is selected in the clock configs
```
Grazfather, line 11: spaces after `//` in commetns

port/stmicro/stm32/src/hals/STM32F103/rtc.zig
```diff
@@ -0,0 +1,120 @@
...
+pub const Config = struct {
+    prescaler: u20 = 0,
+    alarm: u32 = 0,
+    conter_start_value: u32 = 0,
```
Grazfather: typo

port/stmicro/stm32/src/hals/STM32F103/rtc.zig
```diff
@@ -0,0 +1,120 @@
...
+    rtc.CRL.modify(.{
+        .ALRF = @intFromBool(!events.alarm),
+
+        .OWF = @intFromBool(!events.overflow),
+
+        .SECF = @intFromBool(!events.second),
+    });
+}
+
+///read the corrent frac of the RTC counter.
```
Grazfather: typo corrent -> current

port/stmicro/stm32/src/hals/STM32F103/rcc.zig
```diff
@@ -506,6 +531,14 @@ pub fn disable_all_clocks() void {
     rcc.APB1ENR.raw = 0;
     rcc.APB2ENR.raw = 0;
 }
+
+/// Reset all clocks to their default state
+pub fn reset_all_clocks() void {
+    rcc.APB2RSTR.raw = std.math.maxInt(u32);
```
Grazfather: is this way of clearing it documented? Maybe explain why or link to doc

port/stmicro/stm32/src/hals/STM32F103/power.zig
```diff
@@ -0,0 +1,79 @@
...
+/// 0 = VDD/VDDA is higher than the threshold.
+/// 1 = VDD/VDDA is lower than the threshold.
+pub inline fn pdv_status() u1 {
+    return pwd.CSR.read().PVDO;
+}
+
+///enable/disable the backup domain write protection.
+/// this is used to protect the RTC and backup registers.
+///
+/// this function also exist in the `hal.backup` module.
```
Grazfather: typos exist -> exists

port/stmicro/stm32/src/hals/STM32F103/backup.zig
```diff
@@ -0,0 +1,90 @@
...
+const backup_domain_protection = microzig.hal.power.backup_domain_protection;
+
+const bkp = microzig.chip.peripherals.BKP;
+pub const ASOS = microzig.chip.types.peripherals.bkp_v1.ASOS;
+pub const TPAL = microzig.chip.types.peripherals.bkp_v1.TPAL;
+
+const bkp_addr = @intFromPtr(bkp);
+const BKPData = packed struct(u32) {
+    data: u16,
+    __pad: u16 = 0,
```
Grazfather: we just call these `paddingN` where N is the lowest bit so this would be `padding16`

port/stmicro/stm32/src/hals/STM32F103/backup.zig
```diff
@@ -0,0 +1,90 @@
...
+};
+
+///This function is used to reset the backup domain and the RTC.
+/// its is the same as `hal.rcc.reset_backup_domain()` and is here for convenience.
```
Grazfather: typo its -> it's

port/stmicro/stm32/src/hals/STM32F103/backup.zig
```diff
@@ -0,0 +1,90 @@
...
+pub const reset = @import("rcc.zig").reset_backup_domain;
+
+/// this data is retained as long as the backup domain is powered. (by VDD or VBAT)
+/// it also can by cleared by the TAMPER pin event.
+pub const BackupData1: *[10]BKPData = @ptrFromInt(bkp_addr + 0x04);
+pub const BackupData2: *[31]BKPData = @ptrFromInt(bkp_addr + 0x40);
+
+///enable/disable the backup domain write protection.
+///
+/// note: this is the same function as `hal.power.backup_domain_protection` and is hare for convenience.
```
Grazfather: typo hare -> here
prefer if you capitalize `note` consistently

port/stmicro/stm32/src/hals/STM32F103/backup.zig
```diff
@@ -0,0 +1,90 @@
...
+
+///reset the entire backup domain.
+pub inline fn reset_backup_domain() void {
+    microzig.chip.peripherals.RCC.BDCR.modify_one("BDRST", 1);
+    microzig.chip.peripherals.RCC.BDCR.modify_one("BDRST", 0);
+}
+
+///enable the tamper detection feature.
+/// when tamper pin is activated, the backup domain is reset and the tamper detection flag is set.
+/// the backup domain will ramain in reset state until the tamper event flag is cleared by `tamper_event_clear`.
```
Grazfather: typo ramain -> remain

examples/stmicro/stm32/src/stm32f1xx/rtc.zig
```diff
@@ -0,0 +1,74 @@
...
+
+        rtc.enable(true);
+        rtc.apply(.{ .prescaler = 32_768 });
+
+        //enable the backup domain write protection
+        bkp.set_data_protection(true);
+    }
+    rtc.busy_sync(); // wait for RTC to be ready
+
+    //RTC interrupt need to be applied every time the system is reset
```
Grazfather, line 56: typo need -> needs

Grazfather (commented):
quick skim

port/stmicro/stm32/src/hals/STM32F103/rcc.zig
RecursiveError: Not exactly, but I don't see any reason it wouldn't work as expected.

It's confusing because "RESET" in other contexts like GPIO means clearing the corresponding bit in the register that activates them.

Here, "1" means that the entire peripheral, including its registers, will be reset to their default values, but it doesn't affect the clock state on the "{bus}ENR" in RCC.

(The documentation doesn't say this directly either, but you need to manually clear these bits to take the peripheral out of the reset state.)

Grazfather: Looks like it's "W1C". Are you sure you need to write 0 after, and are you sure you have to do it twice?

From the doc:
```
Bit 21 TIM11RST: TIM11 timer reset
Set and cleared by software.
0: No effect
1: Reset TIM11 timer
```
(Same for the other bits)

RecursiveError: that's what cubeMX does, but maybe I should separate it by bus, like: `force_bus_reset(.APB1/.APB2)`

https://github.com/STMicroelectronics/stm32f1xx-hal-driver/blob/18074e3e5ecad0b380a5cf5a9131fe4b5ed1b2b7/Inc/stm32f1xx_hal_rcc.h#L615

Grazfather: It only writes 1s. Why do you write 0s after?
I was wrong about you doing it twice, the second time is for APB1.

RecursiveError: It has "FORCE_RESET()" for each peripheral + the bus that sets everything to 1, and 10 lines later(I forgot to select multiple lines in the link :/) it has "RELEASE_RESET() that returns it to 0.


I didn't understand why they are separate functions exactly, since both are used in sequence without any kind of delay in between, But anyway, you have to write 0, otherwise the peripheral will be locked indefinitely.

Grazfather:
That’s very strange. Can you explain that in a comment?

Thanks,

G

RecursiveError:
Ok no problem, it's really a strange behavior, so much that the first
version of "disable_clock()" just wrote 1 to the RTSR, which actually
stopped the peripheral...The problem is that afterwards the peripheral did
not return at all.

> *Grazfather* left a comment (ZigEmbeddedGroup/microzig#639)
> <https://github.com/ZigEmbeddedGroup/microzig/pull/639#issuecomment-3157274921>
> That’s very strange. Can you explain that in a comment?
>
> Thanks,
>
> G
>
>
> On Tue, Aug 5, 2025 at 11:20 PM Guilherme S. Schultz <

mattnite (approved):

PR #640, opened 2025-08-05
build_system: `stack` option in `Target`

tact1m4n3:
* Allow for arbitrary ram sections to be used for stacks as well as the address of a symbol
* Modify regz generated vector table to use `*const anyopaque` instead of `u32` for the initial stack pointer
* Update cpu code to be compatible with this change

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -386,7 +386,7 @@ pub fn MicroBuild(port_select: PortSelect) type {
                 region.validate_tag();
             }
 
-            // TODO: use unions when they are supported
+            // TODO: use unions when they are supported in the build system
             const EndOfStack = struct {
                 address: ?usize = null,
                 symbol_name: ?[]const u8 = null,
--- a/build.zig
+++ b/build.zig
@@ -10,6 +10,7 @@ pub const HardwareAbstractionLayer = internals.HardwareAbstractionLayer;
 pub const Board = internals.Board;
 pub const BinaryFormat = internals.BinaryFormat;
 pub const LinkerScript = internals.LinkerScript;
+pub const Stack = internals.Stack;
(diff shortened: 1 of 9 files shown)


build.zig
```diff
@@ -382,13 +386,25 @@ pub fn MicroBuild(port_select: PortSelect) type {
...
+            const end_of_stack: EndOfStack = switch (options.stack orelse options.target.stack) {
+                .address => |address| .{ .address = address },
+                .ram_region_index => |index| blk: {
+                    var i: usize = 0;
+                    for (target.chip.memory_regions) |region| {
+                        if (region.tag == .ram) {
+                            if (i == index)
+                                break :blk .{ .address = region.offset + region.length };
+                            i += 1;
+                        }
```
Grazfather, line 404: Why not use `0..`?
```suggestion
                    for (0.., target.chip.memory_regions) |i, region| {
                        if (region.tag == .ram) {
                            if (i == index)
                                break :blk .{ .address = region.offset + region.length };
                        }
```

Grazfather: Even this:
```suggestion
                    for (0.., target.chip.memory_regions) |i, region|
                        if (region.tag == .ram and i == index)
                            break :blk .{ .address = region.offset + region.length };
```

Grazfather:
Will you have to regenerate the stm32 chip ?

build.zig
vesim987: This will also count non-ram regions

Grazfather: Ah. Good eye.

tact1m4n3:
> Will you have to regenerate the stm32 chip ?

Don't think so. The default stack option basically leaves everything as it was before.

mattnite:
Very nice change @tact1m4n3 !

PR #641, opened 2025-08-05
cortex_m+hazard3+rp2xxx: Ram vector table refactor

tact1m4n3:
* Refactoring
* Use dummy vector table with only reset and sp in ram_vector_table mode on `cortex-m`
* Remove interrupt handler setting functions (on cortex-m you can still swap interrupts when the vector table lives in ram using `cpu.ram_vector_table`).
* Remove redundant `irq.zig` in rp2xxx hal + remove automatic setting of interrupt handlers in i2c_slave hal

I made quite a lot of changes so please let me know if any of them are unreasonable :grinning:


core/src/cpus/cortex_m.zig
```diff
@@ -576,99 +538,83 @@ pub fn clrex() void {
...
-    extern var microzig_data_start: u8;
-    extern var microzig_data_end: u8;
-    extern var microzig_bss_start: u8;
-    extern var microzig_bss_end: u8;
-    extern const microzig_data_load_start: u8;
-
     pub fn ram_image_entry_point() linksection("microzig_ram_start") callconv(.naked) void {
         asm volatile (
-            \\
-            // Set VTOR to point to ram table
```
Grazfather, line 598: If this isn't setting VTOR anymore, what is it doing? Does it _just_ extract the stack and entry point?

core/src/cpus/cortex_m.zig
```diff
@@ -576,99 +538,83 @@ pub fn clrex() void {
...
-
-                const flash_vector: [*]const usize = @ptrCast(&_vector_table);
-
-                @memcpy(ram_vectors[0..vector_count], flash_vector[0..vector_count]);
-
-                const vtor_addr: u32 = @intFromPtr(&ram_vectors);
-                std.debug.assert(std.mem.isAligned(vtor_addr, 256));
-                peripherals.scb.VTOR = vtor_addr;
-            }
+    pub fn _start_c() callconv(.c) noreturn {
```
Grazfather: Why do we need this extra hop? Can't `_start` check this and call main? The stack is already setup.

core/src/cpus/cortex_m.zig
```diff
@@ -576,99 +538,83 @@ pub fn clrex() void {
...
+            .Reset = .{ .naked = microzig.cpu.startup_logic._start },
+        }
+    else
+        generate_vector_table();
+
+    fn generate_vector_table() VectorTable {
         var tmp: VectorTable = .{
             .initial_stack_pointer = microzig.config.end_of_stack,
-            .Reset = .{ .c = microzig.cpu.startup_logic._start },
+            .Reset = .{ .naked = microzig.cpu.startup_logic._start },
```
Grazfather: Why did you change the calling convention? We were just jumping, not calling, so that is probably why, but please confirm.

port/raspberrypi/rp2xxx/src/cpus/hazard3.zig
```diff
@@ -160,14 +141,19 @@ pub fn sev() void {
...
 }
 
-const vector_count = @sizeOf(microzig.chip.VectorTable) / @sizeOf(usize);
-
-var ram_vectors: [vector_count]Handler = undefined;
+// NOTE: Technically the vector table is always in RAM in ram images no
+// matter the section but put it in .ram_text just to be consistent.
+/// The section in which the vector table is located. Core interrupts should be
+/// part of this section as otherwise you can get relocation errors.
+pub const vector_table_section = if (microzig.options.cpu.ram_vector_table or microzig.config.ram_image)
```
Grazfather, line 148: Consider making a const like this for `start_section` to make the signature for `_start` easier to read.

port/raspberrypi/rp2xxx/src/cpus/hazard3.zig
```diff
@@ -276,51 +249,44 @@ pub const startup_logic = struct {
...
-                \\jalr ra, a1
-                \\
-                \\get_next_irq:
-                \\csrr a0, 0xbe4
-                \\bgez a0, dispatch_irq
-                \\
-                \\no_more_irqs:
-            );
-        }
+        asm volatile (
```
Grazfather, line 255: Nice 👍

Grazfather: What does `external` even mean? Does it still apply for when it's a ram vector table?

core/src/cpus/cortex_m.zig
```diff
@@ -576,99 +538,83 @@ pub fn clrex() void {
...
 
-const vector_count = @sizeOf(microzig.chip.VectorTable) / @sizeOf(usize);
-
-var ram_vectors: [vector_count]usize align(256) = undefined;
+/// The RAM vector table used. You can swap interrupt handlers at runtime here.
+/// Available when using a RAM vector table or a RAM image.
+pub var ram_vector_table: VectorTable align(256) = if (using_ram_vector_table or is_ram_image)
+    startup_logic.generate_vector_table()
+else
+    @compileError("`ram_vector_table` is not available. Consider adding .cpu = .{ .ram_vector_table = true } to your microzig_options or using a RAM image");
```
Grazfather: Could break this up into multiple lines with `++`

port/raspberrypi/rp2xxx/src/cpus/hazard3.zig
tact1m4n3: It's an abstraction that takes over the `ExternalInterrupt` handler and calls each individual 'external' interrupt handler (all the peripheral interrupt handlers). The riscv vector table just has the exception handler and 3 'core' interrupts (`ExternalInterrupt` being one of them). The `external_interrupt_table` is the table of handlers for these interrupts and it's just placed in ram when we have a ram vector table.

core/src/cpus/cortex_m.zig
tact1m4n3: I reverted some of those changes. Idk what I was thinking :)

core/src/cpus/cortex_m.zig
tact1m4n3: I reverted back to _start being callconv(.c)

core/src/cpus/cortex_m.zig
tact1m4n3: It just sets up the stack pointer now. I think we can just set it directly and not fetch it from the vector table, no?

tact1m4n3: Something like this:
```zig
    pub fn ram_image_start() linksection("microzig_ram_start") callconv(.naked) void {
        asm volatile (
            // Set up stack and jump to _start
            \\msr msp, %[eos]
            \\b _start
            :
            : [eos] "r" (microzig.config.end_of_stack),
            : "memory" // <-- is this needed?
        );
    }
```

Grazfather: Not sure if the `"memory"` part is needed, but it should work.

tact1m4n3: Was there a reason to read those from the vector table (like it was before)?

Grazfather: I don't remember. I probably didn't know that the stack pointer symbol was exposed to zig code. I basically just followed how it's typically done

PR #642, opened 2025-08-12
I2c interface

Grazfather:


drivers/base/I2C_Device.zig
```diff
@@ -0,0 +1,112 @@
...
+
+    /// Reads multiple `datagrams` from the device.
+    /// Function returns the number of bytes written in `datagrams`.
+    pub fn readv(dev: I2C_Device, datagrams: []const []u8) I2CErrorI!usize {
+        const readv_fn = dev.vtable.readv_fn orelse return I2CErrorI.Unsupported;
+        return readv_fn(dev.ptr, datagrams);
+    }
+
+    pub const VTable = struct {
+        set_address_fn: ?*const fn (*anyopaque, I2CAddress) I2CErrorI!void,
```
Grazfather: Other than the error types being adjusted, this is the only change from the datagram interface.

Does it make sense to add anything else?
Does it make sense to remove any of these?

drivers/base/I2C_Device.zig
```diff
@@ -0,0 +1,304 @@
+//!
+//! An abstract I2C device with runtime dispatch
+//!
+
+const std = @import("std");
+
+/// I2CError is a set of errors that make sense for I2C at the protocol level
+pub const I2CError = error{
```
tact1m4n3: shouldn't this be `I2C_Error`?

drivers/base/I2C_Device.zig
```diff
@@ -0,0 +1,304 @@
...
+    Timeout,
+    TargetAddressReserved,
+    NoData,
+    BufferOverrun,
+    UnknownAbort,
+    IllegalAddress,
+};
+// I2CInterfaceError adds additional errors that only make sense for the interface (e.g. unsupported
+// feature).
+pub const I2CInterfaceError = I2CError || error{Unsupported};
```
tact1m4n3: naming convention

drivers/base/I2C_Device.zig
```diff
@@ -0,0 +1,304 @@
...
+
+///
+/// 7-bit I²C address, without the read/write bit.
+///
+pub const I2CAddress = enum(u7) {
```
tact1m4n3: naming convention

drivers/base/I2C_Device.zig
```diff
@@ -0,0 +1,304 @@
...
+/// Function returns the number of bytes written in `datagrams`.
+pub fn readv(dev: I2C_Device, datagrams: []const []u8) I2CInterfaceError!usize {
+    const readv_fn = dev.vtable.readv_fn orelse return I2CInterfaceError.Unsupported;
+    return readv_fn(dev.ptr, datagrams);
+}
+
+pub const VTable = struct {
+    set_address_fn: ?*const fn (*anyopaque, I2CAddress) I2CInterfaceError!void,
+    writev_fn: ?*const fn (*anyopaque, datagrams: []const []const u8) I2CInterfaceError!void,
+    readv_fn: ?*const fn (*anyopaque, datagrams: []const []u8) I2CInterfaceError!usize,
```
tact1m4n3: Why are these optional? Wouldn't all i2c devices have these defined?

port/raspberrypi/rp2xxx/src/hal/i2c.zig
```diff
@@ -26,48 +27,13 @@ pub const Config = struct {
...
-    TxFifoFlushed,
-    UnknownAbort,
-};
+pub const Address = drivers.I2C_Device.I2CAddress;
+// NOTE In this context we are adding chip-specific errors. The naming is messy, I'd rather name
+// this error set something that indicates it includes device-specific errors, and the interface
+// implementation is where where'd map these to the smaller generic error set.
+// But we also can do that here, which simplifies the interface implementation, while losing error
+// detail for any client that uses the direct driver implementation.
+pub const I2CError = drivers.I2C_Device.I2CError || error{TxFifoFlushed};
```
tact1m4n3: naming convention

drivers/base/I2C_Device.zig
tact1m4n3: also, maybe we could drop the I2C prefix for these since we already have it in the namespace? eg: I2CError => Error

drivers/base/I2C_Device.zig
tact1m4n3: also I2CInterfaceError => InterfaceError

drivers/base/I2C_Device.zig
Grazfather: Seems that we do not add underscores to error types (e.g. `TransactionError`). Happy to remove the I2C part.

drivers/base/I2C_Device.zig
Grazfather: could technically only support read or write, and definitely not write then read. That's what the `Unsupported` error is for.

tact1m4n3 (approved):

PR #635, opened 2025-08-04
rp2xxx: Remove SchmittTrigger, use  Enabled

Gnyblast:
Code:
```zig
const pin_config = rp2xxx.pins.GlobalConfiguration{
    .GPIO29 = .{
        .name = "relay",
        .direction = .out,
    },
    .GPIO28 = .{
        .name = "sensor",
        .direction = .in,
    },
    .GPIO27 = .{
        .name = "capacitor",
        .direction = .in,
    },
    .GPIO4 = .{
        .name = "SDA",
        .function = .I2C0_SDA,
        .schmitt_trigger = .enabled,
        .slew_rate = .slow,
    },
    .GPIO5 = .{
        .name = "SCL",
        .function = .I2C0_SCL,
        .schmitt_trigger = .enabled,
        .slew_rate = .slow,
    },
};

pub fn main() void {
    pin_config.apply();
}
```

Causing:
```sh
install
└─ install generated to light-mcu.uf2
   └─ run elf2uf2 (light-mcu.uf2)
      └─ zig build-exe light-mcu ReleaseSmall thumb-freestanding-eabi 1 errors
/home/guney/.cache/zig/p/mz_port_raspberrypi_rp2xxx-0.0.0-wCvtidmzDQDWXpsuqBY_8dkz96PsJ_rrMKX04haxvjRt/src/hal/pins.zig:944:60: error: expected type 'hal.gpio.Enabled', found 'hal.gpio.SchmittTrigger'
                    gpio.num(gpio_num).set_schmitt_trigger(enabled);
```

When I dig into the code base. I realized that gpio config uses `SchmittTrigger` but `set_schmitt_trigger` in `pins.zig` uses `Enabled` so if you pass `SchmittTrigger` enum the pins.zig fails, if you pass `Enabled` then `gpio.zig` fails.

I honestly couldn't see if there was a purpose of having them twice(ish), but looks like causing trouble if I'm not really doing something completely stupid.

PR #643, opened 2025-08-13
rp2xxx: Fix pins.zig and examples

tact1m4n3:
Should fix all ci errors of main


Grazfather:
Broken from #635 I presume?

tact1m4n3:
Yep

PR #644, opened 2025-08-14
rp2xxx: Add set_interrupts_enabled and non blocking read to uart hal

tact1m4n3:


port/raspberrypi/rp2xxx/src/hal/uart.zig
```diff
@@ -60,9 +60,11 @@ pub const ReceiveError = error{
     BreakError,
     ParityError,
     FramingError,
-    Timeout,
 };
 
+pub const ReceiveNonBlockingError = ReceiveError || error{WouldBlock};
+pub const ReceiveBlockingError = ReceiveError || error{Timeout};
```
tact1m4n3: Is splitting the errors here ok?

port/raspberrypi/rp2xxx/src/hal/uart.zig
```diff
@@ -367,14 +400,21 @@ pub const UART = enum(u1) {
...
         try uart.read_blocking(&byte, timeout);
         return byte[0];
     }
 
+    /// Read a single byte from the RX line if available otherwise returns
+    /// `error.WouldBlock`.
+    pub fn read_word(uart: UART) ReceiveNonBlockingError!u8 {
+        if (!uart.is_readable()) return ReceiveNonBlockingError.WouldBlock;
+        return try uart.read_rx_fifo_with_error_check();
+    }
```
ikskuh: Why not just make that function be `ReceiveError!?u8` instead of `error.WouldBlock`?

Then you can do something like this:
```zig
while(try uart.read_word()) |word| {
    …
}
```

port/raspberrypi/rp2xxx/src/hal/uart.zig
ikskuh: Fine for me. We can also drop the `ReceiveNonBlockingError` (see comment below)

ikskuh (changes requested):

mattnite (approved):
Love to see improvements like this

PR #645, opened 2025-08-14
Overhaul i2c_device interfaces, implement, update some drivers

Grazfather:
# Refactor I2C Device Interface: Split into I2C_Device and I2C_Datagram_Device

##  Summary

  - Split the I2C device interface into two specialized interfaces: I2C_Device and I2C_Datagram_Device to better separate concerns and provide more appropriate abstractions.
  - Updated HAL implementations across multiple platforms (Raspberry Pi RP2xxx, Nordic nRF5x, ESP32) to support the new interface structure.
  - Converted device drivers and examples to use the refactored interfaces

## Changes Made

  - Core Interface Changes: Modified _drivers/base/I2C_Device.zig_ to implement the new split architecture
  - Platform HAL Updates: Updated I2C implementations in:
    - port/raspberrypi/rp2xxx/src/hal/ (i2c.zig, drivers.zig)
    - port/nordic/nrf5x/src/hal/ (i2c.zig, i2cdma.zig, drivers.zig)
    - port/espressif/esp/src/hal/ (i2c.zig, drivers.zig)
    - port/stmicro/stm32/src/hals/STM32F103/ (i2c.zig, drivers.zig)
  - Driver Updates: Converted device drivers to new interfaces:
    - ICM-20948 accelerometer driver
    - MLX90640 thermal camera driver
    - PCF8574 I/O expander driver
    - TLV493D hall effect sensor
    - TMP117 temperature sensor
  - Example Updates: Updated I2C examples across platforms and added new i2c_accel.zig example for RP2xxx


port/nordic/nrf5x/src/hal/drivers.zig
```diff
@@ -89,14 +98,23 @@ pub const I2C_Device = struct {
...
             error.TargetAddressReserved,
+            error.IllegalAddress,
             => error.Unsupported,
 
+            error.BufferOverrun,
+            error.DeviceNotPresent,
+            error.NoAcknowledge,
             error.UnknownAbort,
             error.Overrun,
             => error.IoError,
```
Grazfather, line 106: This rearrangement makes more sense to me but it's up for debate

drivers/base/I2C_Device.zig
```diff
@@ -27,15 +27,37 @@ pub const Address = enum(u7) {
     /// The general call addresses all devices on the bus using the I²C address 0.
     pub const general_call: Address = @enumFromInt(0x00);
 
+    const AllowReserved = enum { AllowReserved, DontAllowReserved };
```
tact1m4n3: These should be snake case: `allow_reserved`, `dont_allow_reserved`/`disallow_reserved`

tact1m4n3 (approved):
Looks good!

drivers/base/I2C_Device.zig
```diff
@@ -27,15 +27,37 @@ pub const Address = enum(u7) {
...
-    /// Returns `true` if the Address is a reserved I²C address.
+    /// Returns an Address.Error if the Address is a reserved I²C address.
+    /// The error gives detail on why the address is reserved, allowing the client to determine
+    /// whether it should allow it.
     ///
     /// Reserved addresses are ones that match `0b0000XXX` or `0b1111XXX`.
     ///
     /// See more here: https://www.i2c-bus.org/addressing/
-    pub fn is_reserved(addr: Address) bool {
+    pub fn is_reserved(addr: Address) Address.Error!void {
```
tact1m4n3: The naming here is a bit confusing. I would expect something starting with `is_` to return a bool.

drivers/base/I2C_Device.zig
Grazfather: Done in #648

drivers/base/I2C_Device.zig
Grazfather: Yeah. I don't like the name either. I could make it return an error or false?

Grazfather: check_reserved work?

port/nordic/nrf5x/src/hal/i2c.zig
```diff
@@ -220,15 +182,18 @@ pub const I2C = enum(u1) {
         });
     }
 
-    fn set_address(i2c: I2C, addr: Address) void {
+    fn set_address(i2c: I2C, addr: Address) Error!void {
+        // TODO: Allow arg
+        addr.check_reserved() catch return Error.IllegalAddress;
```
tact1m4n3: Isn't this check done twice? One time here and the other in the write function/read functions

tact1m4n3: Isn't this check done twice? One time here and the other in the write function/read functions

port/nordic/nrf5x/src/hal/i2c.zig
```diff
@@ -279,18 +244,20 @@ pub const I2C = enum(u1) {
     ///       suffixes won't need to be concatenated/inserted to the original buffer, but can be managed
     ///       in a separate memory.
     ///
-    pub fn writev_blocking(i2c: I2C, addr: Address, chunks: []const []const u8, timeout: ?mdf.time.Duration) TransactionError!void {
-        if (addr.is_reserved())
-            return TransactionError.TargetAddressReserved;
+    pub fn writev_blocking(i2c: I2C, addr: Address, chunks: []const []const u8, timeout: ?mdf.time.Duration) Error!void {
+        addr.check_reserved() catch |err| switch (err) {
```
tact1m4n3: Here

port/nordic/nrf5x/src/hal/i2c.zig
tact1m4n3: Duplicate :)

Grazfather: Yes, and we also have `IllegalAddress` and `TargetAddressReserved`. I think we can probably combine them, or define when each is used.

I think we want to use the `AllowReserved` rg to be passed to set address, and that would imply that a failure is not that the address is reserved, but just that it's illegal (e.g. write with general is reserved by legal in this context). (That's what the TODO is for). I have like 3 branches stacked on this one, so I'll pull some changes down to resolve that.

Grazfather: Check out 145ce5a0 and lmk if that looks OK to you.

Grazfather: Another weird thing is that `set_address` in the interface only sets the address field in the struct, and write and read is called with that argument, then a 'real' set address is called in those functions to configure the peripheral.

Maybe these write/read shouldn't take a address, and should count on the address already being set? And then the interface can actually call the hal set_address.

drivers/base/I2C_Device.zig
tact1m4n3: Yep

PR #647, opened 2025-08-15
rp2xxx: rom refactoring and support for rp2350

tact1m4n3:
- [x] add remaining rp2350 methods
- [x] move stuff around

For the future:
- export rom optimized math functions for rp2040


piotrfila:
Nice! The rom functions needed a rework and this is excellent.

PR #648, opened 2025-08-15
I2C_Device: Make is_reserved return an AddressError

Grazfather:
Also, make `I2C_Device.set_address` take an `allow_reserved` argument.

### DIFF
--- a/drivers/base/I2C_Device.zig
+++ b/drivers/base/I2C_Device.zig
@@ -77,7 +77,7 @@ ptr: *anyopaque,
 /// Virtual table for the datagram device functions.
 vtable: *const VTable,
 
-pub fn set_address(dev: I2C_Device, addr: Address, allow_reserved: AllowReserved) InterfaceError!void {
+pub fn set_address(dev: I2C_Device, addr: Address, allow_reserved: Allow_Reserved) InterfaceError!void {
     const set_address_fn = dev.vtable.set_address_fn orelse return InterfaceError.Unsupported;
     return set_address_fn(dev.ptr, addr, allow_reserved);
 }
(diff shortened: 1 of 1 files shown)


Grazfather:
This was discussed a bit on the discord server. I am OK if we don't want to go with the approach, since it almost forces the i2c_device interface more complicated (since we want to plumb through whether or not an invalid address is OK).

Basically:

1. Write to i2c functions should probably allow writing to GeneralCall address only (needed for TLV493D for example).
2. Read should not allow any restricted address
3. Write-then-read should not
4. Supporting 10-bit addresses will probably require working around certain 'reserved' addresses.
5. `set_address` should probably work for GeneralCall, but not for reads.

1-4 are nicely handled by the error type.
#5 I handle by adding a extra arg to set_address. This is not ideal. We could also just have it not check whether the address is valid, and have the chips fall as they may (maybe erroring out on a subsequent read).

drivers/base/I2C_Device.zig
```diff
@@ -100,8 +121,10 @@ pub fn readv(dev: I2C_Device, datagrams: []const []u8) InterfaceError!usize {
     return readv_fn(dev.ptr, datagrams);
 }
 
+const AllowReserved = enum { AllowReserved, DontAllowReserved };
```
ikskuh: Use `snake_case` here! I would also probably solve the "allow generic call or not" with this?

ikskuh (changes requested):

drivers/base/I2C_Device.zig
Grazfather: `AllowGeneric` could be added, that's a good idea, and we could add it to set_address -- I'll do that.

But for `writev` we don't want to add that argument, so we'll still have to just explicitly allow it like so:

```zig
    pub fn writev_blocking(i2c: I2C, addr: Address, chunks: []const []const u8, timeout: ?mdf.time.Duration) Error!void {
        addr.is_reserved() catch |err| switch (err) {
            .GeneralCall => void,
            else => return Error.TargetAddressReserved,
        };
```
which used to be `if (addr.is_reserved()) return Error.TargetAddressReserved;`

ikskuh: > But for writev we don't want to add that argument, so we'll still have to just explicitly allow it like so:

Why wouldn't we want that argument there? :thinking:

Grazfather: I suppose we could, I just don't think that we want to add more to the writev/writev_blocking/write/write_blocking/write_then_read/read/readv etc argument list. I think it's probably in general to just allow general for writing but not reading, since as far as I can tell none of the other types make sense in any other context.

We could maybe make it an option in the hal? so we'd do like `i2c0.apply(.{..., .allow_reserved = .AllowGeneral})`

Grazfather: But also that can be figured out later since none of the HALs implement this interface yet.

drivers/base/I2C_Device.zig
```diff
@@ -100,8 +121,10 @@ pub fn readv(dev: I2C_Device, datagrams: []const []u8) InterfaceError!usize {
     return readv_fn(dev.ptr, datagrams);
 }
 
+const Allow_Reserved = enum { AllowGeneral, AllowReserved, DontAllowReserved };
```
ikskuh: Can you make them snake_case? I'm fine with the PR then

ikskuh (changes requested):

PR #649, opened 2025-08-16
feat: add nxp mcx port

rmrt1n:
Added support for NXP MCX series. Tested only on the MCXA153 since that's the only one I have.

ATM the only feature of the HAL is GPIO, but I plan on adding more soon.


examples/nxp/mcx/src/blinky.zig
```diff
@@ -0,0 +1,24 @@
+const microzig = @import("microzig");
+const hal = microzig.hal;
+
+const port3 = hal.port.init(3);
+const pin_led_red = port3.get_gpio(12);
```
rmrt1n: I went with `init` instead of `num` like in the other ports since `init` sounds more "ziggish", wdyt? Or is `num` preferred for these kind of functions?

tact1m4n3: Imo num is preferred (at least for now). Also if we ever decide to change it in the future we can grep for it :)

tact1m4n3:
Instead of adding the svd to the repo you can just use the github repo where you've taken it as a dependency and fetch it automatically. You can look at rp2xxx port for an example. The svd is taken from pico sdk.

examples/nxp/mcx/src/blinky.zig
rmrt1n: updated

port/nxp/mcx/src/hal/gpio.zig
```diff
@@ -0,0 +1,114 @@
...
+
+    pub fn toggle(gpio: GPIO) void {
+        const registers = gpio.get_registers();
+        const old: u32 = registers.PTOR.raw;
+
+        registers.PTOR.write_raw(old | gpio.get_mask());
+    }
+
+    pub fn set_direction(gpio: GPIO, direction: Direction) void {
+        const registers = gpio.get_registers();
```
tact1m4n3: nit: Could you use regs/get_regs instead of registers just to be consistent with the other hals?

tact1m4n3:
Also, could you please add to ci a command to build the examples?

rmrt1n:
> Also, could you please add to ci a command to build the examples?

Added to the existing `build-examples` job!

tact1m4n3:
Thanks!

PR #652, opened 2025-08-20
General bug fixing

tact1m4n3:
* fix rp2040 crash
* rp2040: remove rom intrinsics for now
* fix flash image bss not getting initialised (discovered by accident while working on the thing above)
* fix `rom.lookup_with_cache_xxxx`: it was previously creating only one static cache no matter the data/function code (not the behaviour I was expecting)


piotrfila:
Was there some problem with the intrinsics?

PR #653, opened 2025-08-22
feat: add fpu and dcp support

tact1m4n3:
* add fpu register definitions to cortex_m33
* add dcp intrinsics
* enable fpu and dcp in hal

### DIFF
--- a/core/src/cpus/cortex_m.zig
+++ b/core/src/cpus/cortex_m.zig
@@ -656,10 +656,12 @@ const systick_base = scs_base + 0x0010;
 const nvic_base = scs_base + 0x0100;
 const scb_base = scs_base + core.scb_base_offset;
 const mpu_base = scs_base + 0x0D90;
+const fpu_base = scs_base + 0x0F34;
 
 const properties = microzig.chip.properties;
 // TODO: will have to standardize this with regz code generation
-const mpu_present = @hasDecl(properties, "__MPU_PRESENT") and std.mem.eql(u8, properties.__MPU_PRESENT, "1");
+const mpu_present = @hasDecl(properties, "cpu.mpuPresent") and std.mem.eql(u8, properties.@"cpu.mpuPresent", "true");
+const fpu_present = @hasDecl(properties, "cpu.fpuPresent") and std.mem.eql(u8, properties.@"cpu.fpuPresent", "true");
 
 const core = blk: {
     break :blk switch (cortex_m) {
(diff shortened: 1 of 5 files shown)


port/raspberrypi/rp2xxx/src/hal.zig
```diff
@@ -72,9 +93,40 @@ pub inline fn init() void {
...
+        }
+
+        if (microzig.options.hal.use_dcp) {
+            // enable the DCP
+            cpacr |= 0b11 << 8;
+        }
+
+        microzig.cpu.peripherals.scb.CPACR = cpacr;
+    }
+
```
tact1m4n3: Is it fine to enable the FPU here? Should we make state preservation configurable?

ikskuh: Imho it's completely fine to enable it by default! It's definitly giving us benefits. 

Alternatively we could make it depend on the selected CPU features if it's going to get enabled or not

tact1m4n3: > Alternatively we could make it depend on the selected CPU features if it's going to get enabled or not

This is already done :) Though not by using cpu features but by checking the abi is soft of hard. If it is soft the fpu gets automatically disabled by default and gives a compiler error if you enable the option

tact1m4n3: Would it better to check the cpu features instead of the abi?

core/src/cpus/cortex_m.zig
```diff
@@ -692,7 +700,7 @@ pub const peripherals = struct {
     pub const mpu: *volatile types.peripherals.MemoryProtectionUnit = if (mpu_present)
         @ptrFromInt(mpu_base)
     else
-        @compileError("This chip does not have a MPU.");
+        @compileError("this CPU does not have a MPU");
```
Grazfather:
```suggestion
        @compileError("this CPU does not have an MPU");
```

core/src/cpus/cortex_m.zig
```diff
@@ -715,6 +723,12 @@ pub const types = struct {
         /// System Control Block (SCB).
         pub const SystemControlBlock = core.SystemControlBlock;
 
+        /// Floating Point Unit (FPU).
+        pub const FloatingPointUnit = if (@hasDecl(core, "FloatingPointUnit"))
+            core.FloatingPointUnit
+        else
+            @compileError("this CPU does not have a FPU definition");
```
Grazfather:
```suggestion
            @compileError("this CPU does not have an FPU definition");
```

core/src/cpus/cortex_m.zig
```diff
@@ -779,6 +793,6 @@ pub const types = struct {
         pub const MemoryProtectionUnit = if (@hasDecl(core, "MemoryProtectionUnit"))
             core.MemoryProtectionUnit
         else
-            @compileError("This cpu does not have a MPU.");
+            @compileError("this CPU does not have a MPU definition");
```
Grazfather:
```suggestion
            @compileError("this CPU does not have an MPU definition");
```

port/raspberrypi/rp2xxx/src/hal.zig
```diff
@@ -56,6 +67,16 @@ pub const HAL_Options = switch (compatibility.chip) {
...
             next_block: ?*const anyopaque = null,
         } = .{},
+
+        /// Enable the FPU and lazy state preservation. Leads to much faster
+        /// single precision floating point arithmetic. If you want a custom
+        /// setup set this to false and configure the fpu yourself. Ignored on
+        /// riscv.
+        enable_fpu: bool = is_fpu_used,
+
+        /// Enable the DSP and export intrinsics. Leads to faster double
```
Grazfather: DCP

PR #654, opened 2025-08-22
drivers: Add MPU-6050 driver

tact1m4n3:


drivers/sensor/MPU-6050.zig
```diff
@@ -0,0 +1,336 @@
...
+        var value: u8 = undefined;
+        try self.dev.write_then_read(&.{@intFromEnum(reg)}, (&value)[0..1]);
+        return value;
+    }
+
+    inline fn read_reg(self: MPU_6050, reg: Register, T: type) !T {
+        return @bitCast(try self.read_byte(reg));
+    }
+
+    inline fn modify_reg(self: MPU_6050, reg: Register, T: type, fields: anytype) !void {
```
Grazfather: Nice. This looks like ICM-20948. I can do in followup (happy to change that driver to follow this one) but that driver uses `modify_register` and `read_register` vs. reg. Either name works, but let's try to be consistent.

drivers/sensor/MPU-6050.zig
```diff
@@ -0,0 +1,336 @@
...
+//! TODO:
+//! * MPU-6000
+//! * Add missing functionality (maybe also getters for some registers)
+//! * DMP support
+
+const mdf = @import("../framework.zig");
+const Datagram_Device = mdf.base.Datagram_Device;
+const Clock_Device = mdf.base.Clock_Device;
+
+pub const MPU_6050 = struct {
```
Grazfather, line 29: Would be nice to define an error type if/when we can tell something better than what datagram errors can tell us.

Grazfather: But maybe we don't really have anything better, if so, that's cool.

Grazfather (approved):

drivers/sensor/MPU-6050.zig
tact1m4n3: I added some errors. Let me know what you think

drivers/sensor/MPU-6050.zig
tact1m4n3: Hmmm. Tough choice :) I generally like to type less and reg seems explicit enough. We use `regs` in the hals quite a bit (`get_regs`). But idk. I'll let you decide

tact1m4n3: > This looks like ICM-20948

I took inspiration from your implementation on this :)

drivers/base/Datagram_Device.zig
```diff
@@ -23,6 +23,7 @@ const BaseError = error{ IoError, Timeout };
 pub const ConnectError = BaseError || error{DeviceBusy};
 pub const WriteError = BaseError || error{ Unsupported, NotConnected };
 pub const ReadError = BaseError || error{ Unsupported, NotConnected, BufferOverrun };
+pub const AnyError = ConnectError || WriteError || ReadError;
```
tact1m4n3, line 26: What do you think of this? I went with this solution as I think more drivers would need these combined.

drivers/sensor/MPU-6050.zig
Grazfather: I can update mine to match

drivers/sensor/MPU-6050.zig
Grazfather: Yeah looks good to me. Things might change if i figure out the i2c_device thing but this is good for now.

drivers/sensor/MPU-6050.zig
```diff
@@ -0,0 +1,412 @@
...
+//! Generic driver for the MPU-6050 accelerometer/gyroscope(/temperature)
+//! sensor.
+//!
+//! Datasheet:
+//! * https://invensense.tdk.com/wp-content/uploads/2015/02/MPU-6000-Datasheet1.pdf
+//! * https://invensense.tdk.com/wp-content/uploads/2015/02/MPU-6000-Register-Map1.pdf
+//!
+//! Example usage:
+//! ```zig
+//! const mpu_6050: MPU_6050 = try .init(i2c_driver.datagram_device(), clock_device.clock_device());
```
Grazfather, line 11: this would change (clock device) with your other PR

drivers/sensor/MPU-6050.zig
```diff
@@ -0,0 +1,412 @@
...
+        const data = try self.read_raw_data(.acc_regx_h, 7);
+        return .{
+            .accel = .{ data[0], data[1], data[2] },
+            .temp = data[3],
+            .gyro = .{ data[4], data[5], data[6] },
+        };
+    }
+
+    pub fn read_accel_temp_gyro(self: MPU_6050) Error!struct {
+        accel: [3]f32,
```
Grazfather: might be nice to split these (could share a type for accel and gyro) so we can access .x .y and .z

drivers/sensor/MPU-6050.zig
```diff
@@ -0,0 +1,412 @@
...
+            ZA_ST: bool,
+            YA_ST: bool,
+            XA_ST: bool,
+        };
+
+        pub const USER_CTRL = packed struct(u8) {
+            SIG_COND_RESET: bool,
+            I2C_MST_RESET: bool,
+            FIFO_RESET: bool,
+            reserved0: u1,
```
Grazfather: usually we put the num here say which bit it overlaps so this would be reserved3

drivers/sensor/MPU-6050.zig
```diff
@@ -0,0 +1,412 @@
...
+            I2C_IF_DIS: bool,
+            I2C_MST_EN: bool,
+            FIFO_EN: bool,
+            reserved1: u1,
```
Grazfather: reserved7

drivers/sensor/MPU-6050.zig
tact1m4n3: Yeah! We will have to merge them in order. I would merge this one first and fix all comments in the other one (There is a similar thing in ICM-20948).

Grazfather (approved):

drivers/sensor/MPU-6050.zig
tact1m4n3: Oh ok. Didn't know this

tact1m4n3:
@Grazfather perhaps in the future we can define a common interface for IMUs and also write some generic calibration algorithms and such

Grazfather:
Yes I like that. Device specific errors can just be ORed in like I am trying to do with i2c device plus Hal specific i2c errors eg fifo error on nrf52

drivers/framework.zig
```diff
@@ -43,10 +43,11 @@ pub const led = struct {
 };
 
 pub const sensor = struct {
-    pub const TMP117 = @import("sensor/TMP117.zig").TMP117;
     pub const ICM_20948 = @import("sensor/ICM-20948.zig").ICM_20948;
     pub const MLX90640 = @import("sensor/MLX90640.zig").MLX90640;
+    pub const MPU_6050 = @import("sensor/MPU-6050.zig").MPU_6050;
```
Grazfather, line 48: Want to add the import in the test part at the bottom (even if there are no tests yet, so that they actually run if/when we add them)

PR #657, opened 2025-08-24
I2C STYLE TEST - for REVIEW ONLY

RecursiveError:
just demonstration of how to include "i2c_datagram device" directly in 'I2C Device"


RecursiveError:
@Grazfather

Issue #452, opened 2025-04-17
RP2xxx USB driver - memcpy issue


arkadiuszwojcik:
There is strange issue in USB driver `usb_start_tx` function when `@memcpy ` is used.

For Zig 0.14:
- RP2350 crashes when @memcpy is used but RP2040 is fine
- RP2040 crashes when `rom.memcpy` code is used even before zig 0.14 update this was working fine
- `std.mem.copyForwards` seems to work for both - but we probably should use it as just temporary solution

PR #658, opened 2025-08-26
Rework rp2xxx USB driver

piotrfila:
This PR is my shot at a better USB driver implementation.

Key changes:
- Multi-function (composite) devices now possible
- Descriptors are automatically constructed at compile time by drivers
- More type safety around descriptors
- Fixed a bug where tx data was lost if inconsistent buffer sizes were used
- Limited use of dynamic dispatch

I'm open to suggestions for changes


core/src/core/usb/types.zig
```diff
@@ -178,313 +154,18 @@ pub const SetupPacket = extern struct {
...
-    }
-
-    /// Driver open (set config) operation. Must return length of consumed driver config data.
-    pub fn open(self: *@This(), cfg: []const u8) !usize {
-        return self.fn_open(self.ptr, cfg);
-    }
-
-    pub fn class_control(self: *@This(), stage: ControlStage, setup: *const SetupPacket) bool {
-        return self.fn_class_control(self.ptr, stage, setup);
+    pub fn from(val: u16) @This() {
```
arkadiuszwojcik, line 160: We probably should check native `val` endianness and make `@byteSwap` if necessary or utilize `std.mem.writeInt`

core/src/core/usb/types.zig
```diff
@@ -178,313 +154,18 @@ pub const SetupPacket = extern struct {
...
+        return .{
+            .lo = @truncate(val),
+            .hi = @intCast(val >> 8),
+        };
     }
 
-    pub fn transfer(self: *@This(), ep_addr: u8, data: []u8) void {
-        return self.fn_transfer(self.ptr, ep_addr, data);
+    pub fn into(this: @This()) u16 {
```
arkadiuszwojcik, line 166: Same here. We should check native endianness and make `@byteSwap` if necessary before return or utilize `std.mem.readInt`

core/src/core/usb/types.zig
```diff
@@ -79,36 +79,40 @@ pub const FeatureSelector = enum(u8) {
...
 
-pub const Endpoint = struct {
-    pub inline fn to_address(num: u8, dir: Dir) u8 {
-        return switch (dir) {
-            .Out => num,
-            .In => num | Dir.DIR_IN_MASK,
-        };
-    }
+pub const Endpoint = packed struct(u8) {
+    // There are up to 15 endpoints for eqch direction.
```
arkadiuszwojcik: typo: 'each'

core/src/core/usb/descriptor/cdc.zig
```diff
@@ -0,0 +1,77 @@
...
+    bcd_cdc: types.U16Le,
+};
+
+pub const CallManagement = extern struct {
+    comptime {
+        assert(@alignOf(@This()) == 1);
+        assert(@sizeOf(@This()) == 5);
+    }
+
+    length: u8 = 5,
```
arkadiuszwojcik: shouldn’t we use `= @sizeOf(@This())` for consistency?

core/src/core/usb/descriptor/cdc.zig
```diff
@@ -0,0 +1,77 @@
...
+    capabilities: u8,
+};
+
+pub const Union = extern struct {
+    comptime {
+        assert(@alignOf(@This()) == 1);
+        assert(@sizeOf(@This()) == 5);
+    }
+
+    length: u8 = 5,
```
arkadiuszwojcik: shouldn’t we use `= @sizeOf(@This())` for consistency?

core/src/core/usb/cdc.zig
```diff
@@ -1,311 +1,210 @@
...
 
-        pub fn write_flush(self: *@This()) usize {
-            if (self.device.?.ready() == false) {
-                return 0;
+    /// Read data from rx buffer into dst.
+    pub fn read(this: *@This(), device: usb.DeviceInterface, dst: []u8) usize {
+        if (this.rx_buf) |rx| {
+            const len = @min(rx.len, dst.len);
+            // TODO: please fixme: https://github.com/ZigEmbeddedGroup/microzig/issues/452
+            std.mem.copyForwards(u8, dst, rx[0..len]);
```
arkadiuszwojcik: Regarding those "TODO #452" `memcpy` should be part of `DeviceInterface` I guess as it is in TinyUSB and CherryUSB projects

core/src/core/usb.zig
```diff
@@ -1,683 +1,390 @@
...
+    pub fn submit_tx_buffer(this: @This(), ep_in: EpNum, buffer_end: [*]const u8) void {
+        this.vtable.submit_tx_buffer(this.ptr, ep_in, buffer_end);
+    }
+    /// Called by drivers to report readiness to receive up to `len` bytes.
+    /// Must be called exactly once before each packet.
+    pub fn signal_rx_ready(this: @This(), ep_out: EpNum, max_len: usize) void {
+        this.vtable.signal_rx_ready(this.ptr, ep_out, max_len);
+    }
+    /// Opens an endpoint according to the descriptor.
+    pub fn endpoint_open(this: @This(), desc: *const descriptor.Endpoint) ?[]u8 {
```
arkadiuszwojcik: Would be good to mention what return value represents

arkadiuszwojcik (commented):
Great rework! The USB drive code is much cleaner now. Here are my initial comments. I'll try to more detailed review later.

core/src/core/usb/types.zig
piotrfila: I know about the builtin, but decided to be extra verbose, so that it's obvious that the memory layout is correct. Especially since it would not be trivial to write tests for a big endian architecture (the driver currently only supports rp2xxx, which are all little endian).

core/src/core/usb/descriptor/cdc.zig
piotrfila: fixed, thanks!

core/src/core/usb/descriptor/cdc.zig
piotrfila: fixed, thanks!

core/src/core/usb/types.zig
piotrfila: fixed, thanks!

core/src/core/usb.zig
piotrfila: Good catch, I meant to change the return value before submitting the PR and just forgot.
Fixed now.

core/src/core/usb/cdc.zig
piotrfila: I disagree, the `@memcpy ` builtin is meant to provide a good implementation regardless of the target.
After checking on actual hardware, it seems like the issue is gone, maybe because of the recent rom changes?

arkadiuszwojcik: Here is issue description in:
- CherryUSB: https://github.com/cherry-embedded/CherryUSB/issues/297 memcpy implementation in newlib (arm toolchain) is wrong.
- TinyUSB: https://github.com/liamfraser/tinyusb/blob/31a979a6cccc2d4dc0a86474643710ae3b7f34fe/src/portable/raspberrypi/rp2040/rp2040_usb.c#L57
- PicoSDK: https://github.com/raspberrypi/pico-sdk/blob/ee68c78d0afae2b69c03ae1a72bf5cc267a2d94c/src/rp2_common/pico_mem_ops/CMakeLists.txt#L15 (beware commpiler memcpy cannot be used for unaligned copies in peripheral space) 

Don't know what to think about it. At least I would extract it to some inline `usb_memcpy` function and use `std.mem.copyForwards` or `@memcpy` inside for now.

piotrfila: The usb DPRAM behaves like regular memory unlike most peripheral memory.

image

As per [zig documentation on @memcpy](https://ziglang.org/documentation/0.14.1/#memcpy) both source and destination can have any alignment.

I can't comment on how good the memcpy implementation is, but judging from the lack of HardFault exceptions raised it must be correctly handling unaligned data. I also disassembled the .elf file of the hid example and the implementation of __aeabi_memcpy uses ldrb and strb instructions, which work on unaligned data.

core/src/core/usb/types.zig
piotrfila: I thought about it and decided to use std.mem.readInt and writeInt like you suggested.

arkadiuszwojcik:
@piotrfila, what toolchain version are you using, and which boards were tested? I am using Zig 0.14.1 on Windows and tried to run this code on the RP2040 and RP2350 (Arm cores). For the RP2350, I get some Windows USB errors, and the RP2040 is somewhat unstable—sometimes it works, and sometimes it doesn't.

piotrfila:
I am also using 0.14.1, but I have done all the testing under Linux, where RP2040 seems to work fine.
RP2350 indeed HardFaults even before enumeration, oops. I should have tested that.

core/src/core/usb/cdc.zig
piotrfila: Not sure how I got this mixed up, but the whole time I was thinking it was the RP2040 that had problems with memcpy. #452 of course clearly states the bug affects RP2350...

And unsurprisingly the datasheet documents this:

image

This is rather annoying. How are we supposed to guarantee that the optimizer does not use unaligned LDR/LDRH? I originally wanted to avoid needlessly copying the buffers but that seems like the simplest option now.

arkadiuszwojcik: So we must read/write data to some intermediate buffer (in main memory) to make sure that aligned nonoptimized `memcpy `operation is used right?

Kytezign:
@piotrfila, @arkadiuszwojcik,

I'm curious if you've looked at somehow making this new implementation compatible at some level with tinyUSB.  I've been playing with linking the current main branch USB stuff up with a tinyUSB static library - it seems possible to me but I'm in over my head enough to know I'm likely not understanding everything right yet.

In my view, one of the major benefits of Zig is that we can leverage existing c libraries and this seems like a good opportunity to not have to re-invent the wheel.
With that said, hope this finds you well and please forgive my obvious ignorance here, I'm relatively new to zig, microzig and embedded programing.

arkadiuszwojcik:
@Kytezign Yes, TinyUSB and CherryUSB ports will probably be the default for the majority of users. The MicroZig custom USB stack implementation is mainly an experiment to see how far we can go with a custom implementation and whether we can design a more user-friendly API by leveraging Zig language features.

Kytezign:
> @Kytezign Yes, TinyUSB and CherryUSB ports will probably be the default for the majority of users. The MicroZig custom USB stack implementation is mainly an experiment to see how far we can go with a custom implementation and whether we can design a more user-friendly API by leveraging Zig language features.

Makes sense.  Are TinyUSB or CherryUSB ports already being worked on or available?
Like I said, I've played around with doing one a little bit but I'm really out of my depth so it's been slow and messy.

arkadiuszwojcik:
> Makes sense. Are TinyUSB or CherryUSB ports already being worked on or available? Like I said, I've played around with doing one a little bit but I'm really out of my depth so it's been slow and messy.

Nothing I am aware of. Would be nice to have such port as a part of MicroZig along with other libraries like lwIP.

Kytezign:
> Nothing I am aware of. Would be nice to have such port as a part of MicroZig along with other libraries like lwIP.

From my limited study of it, one easier route to enable this - at least for tinyUSB -  is if the standardized abstract usb interface was compatible with the tinyUSB interface.  That way all usb ports in microzig would be compatible.
One thing I'm not sure about is how hard other devices would be to enable.  For the RP2040, the HW implementation seems completely reliant on the pico_sdk including it's runtime.  I was not able to build it as a library for microzig to link with.

vesim987:
@piotrfila do you plan to work on this?

piotrfila:
I believe the whole PR needs rewriting from scratch. I'm not going to do it in the near future though since I have uni now. Maybe once the new Io interface comes out and I can rewrite it to be asynchronous?

That said, I now think it would be better to just include tinyusb or some alternative, it's really not feasible to reimplement even a good fraction of their drivers here.

PR #662, opened 2025-08-29
nrf5x: Use RTC for timekeeping, track overflows for extra bits

Grazfather:
The nrf TIMER is 32 bits, and with prescaler set to 4, it would count at exactly 1MHz.

This would overflow every 4294 seconds, which is not enough for many uses.

We could increase the prescaler up to 9, splitting the 32MHz clock to 62500Hz, which would still overflow every 68719 seconds, which is still less than a day.

The original plan was to set an interrupt to fire on overflow, to increment a counter, which could be used to add bits to this value, easily getting us 64 bits. The problem is that it introduced a race condition, because we could try to build the time extra when it overflowed, which would yield us an incorrect count (e.g. off by 4294 seconds). The way that [embassy](https://github.com/embassy-rs/embassy/blob/main/embassy-nrf/src/time_driver.rs) gets around this is by incrementing the counter when it overflows, as well as when it hits the halfway point, and then, depending on the value of the extra bits, choose to toggle the top bit of the counter or not.

Unfortunately, the TIMER register does not have the ability to fire an interrupt when it overflows, only when it hits a compare value. We could probably use two compare values (half way and on overflow), but I think it's probably better to follow the lead of embassy and use RTC, which has an explicit overflow event.

By user 23 bits of timer, with an interrupt that fires at 0x800000 and on overflow, we can track the time for 23 + 32 == 55 ticks.

This timer runs at a slower 32'768 Hz, which means we sacrifice granularity, but for timekeeping usually in the ms (not us) scale, it should not be a problem.

### DIFF
--- a/examples/nordic/nrf5x/src/blinky.zig
+++ b/examples/nordic/nrf5x/src/blinky.zig
@@ -4,6 +4,14 @@ const board = microzig.board;
 const nrf = microzig.hal;
 const time = nrf.time;
 
+const rtc_overflow_interrupt = nrf.time.rtc_overflow_interrupt;
+
+pub const microzig_options = microzig.Options{
+    .log_level = .debug,
+    .logFn = nrf.uart.log,
+    .interrupts = .{ .RTC0 = .{ .c = rtc_overflow_interrupt } },
+};
+
 pub fn main() !void {
     board.init();
 
--- a/port/nordic/nrf5x/src/hal.zig
+++ b/port/nordic/nrf5x/src/hal.zig
@@ -8,7 +8,7 @@ pub const spim = @import("hal/spim.zig");
 pub const time = @import("hal/time.zig");
 pub const uart = @import("hal/uart.zig");
 pub const drivers = @import("hal/drivers.zig");
-// TODO: adc, timers, pwm, rng, rtc, interrupts, i2c, wdt, wifi, nfc, bt, zigbee
+// TODO: adc, timers, pwm, rng, rtc alarms, interrupts, wdt, wifi, nfc, bt, zigbee
 
 pub fn init() void {
     time.init();
--- a/port/nordic/nrf5x/src/hal/time.zig
+++ b/port/nordic/nrf5x/src/hal/time.zig
@@ -21,64 +21,102 @@ const version: enum {
     else => compatibility.unsupported_chip("time"),
 };
 
-const timer = microzig.chip.peripherals.TIMER0;
-// const INTERRUPT_INDEX = 0;
-const READ_INDEX = 1;
+const timer = microzig.chip.peripherals.TIMER0; // DELETEME
+const rtc = microzig.chip.peripherals.RTC0;
+const interrupts = microzig.chip.peripherals.interrupts;
+const COMPARE_INDEX = 3;
+const TIMER_BITS = 23;
 
-var overflow_count: u32 = 0;
+var period: u32 = 0;
 
 pub fn init() void {
     // Stop timer while we set it up
-    switch (version) {
-        .nrf51 => {
-            timer.TASKS_STOP = 1;
-            defer timer.TASKS_START = 1;
+    // switch (version) {
+    //     .nrf51 => {
+    //         timer.TASKS_STOP = 1;
+    //         defer timer.TASKS_START = 1;
+    //
+    //         // Clear the time
+    //         timer.TASKS_CLEAR = 1;
+    //
+    //         timer.MODE.write(.{ .MODE = .Timer });
+    //         timer.BITMODE.write(.{ .BITMODE = .@"32Bit" });
+    //         // 16Mhz / 2^4 = 1MHz: us resolution
+    //         timer.PRESCALER.write(.{ .PRESCALER = 4 });
+    //     },
+    //     .nrf52 => {
+    //         timer.TASKS_STOP.write(.{ .TASKS_STOP = .Trigger });
+    //         defer timer.TASKS_START.write(.{ .TASKS_START = .Trigger });
+    //
+    //         // Clear the time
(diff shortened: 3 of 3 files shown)


examples/nordic/nrf5x/src/blinky.zig
```diff
@@ -4,6 +4,14 @@ const board = microzig.board;
 const nrf = microzig.hal;
 const time = nrf.time;
 
+const rtc_overflow_interrupt = nrf.time.rtc_overflow_interrupt;
+
+pub const microzig_options = microzig.Options{
+    .log_level = .debug,
+    .logFn = nrf.uart.log,
+    .interrupts = .{ .RTC0 = .{ .c = rtc_overflow_interrupt } },
```
Grazfather: We'd have to make this default. I think to do this we would have to merge the option cpu InterruptOptions with a new chip InterruptOptions

port/nordic/nrf5x/src/hal/time.zig
```diff
@@ -1,8 +1,9 @@
 ///
 /// Basic timekeeping for the nRF5x series MCUs.
 ///
-/// This module hogs TIMER0.
-/// It uses CC1 as the register to read the current count from.
+/// This module uses RTC0 and hogs CC[2]
```
Grazfather, line 4: For some reason the `COMPARE3` field in `INTENSET` won't get set for me, as in it never fires and reading it back shows the bit still as 0.

mattnite:
lol I feel attacked

port/nordic/nrf5x/src/hal.zig
```diff
@@ -8,7 +8,12 @@ pub const spim = @import("hal/spim.zig");
 pub const time = @import("hal/time.zig");
 pub const uart = @import("hal/uart.zig");
 pub const drivers = @import("hal/drivers.zig");
-// TODO: adc, timers, pwm, rng, rtc, interrupts, i2c, wdt, wifi, nfc, bt, zigbee
+// TODO: adc, timers, pwm, rng, rtc alarms, interrupts, wdt, wifi, nfc, bt, zigbee
+
+pub const default_interrupts = .{
```
tact1m4n3: You can give this the type microzig.cpu.InterruptOptions

tact1m4n3:
If we wanted to get really elaborate we could throw a compile error if the RTC interrupt handler is overwritten and hal.time is referenced. If the user wants to dismiss that warning he would have to set a flag in HAL_Options. Imo, these warnings that prevent the user from making a mistake should be part of the microzig at some point in the future but maybe now isn't the time :)

port/nordic/nrf5x/src/hal.zig
Grazfather: done, and added a TODO about failing

core/src/cpus/cortex_m.zig
```diff
@@ -612,6 +612,19 @@ pub const startup_logic = struct {
             .Reset = .{ .c = microzig.cpu.startup_logic._start },
         };
 
+        // Apply HAL-level default interrupts first (if any)
+        if (@hasDecl(microzig.hal, "default_interrupts")) {
```
tact1m4n3: The hal module is optional so it's necessary to check microzig.config.has_hal first

core/src/cpus/cortex_m.zig
```diff
@@ -612,6 +612,20 @@ pub const startup_logic = struct {
             .Reset = .{ .c = microzig.cpu.startup_logic._start },
```
tact1m4n3, line 694: Responding here to have more context :)) You can add here the default interrupt handlers.

tact1m4n3: I mean there is no need for a default_interrupts declaration as you can set things directly here

Grazfather: But I can't, since the interrupt is nrf specific, not cortex specific

tact1m4n3: Yep. I agree with you that we need default_interrupts in the hal. This refers to default_interrupts in the cpu module

core/src/cpus/cortex_m.zig
```diff
@@ -538,6 +538,88 @@ pub fn clrex() void {
...
+                ptr.* = new_value;
+                return null;
+            }
+            return current;
+        }
+    }
+
+    /// Atomic read-modify-write
+    pub fn rmw(comptime T: type, ptr: *T, comptime op: std.builtin.AtomicRmwOp, operand: T, comptime ordering: std.builtin.AtomicOrder) T {
+        if (comptime has_native_atomics) {
```
tact1m4n3: I think you can remove `comptime` here

port/nordic/nrf5x/src/hal.zig
```diff
@@ -8,7 +8,12 @@ pub const spim = @import("hal/spim.zig");
 pub const time = @import("hal/time.zig");
 pub const uart = @import("hal/uart.zig");
 pub const drivers = @import("hal/drivers.zig");
-// TODO: adc, timers, pwm, rng, rtc, interrupts, i2c, wdt, wifi, nfc, bt, zigbee
+// TODO: adc, timers, pwm, rng, rtc alarms, interrupts, wdt, wifi, nfc, bt, zigbee
+
+pub const default_interrupts: microzig.cpu.InterruptOptions = .{
+    // Required for timekeeping longer than 512 seconds
+    .RTC0 = microzig.interrupt.Handler{ .c = time.rtc_interrupt },
```
tact1m4n3: You can replace `microzig.interrupt.Handler{}` with `.`

tact1m4n3 (approved):

core/src/cpus/cortex_m.zig
tact1m4n3: in `if (comptime has_native_atomics) {`

PR #666, opened 2025-09-03
nrf52: Add basic clocks hal

Grazfather:


port/nordic/nrf5x/src/hal/clocks.zig
```diff
@@ -0,0 +1,156 @@
...
+                while (CLOCK.EVENTS_DONE == 0) {}
+            },
+            .nrf52 => {
+                CLOCK.TASKS_CAL.write_raw(1);
+                while (CLOCK.EVENTS_DONE.raw == 0) {}
+            },
+        }
+    }
+
+    pub fn set_source(source: Source) !void {
```
Grazfather: Could make this comptime so that the validation is done earlier.

Grazfather: I did this in 9364b43. It makes the function about 80% shorter. I am not sure if we realistically have a case where we'd want the value of Source to be dynamically configured, so I think that it's better with comptime

port/nordic/nrf5x/src/hal/clocks.zig
```diff
@@ -0,0 +1,154 @@
...
+
+    pub fn set_source(comptime source: Source) !void {
+        switch (version) {
+            .nrf51 => {
+                CLOCK.LFCLKSRC.write(.{
+                    .SRC = @enumFromInt(@intFromEnum(source)),
+                });
+            },
+            .nrf52 => {
+                try comptime source.validate();
```
tact1m4n3: Wouldn't it be better to print a compile error here? (we know all the variables at comptime)

Grazfather: Ah yes of course. Best not to have to percolate an error up. This is left behind from when it wasn't comptime.

mattnite (approved):
Once that comment from @tact1m4n3 is resolved this is looking great to me

PR #672, opened 2025-09-05
Aviron: Add debugging features, add basic SRAM mapping

Grazfather:
* Add a feature to detect infinite loop halts
* Initialize SP properly (though gcc-compiled binaries will set this for us)
* Map SRAM to 0x100. Needed for when low registers are modified via writes rather than `in`
* Detect when killed via IO, and return normally.
* Dump full state when the program halts, when trace is set.


sim/aviron/.gitignore
```diff
@@ -0,0 +1 @@
+**/*.elf
```
Grazfather: TODO: Remove and re-add (plus update) elf files

sim/aviron/src/lib/Cpu.zig
```diff
@@ -143,6 +159,44 @@ pub fn run(cpu: *Cpu, mileage: ?u64) RunError!RunResult {
             }
         }
     }
+    // Dump complete system state
```
Grazfather: TODO: Remove or hide behind a flag

sim/aviron/src/lib/Cpu.zig
```diff
@@ -172,31 +230,33 @@ fn push_code_loc(cpu: *Cpu, val: u24) void {
...
+    // With write-then-decrement PUSH, we push least significant byte first.
     if ((mask & 0x0000FF) != 0) {
         cpu.push(@truncate(pc >> 0));
     }
     if ((mask & 0x00FF00) != 0) {
         cpu.push(@truncate(pc >> 8));
     }
     if ((mask & 0xFF0000) != 0) {
-        cpu.push(@truncate(pc >> 16));
+        cpu.push(@truncate(pc >> 0));
```
Grazfather: fixme

sim/aviron/src/lib/Cpu.zig
```diff
@@ -141,6 +198,12 @@ pub fn run(cpu: *Cpu, mileage: ?u64) RunError!RunResult {
                     }
                 },
             }
+
+            // Check if the program requested exit via I/O
+            if (cpu.io.check_exit()) |exit_code| {
```
Grazfather, line 204: This is a big janky: IO has to store whether we need to exit, so that the CPU knows to stop running.

I do this so that we don't just kill the process when the inferior writes to 0. This way we can do cleanup, dump debug info, etc. Before this made debugging annoying since a proper exit would just skip all my debug prints.

sim/aviron/src/lib/io.zig
```diff
@@ -120,12 +122,18 @@ pub const RAM = struct {
 
             fn mem_read(ctx: ?*anyopaque, addr: Address) u8 {
                 const mem: *Self = @ptrCast(@alignCast(ctx.?));
-                return mem.data[addr];
+                // ATmega328P memory map: SRAM starts at 0x0100
+                // Map logical addresses 0x0100-0x08FF to array indices 0x0000-0x07FF
+                const sram_offset: Address = if (addr >= 0x100) addr - 0x100 else addr;
```
ikskuh, line 127: I would rather make this a configurable option, as it's part of a *Static* memory, and not a ATmega328P memory. This seems to be incredibly unexpected.

sim/aviron/src/lib/io.zig
```diff
@@ -120,12 +122,18 @@ pub const RAM = struct {
...
+                return mem.data[sram_offset];
             }
 
             fn mem_write(ctx: ?*anyopaque, addr: Address, value: u8) void {
                 const mem: *Self = @ptrCast(@alignCast(ctx.?));
-                mem.data[addr] = value;
+                // ATmega328P memory map: SRAM starts at 0x0100
+                // Map logical addresses 0x0100-0x08FF to array indices 0x0000-0x07FF
+                const sram_offset: Address = if (addr >= 0x100) addr - 0x100 else addr;
```
ikskuh, line 135: See comment in `mem_read`

sim/aviron/src/main.zig
```diff
@@ -143,13 +155,17 @@ const Cli = struct {
     mcu: MCU = .atmega328p,
     info: bool = false,
     format: FileFormat = .elf,
+    break_pc: ?u24 = null,
```
ikskuh, line 158: I would rename this to `breakpoint`, so `aviron --breakpoint=0x1234`.

Or `break-pc` for `aviron --break-pc=0x1234`

ikskuh (changes requested):
I'm not sure about the overall `0x100` thingy. I'd rather see this in the application `aviron` which currently hardcodes the ATmega328P, but i'm not sure why we should have that in the test suite (which should effectively run with *any* AVR cpu) and in the emulator core. This 0x100 offset will be different for most CPUs, and considering we're actually using a `attiny816` for testing

sim/aviron/src/lib/io.zig
Grazfather: Yes, I have another change that lets memory ranges to take bases, which is configurable

PR #678, opened 2025-09-18
Fix incorrect calculation for get_readable_len in CircularBuffer

### LINKED ISSUE #677: Mistake in CircularBuffer.get_readable_len() cause interger overflow
Within the circular buffer inplmentations in core\src\utilities.zig, there is a mistake in get_readable_len(). 
In the case the end position is before the start position in buffer the readbale length is incorrectly calcualted as `len - buffer.start - buffer.end`
This can cause interger overflow and panic the program.
I beleve the correct calcualtion should `len - buffer.start + buffer.end`
I will submit a PR tho it is litterly a one line change.

luke-thom:
Fixes #677

### DIFF
--- a/core/src/utilities.zig
+++ b/core/src/utilities.zig
@@ -546,7 +546,7 @@ pub fn CircularBuffer(comptime T: type, comptime len: usize) type {
             return if (buffer.start <= buffer.end)
                 buffer.end - buffer.start
             else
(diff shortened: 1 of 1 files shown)


tact1m4n3 (approved):
Thanks! Nice catch!

PR #681, opened 2025-09-21
STM32F1 Time Driver

RecursiveError:
RTC and GP-Timer time driver for the STM32F1xx

Bug fix: 
- HSE/128 is now configured correctly.

### DIFF
--- a/port/stmicro/stm32/src/hals/STM32F103/time.zig
+++ b/port/stmicro/stm32/src/hals/STM32F103/time.zig
@@ -1,0 +1,53 @@
+const microzig = @import("microzig");
+const rcc = @import("rcc.zig");
+const bkp = @import("backup.zig");
+const rtc = @import("rtc.zig");
+const time = microzig.drivers.time;
+
+pub fn init() void {
+    //Enable Backup domain clocks
+    rcc.enable_clock(.PWR);
+    rcc.enable_clock(.BKP);
+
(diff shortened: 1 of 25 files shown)


port/stmicro/stm32/src/hals/STM32F103/time.zig
```diff
@@ -0,0 +1,53 @@
...
+    });
+    // HSE/128 is a special case that requires DBP always disabled
+    if (microzig.chip.peripherals.RCC.BDCR.read().RTCSEL != .HSE) {
+        bkp.set_data_protection(true);
+        rcc.disable_clock(.PWR);
+        rcc.disable_clock(.BKP);
+    }
+}
+
+pub fn get_time_since_boot() time.Absolute {
```
RecursiveError: In Embassy, interrupts are commonly used to extend counter size by incrementing a variable on each event. This approach is mostly relevant for general-purpose timers.

For the RTC, however, this extension is unnecessary. On ST chips, the RTC is already implemented with two cascaded counters: the RTC clock (RTC_CLK) feeds the prescaler counter (TR_CLK), which on overflow increments the RTC counter (RTC_CNT). Together, they provide a 52-bit counter.

This driver does not take advantage of the full 52 bits, since even the RTC_CNT counting in seconds would take many years to overflow.

PR #683, opened 2025-09-24
Aviron features

Grazfather:
* Improve memory dump
* Add address translation/mapping to send reads/writes to correct device.
* Add base address to `Static` memory.


sim/aviron/src/lib/Cpu.zig
```diff
@@ -109,20 +101,82 @@ pub fn dump_system_state(cpu: *Cpu) void {
...
+    // Dump all registers
+    std.debug.print("\nREGISTERS:\n", .{});
+    for (cpu.regs, 0..) |reg, i| {
+        if (i % 8 == 0) std.debug.print("r{d:0>2}-r{d:0>2}: ", .{ i, @min(i + 7, 31) });
+        std.debug.print("{X:0>2} ", .{reg});
+        if ((i + 1) % 8 == 0) std.debug.print("\n", .{});
+    }
+
+    // Dump SRAM using absolute addresses based on configured base
+    const sram_base: u24 = cpu.sram.get_base();
```
Grazfather: This is the only use of `get_base`, which is accessed through the interface (a `RAM` object). We need the field in the `Static` struct (to know how to translate an address into an array offset), and so I needed to add this to the vtable to make it available here. I *could* also copy the base into the base struct when `memory()` is called, which would clean up the interface, but I am not sure what you think is nicer, so please weigh in.

sim/aviron/src/lib/Cpu.zig
```diff
@@ -109,20 +101,82 @@ pub fn dump_system_state(cpu: *Cpu) void {
...
+        var j: usize = 0;
+        while (j < row_len) : (j += 1) {
+            cur_row[j] = cpu.sram.read(sram_base + @as(u24, @intCast(i + j)));
         }
-        std.debug.print("{X:0>2} ", .{cpu.sram.read(@as(u24, @intCast(0x0100 + i)))});
-        if ((i + 1) % 16 == 0) {
-            std.debug.print("\n", .{});
+
+        // 'Don't show repeated lines to minimize the size of the dump
+        const same_as_prev = have_prev and prev_len == row_len and std.mem.eql(u8, prev_row[0..prev_len], cur_row[0..row_len]);
```
Grazfather: Maybe instead I should just check that the line is all zeroes?

sim/aviron/src/lib/Cpu.zig
```diff
@@ -109,20 +101,92 @@ pub fn dump_system_state(cpu: *Cpu) void {
...
+    var elided = false;
+
+    var i: usize = 0;
+    while (i < cpu.sram.size) : (i += @min(row_width, cpu.sram.size - i)) {
+        const row_len: usize = @min(row_width, cpu.sram.size - i);
+        var cur_row: [row_width]u8 = undefined;
+
+        var j: usize = 0;
+        while (j < row_len) : (j += 1) {
+            cur_row[j] = cpu.sram.read(sram_base + @as(u24, @intCast(i + j)));
```
Grazfather: probably doesn't matter, but to avoid copying, we could maybe add a method that returns a slice of the underlying memory, which we can hold a handle to.

sim/aviron/src/main.zig
```diff
@@ -96,10 +99,14 @@ pub fn main() !u8 {
                         &flash_storage.data;
 
                     const addr_masked: u24 = @intCast(phdr.p_paddr & 0x007F_FFFF);
+                    const target_addr: u24 = if (phdr.p_paddr >= 0x0080_0000)
+                        addr_masked - sram.base
```
Grazfather:
```
❯ avr-readelf -s test.elf

Symbol table '.symtab' contains 396 entries:
   Num:    Value  Size Type    Bind   Vis      Ndx Name
     0: 00000000     0 NOTYPE  LOCAL  DEFAULT  UND
     1: 00800100     0 SECTION LOCAL  DEFAULT    1 .data
```

.data is at 0x00800100, but it should start at 0 in the array, so we mask off the top 8, then adjust by the base (0x100)

sim/aviron/src/main.zig
```diff
@@ -143,13 +159,17 @@ const Cli = struct {
...
+    break_pc: ?u24 = null,
+    gas: ?u64 = null,
 
     pub const shorthands = .{
         .h = "help",
         .t = "trace",
         .m = "mcu",
         .I = "info",
         .f = "format",
+        .B = "break_pc",
```
Grazfather: could be lower case?

sim/aviron/src/main.zig
```diff
@@ -166,6 +186,8 @@ const Cli = struct {
             .mcu = "Selects the emulated MCU.",
             .info = "Prints information about the given MCUs memory.",
             .format = "Specify file format.",
+            .break_pc = "Break when PC reaches this address (hex or dec)",
```
Grazfather: You wanted to call this 'breakpoint'?

sim/aviron/src/main.zig
```diff
@@ -293,7 +324,9 @@ const IO = struct {
             .sp_h => write_masked(high_byte(&io.sp), mask, value),
             .sreg => write_masked(@ptrCast(io.sreg), mask, value),
 
-            _ => std.debug.panic("illegal i/o write to undefined register 0x{X:0>2} with value=0x{X:0>2}, mask=0x{X:0>2}", .{ addr, value, mask }),
+            _ => {
```
Grazfather: should probably restore the panic.

sim/aviron/src/lib/Cpu.zig
ikskuh: Ok for now, we gotta refactor all of that anyways for better device support

sim/aviron/src/lib/Cpu.zig
ikskuh: also fine for now

sim/aviron/linker.ld
```diff
@@ -10,6 +10,9 @@ SECTIONS
     {
         KEEP(*(.vectors))
 
+        /* Ensure _start comes first */
+        KEEP(*(.text._start))
```
ikskuh: This won't always hold true as `.text._start` will only be present with `-ffunction-sections` compile flag (which is on in default for Zig code, but not for C and definitly not for Asm)

sim/aviron/src/lib/io.zig
```diff
@@ -134,7 +177,8 @@ pub const RAM = struct {
 pub const EEPROM = RAM; // actually the same interface *shrug*
 
 pub const IO = struct {
-    pub const Address = u6;
+    // Some AVR families (e.g., XMEGA) expose extended I/O up to 0xFFF (12 bits).
+    pub const Address = u12;
```
ikskuh: Which instructions would use these? Afaik the 1024 byte I/O is only usable via the RAM interface?

sim/aviron/src/lib/io.zig
```diff
@@ -154,14 +198,31 @@ pub const IO = struct {
...
+    /// Returns null when the address is not mapped to I/O and should be served by SRAM.
+    pub fn translate_address(mem: IO, data_addr: u24) ?Address {
+        return mem.vtable.translateAddressFn(mem.ctx, data_addr);
+    }
+
     pub const VTable = struct {
         readFn: *const fn (ctx: ?*anyopaque, addr: Address) u8,
         writeFn: *const fn (ctx: ?*anyopaque, addr: Address, mask: u8, value: u8) void,
+        checkExitFn: *const fn (ctx: ?*anyopaque) ?u8,
+        translateAddressFn: *const fn (ctx: ?*anyopaque, data_addr: u24) ?Address,
```
ikskuh: We could add default implementations for these two, where both functions will return `null` by default

sim/aviron/src/main.zig
ikskuh: The logic here is flawed as two LOAD headers at `00800100` and `00800200` would both be offset, while the third LOAD header at `00800000` would not.

You have to basically go through `cpu.data_write` for this instead of hijacking the data in a conditional here.

With a proper mapped SRAM object instead of a single "Static" this would not be a problem.

I think we should introduce a real mapper object here already, as we're piling up workarounds on workarounds just to *not* have a mapper and introduce more complexity than necessary by that.

The branch would not be necessary if we could just do `ram.write(addr_masked, data)`

sim/aviron/src/main.zig
ikskuh: I'd go with either `aviron --break-at 0x1000` or `aviron --breakpoint 0x1000`

sim/aviron/src/main.zig
ikskuh: We should definitly crash/stop here in any case, and not silently ignoring it

sim/aviron/src/main.zig
Grazfather: I've written a mapper, but it only supports writing individual bytes. I could extend it to add a method to write a slice, but with the changes to this part (take a look at new commits) I think this is probably fine.

In your example case, though: How would there be a load at 0x00800000 if SRAM starts at 0x100?

sim/aviron/linker.ld
Grazfather: is `_start` not exported normally for gcc? If not, what can we add here as an alternative?

sim/aviron/src/lib/io.zig
Grazfather: It's not for io instructions, it's for MCUs with IO mapped and read/writable with normal load and stores.

sim/aviron/src/lib/io.zig
Grazfather: Added a `empty()` function to return an IO with dummy functions. Would you prefer that the vtable have default values so an empty version can be initialized with`.{}` instead? I find this more expressive but I don't mind either way.

sim/aviron/linker.ld
ikskuh: The point is that `_start` is a symbol name, but `.text._start` is a section name *only* emitted when `-ffunction-sections` is passed for the object in question being compiled with that flag. And that object is usually `crt0.o`, which i have no idea if it is compiled that way

sim/aviron/src/lib/io.zig
ikskuh: Thought so, so it's a workaround for not having the ability to map the I/O into a memory space ;)

sim/aviron/src/lib/io.zig
ikskuh: Yeah i would use a default function for `checkExitFn` and best just delete  `translateAddressFn`, as we can solve that better with a mapper object

sim/aviron/src/main.zig
ikskuh: > How would there be a load at 0x00800000 if SRAM starts at 0x100?

It would be a faulty object file not suitable for that system setup, but you could actually construct that and have the ELF loader initialize/perform I/O :laughing: 

> I've written a mapper, but it only supports writing individual bytes. 

Can you explain that? If the mapper object is just a `SRAM` interface, it should be no change required on the CPU side and the code here would not change at all

Grazfather: > Can you explain that? If the mapper object is just a SRAM interface, it should be no change required on the CPU side and the code here would not change at all

It delegates to SRAM or IO, but the interface is to read or write single bytes, so instead of using memcpy we'd have to loop over the slice and call write once for each byte

Grazfather: Just thought of another thing that we should figure out: Should we actually load SRAM at all? A 'real' binary will always have a stub that loads data from flash to sram, since that is how hardware would work. That means we don't need to check the entry point or load anything in the sram range. We just unconditionally copy everything in the flash range to flash and start at 0.

avr-gcc already makes elfs that work like this, but it seems that the zig code does not (e.g. the _write-chan.zig_ test), but of course, we can just make sure `_start` is at 0 and make that function load everything.

I think that I would rather go this way since it's closer to how the hardware behaves. What do you think?

sim/aviron/src/lib/io.zig
Grazfather: Yeah I suppose it is not needed with the `MemorySpace` now

sim/aviron/src/lib/mcu.zig
```diff
@@ -0,0 +1,172 @@
...
+    data_seg_buf[0] = .{ .at = cfg.io_window_base, .size = io_size, .backend = memory.Backend.fromIO(io_mem) };
+    data_seg_buf[1] = .{ .at = cfg.sram_base, .size = ram.size, .backend = memory.Backend.fromRAM(ram) };
+    const data_space = try memory.MemorySpace.init(alloc, data_seg_buf[0..]);
+
+    // IO space: IO addresses starting at 0
+    var io_seg_buf: [1]memory.Segment = undefined;
+    io_seg_buf[0] = .{ .at = 0, .size = io_size, .backend = memory.Backend.fromIO(io_mem) };
+    const io_space = try memory.MemorySpace.init(alloc, io_seg_buf[0..]);
+
+    // Program space: byte-addressable view over Flash
```
Grazfather: Flash accesses are a bit uglier now since I shoe-horned them into an interface that only does read8. I could add a read16 method to the vtable, and make them error out in the ram/io implementations maybe?

sim/aviron/src/main.zig
```diff
@@ -4,30 +4,18 @@ const aviron = @import("aviron");
...
-        return if (cli.options.help) @as(u8, 0) else 1;
-    }
-
-    // TODO: Add support for more MCUs!
-    std.debug.assert(cli.options.mcu == .atmega328p);
-    const mcu_config = aviron.mcu.atmega328p;
-
+fn run_with_mcu(
+    allocator: std.mem.Allocator,
+    comptime mcu_config: anytype,
```
Grazfather, line 9: This approach bloats the binary size but avoids needing a dynamically sized bus

PR #690, opened 2025-09-27
uf2: Refactor and `Archive.read_from`

tact1m4n3:
- Use `takeStruct` and `writeStruct` for blocks
- Use `ArrayHashMap` for keys
- Add `Archive.read_from` for reading an archive from a reader
- Update `src/example.zig`
- Remove the need for submodules when generating `uf2.FamilyId`: Here I am not sure how to remove the uf2 submodule in the deps directory, but it is basically redundant now


tools/uf2/README.md
```diff
@@ -21,7 +25,7 @@ pub fn build(b: *Build) void {
 
 ```
 
-# Manually Executing elf2uf2
+Or execute elf2uf2 manually:
```
Grazfather: nit: This is a subheading of `...in build.zig` which doesn't make sense. I'd either drop the header for that line, or make this one an h2 as well.

tools/uf2/src/gen.zig
```diff
@@ -11,24 +11,19 @@ pub fn main() !void {
     const allocator = arena.allocator();
     defer arena.deinit();
 
-    const json_file = try std.fs.cwd().openFile("deps/uf2/utils/uf2families.json", .{});
-    defer json_file.close();
+    const uf2families = @embedFile("uf2families");
```
Grazfather: Could we generated family_id.zig at comptime from this dep, rather than have a copy of what is probably similar data?

tools/uf2/src/example.zig
```diff
@@ -2,11 +2,15 @@ const std = @import("std");
 const uf2 = @import("uf2.zig");
 
 pub fn main() !void {
-    const args = try std.process.argsAlloc(std.heap.page_allocator);
-    defer std.process.argsFree(std.heap.page_allocator, args);
+    var debug_allocator: std.heap.DebugAllocator(.{}) = .init;
```
Grazfather, line 5: why did you switch the allocator?

tools/uf2/src/gen.zig
Grazfather: Or did you not mean to commit family_id.zig?

tools/uf2/src/uf2.zig
```diff
@@ -30,15 +22,65 @@ pub const Archive = struct {
...
+            block.total_blocks = @as(u32, @intCast(self.blocks.items.len));
+            try writer.writeStruct(block.*, .little);
+        }
+        try writer.flush();
+    }
+
+    pub const ReadFromOptions = struct {
+        /// If true, checks if any new block has a family_id already present in
+        /// the archive.
+        check_family_id_collission: bool = true,
```
Grazfather: typo collision

tools/uf2/src/gen.zig
tact1m4n3: > Could we generated family_id.zig at comptime from this dep, rather than have a copy of what is probably similar data?

This is a great idea, I even implemented it... But then I realized that you need to access the enum from the build script :)

tools/uf2/src/example.zig
tact1m4n3: Well it's an example and I thought that using page_allocator directly isn't really recommended :)

Grazfather: That's fair! I was just curious.

tact1m4n3:
I made the microsoft_uf2 dep lazy now so no need to fetch it every time and I also updated the from_elf function in the style of esp_image (I also used it in the main microzig build.zig).

Grazfather (approved):

PR #697, opened 2025-09-30
CircularBuffer fn get_readable_len did not account for full case

Kytezign:
I've never tried to do a pull request before please let me know if there is a better or more correct way.  

When buffer.start == buffer.end it could be full or empty.


mattnite (approved):
Thanks for that catch!

PR #698, opened 2025-10-01
Update cdc.zig with line state tracking

Kytezign:
Line state can be useful when waiting for a terminal to connect before trying to write/read.  This adds a line state variable to track that.


mattnite (approved):
Thank you again!

Issue #684, opened 2025-09-24
RP2040 CDC usb example fails


Kytezign:
I've followed the getting started instructions then replaced the main.zig with the usb_cdc example here: 
https://github.com/ZigEmbeddedGroup/microzig/blob/main/examples/raspberrypi/rp2xxx/src/usb_cdc.zig 

using a normal Raspberry Pi Pico.

Watching the UART logger I see this. I'm new to microzig (and relatively new to zig) but am happy to do more debug work for this.

Also curious if I should/could try the new USB implementation being worked on since maybe this has been resolved there. 

# USB output after connecting (clearly truncated even when working)
```
 CDC to your device: 2
 device: 3
Your message to me was: test 0
 device: 4
 device: 5
Your message to me was: test 1
```

# UART output
```
================ STARTING NEW LOGGER ================
[0.506618] info: cdc test: 1
[1.006610] info: cdc test: 2
[1.506611] info: cdc test: 3
[2.006613] info: cdc test: 4
[2.506614] info: cdc test: 5
[3.006615] info: cdc test: 6
[3.009325] error: panic: integer overflow
[3.013256] error: panic: unhandled interrupt
```

Kytezign:
Maybe a duplicate of this:
https://github.com/ZigEmbeddedGroup/microzig/issues/677 (I'm not at my computer to check right now)

If the PR is closed, does that mean it's in the release 
https://github.com/ZigEmbeddedGroup/microzig/releases/download/0.15.0/microzig.tar.gz

Kytezign:
I've confirmed the above.  But now I'm seeing a different issue.  For the very very long text being sent, I'm missing the first 64 bytes (see device: # line below).  Assuming something related to a buffer somewhere?


```
 CDC to your device: 2
 device: 3
Your message to me was: test 0
 device: 4
 device: 5
Your message to me was: test 1
 device: 6
 device: 7
Your message to me was: test 2
 device: 8
 device: 9
Your message to me was: test 3
 device: 10
 device: 11
Your message to me was: test 4
 device: 12
 device: 13
Your message to me was: test 5
```

Kytezign:
I do not yet have a satisfying explanation but after changing the cdc.zig write function to this it's working for me.

Really curious at this point, is this just a me problem somehow? 
```        
pub fn write(self: *@This(), data: []const u8) []const u8 {
            var write_count: usize = 0;
            const w_len = self.tx.get_writable_len() - 1;

[...]

PR #699, opened 2025-10-01
Ensure RP2040 tx buffer control is respected

Kytezign:
See also "4.1.2.7.1. Concurrent access" in the RP2040 datasheet.


Kytezign:
https://github.com/ZigEmbeddedGroup/microzig/issues/684

mattnite (approved):

Kytezign:
@mattnite, 
Is there something I can/should do to fix this formatting check failure?

arkadiuszwojcik:
> @mattnite, Is there something I can/should do to fix this formatting check failure?

@Kytezign  Did you try: `zig fmt <files>` command?

Kytezign:
> @Kytezign Did you try: `zig fmt <files>` command?

No, I'm not great with git.  I just made the edit in the github interface - which doesn't have that option. I'll try to figure it out.

Issue #701, opened 2025-10-04
proposal: rework timeouts, StreamDevices and DatagramDevices


RecursiveError:
Summary:
- Change the default timeout behavior in protocols that behave like streams

- Analyze ambiguity between the current Datagram and StreamDevices interfaces

- Extra case: Analyze the use case of drivers exposing generic interfaces

## How a timeout can turn a stream into a datagram

A **stream** represents a communication channel with no predefined size.  
Because of this, a timeout **should not be treated as an error by default** — it’s expected that a device might not send an entire buffer at once, or that a receiver might not always have enough incoming data to fill its read buffer.

However, the way we currently handle timeouts in most UART HALs treats a timeout as a **failed transmission of a fixed-size buffer in single transaction**, effectively turning the stream into a **datagram-like** operation.

This raises an interesting question:  

> Is UART really a *StreamDevice*?

Streams often use UART as their canonical example, since UART is a very common way to perform raw communication between two controllers.  
But thinking more deeply — what prevents a *fixed-size protocol* from using UART?  
**Nothing.** It’s perfectly valid (even if less common in some designs) to implement length-defined packet protocols over UART.

That means `StreamDevice` has an **edge case** where it behaves like a datagram.  
But then, what are the edge cases for **Datagram**?

---

## Is a datagram also a kind of stream?

By analogy, datagram interfaces are typically represented by **I2C** or **SPI**, which are usually used for fixed-size packet transfers.  
And that’s true — *to some extent.*

These buses are very common in simple modules where communication rules are rigidly defined, and a write timeout or failed full transfer is indeed an error.

But both I2C and SPI can also be used in a **stream-like context**, especially SPI, which is often used for raw controller-to-controller communication (similar to USART).

I2C as a special case:

In a host → device transaction, the device can send a NACK to indicate it no longer wants to receive data — this is valid behavior.

In a device → host transaction, the specification requires the device to send 0xFF if it has no more data to transmit; This means that a reader with an arbitrary buffer length does not generate an error, which is problematic for streams: here, 0xFF is ambiguous, as it could be either valid data or a placeholder.

Despite this,  **REPEATED START**, makes it much closer to stream-like behavior over a datagram, And there are also cases of 
some I2C modules — especially those that also support SPI — to typically provide a burst read/write function, allowing the host to read or write N bytes sequentially (as long as a certain limit is not exceeded), edge-case of stream-like behavior on I2C. 

[...]

PR #702, opened 2025-10-04
update Stream_Device IO and STM32 uart

RecursiveError:
- Update StreamDevice to new IO API

The modification in the STM32 UART is only a test for timeout behavior on buses abstracted as **StreamDevices**.
It’s a **necessary change** for the new I/O system to work correctly, but it **might not be desirable as a default** for the protocol’s HAL implementation.
More information about the motivation behind this change can be found in the following issue: #701

### DIFF
--- a/drivers/base/Stream_Device.zig
+++ b/drivers/base/Stream_Device.zig
@@ -74,8 +74,8 @@ pub const VTable = struct {
     readv_fn: ?*const fn (*anyopaque, datagram: []const []u8) ReadError!usize,
 };
 
-///microzig already has this function in SliceVector, but no other base driver include microzig as a dependency
-///so we duplicate it here for now
+/// microzig already has this function in SliceVector, but no other base driver include microzig as a dependency
+/// so we duplicate it here for now
 fn byte_sum(data: []const []const u8) usize {
     var sum: usize = 0;
     for (data) |bytes| {
@@ -105,8 +105,8 @@ pub const Writer = struct {
             return error.WriteFailed;
         };
 
-        //TODO: maybe check if ret is 0 across multiple calls to detect broken stream?
-        //check if we wrote everything we wanted before splatting
+        // NOTE: maybe check if ret is 0 across multiple calls to detect broken stream?
+        // check if we wrote everything we wanted before splatting
         if (ret != total_size) {
             return ret;
         }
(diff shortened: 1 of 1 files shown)


drivers/base/Stream_Device.zig
```diff
@@ -84,6 +74,104 @@ pub const VTable = struct {
...
+};
+
+pub const Reader = struct {
+    dev: Stream_Device,
+    interface: std.Io.Reader,
+    err: ?ReadError = null,
+
+    fn stream(r: *std.Io.Reader, w: *std.Io.Writer, limit: std.Io.Limit) std.Io.Reader.StreamError!usize {
+        const rd: *Reader = @alignCast(@fieldParentPtr("interface", r));
+        const max: usize = @intFromEnum(limit);
```
Grazfather: Can probably just use `@intFromEnum` once below

drivers/base/Stream_Device.zig
```diff
@@ -84,6 +74,104 @@ pub const VTable = struct {
...
+        var aux_buffer: [256]u8 = undefined; //mandatory minimum buffer
```
Grazfather: Can you explain this better?

drivers/base/Stream_Device.zig
```diff
@@ -84,6 +74,104 @@ pub const VTable = struct {
...
+            return error.WriteFailed;
+        };
+        _ = w.consume(n);
+
+        ret = wt.dev.writev(data) catch |err| {
+            wt.err = err;
+            return error.WriteFailed;
+        };
+
+        //TODO: maybe check if ret is 0 across multiple calls to detect broken stream?
```
Grazfather: Nit on style across basically all your PRs: Please put a space after the `//` in your comments.

Grazfather (commented):
Please update the description / title of this PR so it explains it brief what it does.

For those looking at git history in their local tree, they shouldn't have to open github and go to the PR, and then to the issue to understand your intent.

drivers/base/Stream_Device.zig
RecursiveError: Oh, don't worry about it, it's just a basic "minimum mandatory buffer" style for testing, I'll change it to modify the internal read buffer when the PR is ready

PR #703, opened 2025-10-05
Aviron: Add breakpoint, set gas, detect infinite loop

Grazfather:


sim/aviron/src/main.zig
```diff
@@ -176,6 +187,10 @@ const IO = struct {
     sp: u16,
     sreg: *aviron.Cpu.SREG,
 
+    // Exit status tracking
+    exit_requested: bool = false,
+    exit_code: u8 = 0,
```
ikskuh: `exit_code: ?u8 = null,` stores both values in one element

ikskuh (approved):
Minor nitpick, but feel free to hit merge anyways

sim/aviron/src/main.zig
Grazfather: good call

PR #704, opened 2025-10-05
Unify interface, add memory mapping

Grazfather:
* Unify interface to read/writes in IO and RAM
* Add a memory space that also conforms to this interface
* Store these interface objects directly in the CPU
* _mcu.zig_ that contains configs for different MCUs
* Add `build_spaces` helper that uses an mcu config to create data, io, and eeprom data spaces

### DIFF
--- a/sim/aviron/src/lib/Cpu.zig
+++ b/sim/aviron/src/lib/Cpu.zig
@@ -261,9 +261,9 @@ fn write_wide_reg(cpu: *Cpu, reg: WideReg, value: u24, comptime mode: IndexRegWr
     cpu.regs[reg.base() + 1] = parts[1];
     if (mode == .ramp) {
         switch (reg) {
-            .x => if (cpu.sio.ramp_x) |ramp| cpu.io.write(ramp, parts[2]),
-            .y => if (cpu.sio.ramp_y) |ramp| cpu.io.write(ramp, parts[2]),
-            .z => if (cpu.sio.ramp_z) |ramp| cpu.io.write(ramp, parts[2]),
+            .x => if (cpu.sio.ramp_x) |ramp| cpu.io.write(ramp, 0xFF, parts[2]),
+            .y => if (cpu.sio.ramp_y) |ramp| cpu.io.write(ramp, 0xFF, parts[2]),
+            .z => if (cpu.sio.ramp_z) |ramp| cpu.io.write(ramp, 0xFF, parts[2]),
         }
     }
 }
@@ -771,7 +771,7 @@ const instructions = struct {
     /// Stores data from register Rr in the Register File to I/O Space (Ports, Timers, Configuration Registers, etc.).
     inline fn out(cpu: *Cpu, info: isa.opinfo.a6r5) void {
         // I/O(A) ← Rr
-        cpu.io.write(info.a, cpu.regs[info.r.num()]);
+        cpu.io.write(info.a, 0xFF, cpu.regs[info.r.num()]);
     }
 
     /// IN - Load an I/O Location to Register
@@ -786,14 +786,14 @@ const instructions = struct {
     /// addresses 0-31.
     inline fn cbi(cpu: *Cpu, info: isa.opinfo.a5b3) void {
         // I/O(A,b) ← 0
-        cpu.io.write_masked(info.a, info.b.mask(), 0x00);
+        cpu.io.write(info.a, info.b.mask(), 0x00);
     }
 
     /// SBI – Set Bit in I/O Register
     /// Sets a specified bit in an I/O Register. This instruction operates on the lower 32 I/O Registers – addresses 0-31.
     inline fn sbi(cpu: *Cpu, info: isa.opinfo.a5b3) void {
         // I/O(A,b) ← 1
-        cpu.io.write_masked(info.a, info.b.mask(), 0xFF);
+        cpu.io.write(info.a, info.b.mask(), 0xFF);
     }
 
     // Branching:
@@ -1662,8 +1662,8 @@ fn get_sp(cpu: *Cpu) u16 {
 fn set_sp(cpu: *Cpu, value: u16) void {
     const lo: u8 = @truncate(value >> 0);
     const hi: u8 = @truncate(value >> 8);
-    cpu.io.write(cpu.sio.sp_l, lo);
-    cpu.io.write(cpu.sio.sp_h, hi);
+    cpu.io.write(cpu.sio.sp_l, 0xFF, lo);
+    cpu.io.write(cpu.sio.sp_h, 0xFF, hi);
 }
 
 fn compose24(hi: u8, mid: u8, lo: u8) u24 {
--- a/sim/aviron/src/lib/bus.zig
+++ b/sim/aviron/src/lib/bus.zig
@@ -122,11 +122,42 @@ pub const BusConfig = struct {
     address_type: type,
 };
 
+/// IO Bus interface with masked writes as the primary write method
+pub fn IOBusInterface(comptime address_type: type) type {
+    return struct {
+        const Self = @This();
+        pub const Address = address_type;
+
+        ctx: *anyopaque,
+        vtable: *const VTable,
+
+        pub const VTable = struct {
+            read: *const fn (ctx: *anyopaque, addr: Address) u8,
+            write: *const fn (ctx: *anyopaque, addr: Address, mask: u8, v: u8) void,
+            check_exit: ?*const fn (ctx: *anyopaque) ?u8 = null,
+        };
+
+        pub fn read(self: *const Self, addr: Address) u8 {
+            return self.vtable.read(self.ctx, addr);
+        }
+
+        /// Write with mask - this is the primary write interface for IO
+        /// `mask` determines which bits of `value` are written.
+        pub fn write(self: *const Self, addr: Address, mask: u8, v: u8) void {
+            self.vtable.write(self.ctx, addr, mask, v);
+        }
+
+        pub fn check_exit(self: *const Self) ?u8 {
+            if (self.vtable.check_exit) |f| return f(self.ctx) else return null;
+        }
+    };
+}
+
 /// Standard data bus using 24-bit addressing (for AVR data space)
 pub const DataBus = Bus(.{ .address_type = u24 });
 
 /// Standard IO bus using 12-bit addressing (for AVR IO space)
-pub const IOBus = Bus(.{ .address_type = IO.Address });
+pub const IOBus = IOBusInterface(IO.Address);
 
 /// Unified byte-addressable bus interface used by MemoryMapping for RAM and IO
 pub fn Bus(comptime config: BusConfig) type {
@@ -140,7 +171,6 @@ pub fn Bus(comptime config: BusConfig) type {
         pub const VTable = struct {
             read: *const fn (ctx: *anyopaque, addr: Address) u8,
             write: *const fn (ctx: *anyopaque, addr: Address, v: u8) void,
-            write_masked: *const fn (ctx: *anyopaque, addr: Address, mask: u8, v: u8) void,
             check_exit: ?*const fn (ctx: *anyopaque) ?u8 = null,
         };
 
@@ -150,10 +180,6 @@ pub fn Bus(comptime config: BusConfig) type {
 
         pub fn write(self: *const Self, addr: Address, v: u8) void {
             self.vtable.write(self.ctx, addr, v);
-        }
-
-        pub fn write_masked(self: *const Self, addr: Address, mask: u8, v: u8) void {
-            self.vtable.write_masked(self.ctx, addr, mask, v);
         }
 
         pub fn check_exit(self: *const Self) ?u8 {
@@ -182,7 +208,6 @@ pub fn FixedSizedMemory(comptime size: comptime_int, comptime bus_config: ?BusCo
         pub const bus_vtable = BusType.VTable{
             .read = read,
             .write = write,
-            .write_masked = write_masked,
             .check_exit = null,
         };
 
@@ -196,13 +221,6 @@ pub fn FixedSizedMemory(comptime size: comptime_int, comptime bus_config: ?BusCo
             const mem: *Self = @ptrCast(@alignCast(ctx));
             std.debug.assert(addr < size);
             mem.data[addr] = value;
-        }
-
-        fn write_masked(ctx: *anyopaque, addr: AddressType, mask: u8, value: u8) void {
-            const mem: *Self = @ptrCast(@alignCast(ctx));
-            std.debug.assert(addr < size);
-            const old = mem.data[addr];
-            mem.data[addr] = (old & ~mask) | (value & mask);
         }
     };
 }
@@ -273,14 +291,6 @@ pub fn MemoryMapping(comptime BusType: type) type {
             return;
         }
 
-        pub fn write_masked(self: *const Self, addr: BusType.Address, mask: u8, v: u8) AccessError!void {
-            const seg = self.find(addr) orelse return error.Unmapped;
-            const idx = addr - seg.at;
-            if (idx >= seg.size) return error.OutOfRange;
-            seg.backend.write_masked(idx, mask, v);
-            return;
-        }
-
         fn find(self: *const Self, addr: BusType.Address) ?*const Segment {
             // Linear scan is fine initially; segments are sorted.
             for (self.segments) |*s| {
@@ -301,7 +311,6 @@ pub fn MemoryMapping(comptime BusType: type) type {
         const bus_vtable = BusType.VTable{
             .read = bus_read,
             .write = bus_write,
-            .write_masked = bus_write_masked,
             .check_exit = bus_check_exit,
         };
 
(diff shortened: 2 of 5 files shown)


sim/aviron/build.zig
```diff
@@ -135,6 +135,17 @@ fn add_test_suite(
...
+    const memory_tests = b.addTest(.{
+        .root_module = b.createModule(.{
+            .root_source_file = b.path("src/lib/memory.zig"),
+            .target = host_target,
+            .optimize = optimize,
+        }),
+        .use_llvm = true,
+    });
+    test_step.dependOn(&b.addRunArtifact(memory_tests).step);
+
```
ikskuh: Why do you add a separate executable? You can just pull in `memory.zig` in `main.zig`:
```zig
test {
    _ = @import("lib/memory.zig");
}
```
This will compile much much faster than using a separate unit test per file and won't make problems in the future

sim/aviron/src/lib/Cpu.zig
```diff
@@ -1,11 +1,14 @@
 const std = @import("std");
 const isa = @import("decoder.zig");
 const io_mod = @import("io.zig");
+const builtin = @import("builtin");
+const memory = @import("memory.zig");
```
ikskuh: Can be deleted

sim/aviron/src/lib/io.zig
```diff
@@ -191,3 +161,32 @@ pub const IO = struct {
         return null;
     }
 };
+
+// Unified byte-addressable device interface used by MemorySpace for RAM and IO
+pub const Device = struct {
```
ikskuh: I would rename `Device` to `Bus` as it's not a "device" at all, but an abstraction over a memory bus

sim/aviron/src/lib/io.zig
```diff
@@ -191,3 +161,32 @@ pub const IO = struct {
...
+    ctx: *anyopaque,
+    vtable: *const VTable,
+
+    pub const VTable = struct {
+        read8: *const fn (ctx: *anyopaque, idx: usize) u8,
+        write8: *const fn (ctx: *anyopaque, idx: usize, v: u8) void,
+        write_masked: *const fn (ctx: *anyopaque, idx: usize, mask: u8, v: u8) void,
```
ikskuh: `write_masked` can have a default impl which does a read-modify-write cycle by default. 
If an implementation requires true atomics, they can implement it manually

ikskuh: I would use `read` and `write` instead of `read8` and `write8`. Will also require less changes on the whole codebase and doesn't convey any meaningful information as all `Bus` devices will have 8 bit access only anyways

ikskuh: I wouldn't use `idx: usize` here but `idx: u32` or `idx: u24`.

We're implementing a simulator here and the address size is fixed and not system dependent. The max. value of `usize` depends on the host system, but in this case we must depend on the maximum value available on `AVR`. Which is `u24`.

sim/aviron/src/lib/io.zig
```diff
@@ -60,72 +61,40 @@ pub const Flash = struct {
...
-        return 0;
-    }
-
-    fn empty_write(ctx: ?*anyopaque, addr: u16, value: u8) void {
-        _ = value;
-        _ = addr;
-        _ = ctx;
-    }
-
     pub fn Static(comptime size: comptime_int) type {
```
ikskuh: Refactor `RAM.Static` into `FixedSizedMemory` and `RAM.Dynamic` into `FlexibleMemory` and remove the now unnecessary `RAM` type

sim/aviron/src/lib/io.zig
```diff
@@ -134,7 +103,8 @@ pub const RAM = struct {
 pub const EEPROM = RAM; // actually the same interface *shrug*
 
 pub const IO = struct {
-    pub const Address = u6;
+    // Some AVR families (e.g., XMEGA) expose extended I/O up to 0xFFF (12 bits).
+    pub const Address = u12;
```
ikskuh: `IO` should use `u6` here as it's the size of an IO address still.

The XMEGA 12 bit thingy isn't on the IO bus, but on the data bus (which has 24 bits)

sim/aviron/src/lib/Cpu.zig
ikskuh, line 1: Please check if any of these `@intCast` is necessary when changing address size from `usize` to `u24` (or fitting)

sim/aviron/src/lib/mcu.zig
```diff
@@ -0,0 +1,200 @@
...
+};
+
+/// Build memory spaces (data, io, eeprom) for the given MCU configuration.
+pub fn build_spaces(
+    alloc: std.mem.Allocator,
+    cfg: Config,
+    sram_dev: io_mod.Device,
+    io_dev: io_mod.Device,
+    eeprom_dev: io_mod.Device,
+) !Spaces {
```
ikskuh, line 55: There is no need to pass `sram_dev` and `eeprom_dev` here as they can be allocated from `cfg` directly with a flexibly sized memory.

Otherwise it's just more work for the programmer and a source of bugs when they don't match `cfg`.

sim/aviron/src/lib/memory.zig
```diff
@@ -0,0 +1,194 @@
...
+    /// Base address within the enclosing memory space.
+    at: usize,
+    /// Size in bytes of the mapped range.
+    size: usize,
+    /// Backend handling the mapped range starting at index 0.
+    backend: Backend,
+};
+
+/// A logical memory space composed of non-overlapping segments.
+pub const MemorySpace = struct {
```
ikskuh: I would call this `MemoryMapping` this is what it does:
It maps certain parts into a memory bus

sim/aviron/src/lib/memory.zig
```diff
@@ -0,0 +1,194 @@
...
+/// Use the unified device interface
+pub const Backend = io.Device;
+const Device = io.Device;
+
+/// A mapping entry within a MemorySpace.
+pub const Segment = struct {
+    /// Base address within the enclosing memory space.
+    at: usize,
+    /// Size in bytes of the mapped range.
+    size: usize,
```
ikskuh: As above: `usize` is for host memory sizes, not for AVR memory sizes. Imagine someone using a 16 or 20 bit host, but aviron would expect to fit 24 bits of memory here.

sim/aviron/src/lib/memory.zig
ikskuh, line 1: We should merge `io.zig` and `memory.zig` into either `io.zig` or `memory.zig`, but having both is imho more confusing than helping.

I think `bus.zig` might also be a good name

sim/aviron/src/main.zig
```diff
@@ -21,41 +21,51 @@ pub fn main() !u8 {
...
+            .e_ind = mcu_config.special_io.e_ind,
 
-            .sp_l = @intFromEnum(IO.Register.sp_l),
-            .sp_h = @intFromEnum(IO.Register.sp_h),
+            .sp_l = mcu_config.special_io.sp_l,
+            .sp_h = mcu_config.special_io.sp_h,
 
-            .sreg = @intFromEnum(IO.Register.sreg),
+            .sreg = mcu_config.special_io.sreg,
         },
```
ikskuh: this should be `.sio = mcu_config.special_io` and having both parts use the same type

sim/aviron/src/main.zig
```diff
@@ -85,21 +95,33 @@ pub fn main() !u8 {
...
+                    // AVR uses 0x800000 flag in vaddr to indicate data memory
+                    var dest_mem: []u8 = if (phdr.p_vaddr >= 0x0080_0000)
                         &sram.data
                     else
                         &flash_storage.data;
 
-                    const addr_masked: u24 = @intCast(phdr.p_paddr & 0x007F_FFFF);
+                    const addr_masked: u24 = @intCast(phdr.p_vaddr & 0x007F_FFFF);
+                    const target_addr: u24 = if (phdr.p_vaddr >= 0x0080_0000)
+                        addr_masked - @as(u24, mcu_config.sram_base)
```
ikskuh: The subtraction of `mcu_config.sram_base` is still wrong here, as `0x0080_0000` is the start of the *data* bus, not the *sram*. This would always load the memory.

I would fully go through the bus map here and load the data with `databus.write()` byte by byte, only using shortcuts if always applicable.

ikskuh (changes requested):

sim/aviron/src/lib/io.zig
Grazfather: I kept Dynamic out of this PR, but I will change it to Flexible in a followup, and I will remove RAM for FixedSizedMemory

sim/aviron/src/lib/Cpu.zig
Grazfather: Done

sim/aviron/src/lib/io.zig
Grazfather: Right, but if the memory interface has to cast down to u6 it won't be able to address all the only on XMEGA, even if using the data space accessor.

sim/aviron/src/lib/io.zig
Grazfather: Done

sim/aviron/src/lib/mcu.zig
Grazfather: True, but this function lets the device to be fixed size or flexible, deferring to the caller to clean up the memory and avoiding runtime allocations when we can.

I propose I clean it up in the other PR, where I will add the dynamic support

sim/aviron/src/lib/memory.zig
Grazfather: Done

sim/aviron/src/lib/memory.zig
Grazfather: Done

sim/aviron/build.zig
Grazfather: When I tried this it complained that the file was already imported in aviron.zig

sim/aviron/src/lib/Cpu.zig
Grazfather: Done

sim/aviron/src/lib/io.zig
Grazfather: Done

sim/aviron/src/lib/memory.zig
Grazfather: Done

sim/aviron/src/main.zig
Grazfather: Fixed

sim/aviron/src/lib/bus.zig
```diff
@@ -0,0 +1,377 @@
...
+    ctx: ?*anyopaque,
+    vtable: *const VTable,
+
+    /// Size of the flash memory in bytes (word count). Is always 2-aligned.
+    size: Address,
+
+    pub fn read(mem: Flash, addr: Address) u16 {
+        std.debug.assert(addr < mem.size);
+        return mem.vtable.readFn(mem.ctx, addr);
+    }
```
ikskuh, line 16: We could add a `usage_hint: UsageHint` here where `UsageHint = enum { code, data}`.

This would allow us logging memory access patterns in traces, which is sometimes incredibly useful to debug hard to find bugs or optimize algorithms to go as fast as possible

sim/aviron/src/lib/bus.zig
```diff
@@ -0,0 +1,377 @@
...
+                std.debug.assert(addr < @as(Address, @intCast(@divExact(size, 2))));
+                return std.mem.bytesAsSlice(u16, &mem.data)[addr];
+            }
+        };
+    }
+};
+
+pub const IO = struct {
+    // Some AVR families (e.g., XMEGA) expose extended I/O up to 0xFFF (12 bits).
+    pub const Address = u12;
```
ikskuh, line 63: I would still use `u6` here as that's the width of the `IO` bus. If we actually need the "extended I/O", it would go through the `Bus` class anyways, so we have to expose two interfaces (and it may be interesting for logging to see how I/O is performed)

sim/aviron/src/lib/bus.zig
```diff
@@ -0,0 +1,377 @@
...
+
+    fn empty_check_exit(ctx: ?*anyopaque) ?u8 {
+        _ = ctx;
+        return null;
+    }
+};
+
+// Unified byte-addressable bus interface used by MemoryMapping for RAM and IO
+pub const Bus = struct {
+    pub const Address = u24;
```
ikskuh: I guess we should rename this into `Data`, as `bus.Data` reads really fine

Otherwise we could also use `Flash_Bus`, `IO_Bus` and `Data_Bus`

sim/aviron/src/lib/bus.zig
```diff
@@ -0,0 +1,377 @@
...
+        return self.vtable.read(self.ctx, addr);
+    }
+
+    pub fn write(self: *const Bus, addr: Address, v: u8) void {
+        self.vtable.write(self.ctx, addr, v);
+    }
+
+    pub fn write_masked(self: *const Bus, addr: Address, mask: u8, v: u8) void {
+        self.vtable.write_masked(self.ctx, addr, mask, v);
+    }
```
ikskuh: A data bus member doesn't need this function, as we can't have masked writes on the data bus, right?

sim/aviron/src/lib/bus.zig
```diff
@@ -0,0 +1,377 @@
...
+        const self: *const Self = @ptrCast(@alignCast(ctx));
+        // Check all segments for exit condition
+        for (self.segments) |*seg| {
+            if (seg.backend.check_exit()) |code| {
+                return code;
+            }
+        }
+        return null;
+    }
+};
```
ikskuh: We can make the `MemoryMapping` generic over a bus type, this way it keeps the right properties:

```
pub fn MemoryMapping(comptime BusType: type) type {
    return struct {
        pub const Segment = struct {
            …
        };
        
        …
    };
}
```

This way we would keep type confusion at a minimum while also keeping the amount of implementation at a minimum.

We can even safe-guard the implementation by only allowing `Data_Bus`, `Flash_Bus` and `IO_Bus` as the parameter (or we could pass an `enum {flash,data,io}`)

sim/aviron/src/lib/Cpu.zig
```diff
@@ -57,9 +56,9 @@ code_model: CodeModel,
 instruction_set: InstructionSet,
 sio: SpecialIoRegisters,
 flash: Flash,
-sram: RAM,
-eeprom: EEPROM,
-io: IO,
+data: Bus,
+io: Bus,
+eeprom: Bus,
```
ikskuh: I guess we can remove `eeprom` from the cpu busses, as iir we don't have any instructions operating on the EEPROM but it's always exposed through `.data` or `.io`

sim/aviron/src/main.zig
```diff
@@ -21,42 +21,41 @@ pub fn main() !u8 {
...
     // TODO: Add support for more MCUs!
     std.debug.assert(cli.options.mcu == .atmega328p);
+    const mcu_config = aviron.mcu.atmega328p;
 
-    var flash_storage = aviron.Flash.Static(32768){};
-    var sram = aviron.RAM.Static(2048){};
-    var eeprom = aviron.EEPROM.Static(1024){};
+    var flash_storage = aviron.Flash.Static(mcu_config.flash_size){};
+    var sram = aviron.FixedSizedMemory(mcu_config.sram_size){};
+    var eeprom = aviron.FixedSizedMemory(mcu_config.eeprom_size){};
```
ikskuh: We could make `FixedSizedMemory` also expose a `.flash()` interface where the function call invokes a `@compileError` if the size isn't divisible by 2.

This would remove redundant code

sim/aviron/src/main.zig
```diff
@@ -85,21 +84,49 @@ pub fn main() !u8 {
...
 
                     try reader.seekTo(phdr.p_offset);
-                    try reader.interface.readSliceAll(dest_mem[addr_masked..][0..phdr.p_filesz]);
-                    @memset(dest_mem[addr_masked + phdr.p_filesz ..][0 .. phdr.p_memsz - phdr.p_filesz], 0);
+
+                    if (is_data) {
+                        // Load data segment via Bus interface
+                        // Read the segment data into a temporary buffer
+                        const segment_buf = try allocator.alloc(u8, phdr.p_filesz);
+                        defer allocator.free(segment_buf);
```
ikskuh: That buffer can be deleted and we can read byte-by-byte from the reader interface as it already performs buffering internally and we have a 4096 byte large buffer above, so it's already quite optimal and the additional copy would just take time and could fail

ikskuh (changes requested):
I really like the overall changes, but i want to make this incredibly good so we can build a nice and well suited test infrastructure on top of this

sim/aviron/src/main.zig
Grazfather: Done

sim/aviron/src/main.zig
Grazfather: what method do you use to read a byte from a reader?

sim/aviron/src/main.zig
Grazfather: But the interface would have to be expanded for returning u16.

sim/aviron/src/lib/Cpu.zig
Grazfather: Yeah, I guess we don't support doing anything with the eeprom at the moment.

sim/aviron/src/lib/bus.zig
Grazfather: Sure, but as it stands the Cpu doesn't access IO through anything but the IO bus, so e.g. `sbi` wants this method on the interface.

sim/aviron/src/lib/Cpu.zig
Grazfather: done

mattnite:
@Grazfather is this ready to merge?

Grazfather:
@mattnite No, I am taking a break from this right now, but @ikskuh offered to finish up what remains.

I think it's fine to merge as is, personally, but that implies that there's a small refactor that'll have to happen in the future.

On top of that, #683 will have to be modified and rebased.

If @ikskuh is OK with it you could probably merge this but I'd wait for his explicit (or tacit) approval.

PR #705, opened 2025-10-06
rp2xxx: Implement flashless flash hal

tact1m4n3:
The flash hal needs the rp2040 bootloader (at addr `XIP_BASE`) to setup flash which is not available on flashless targets. This PR makes the bootloader available for flashless targets as well. It should only be embedded if the flash hal is used.


examples/raspberrypi/rp2xxx/build.zig
```diff
@@ -25,11 +25,14 @@ pub fn build(b: *std.Build) void {
         .{ .target = raspberrypi.pico, .name = "pico_hd44780", .file = "src/rp2040_only/hd44780.zig" },
         .{ .target = raspberrypi.pico, .name = "pico_pcf8574", .file = "src/rp2040_only/pcf8574.zig" },
         .{ .target = raspberrypi.pico, .name = "pico_i2c_slave", .file = "src/rp2040_only/i2c_slave.zig" },
+
         .{ .target = raspberrypi.pico_flashless, .name = "pico_flashless_blinky", .file = "src/blinky.zig" },
+        .{ .target = raspberrypi.pico_flashless, .name = "pico_flashless_flash-program", .file = "src/rp2040_only/flash_program.zig" },
+
         .{ .target = raspberrypi.pico2_arm_flashless, .name = "pico2_arm_flashless_blinky", .file = "src/blinky.zig" },
         .{ .target = raspberrypi.pico2_riscv_flashless, .name = "pico2_riscv_flashless_blinky", .file = "src/blinky.zig" },
-        .{ .target = raspberrypi.pico2_arm_flashless, .name = "pico2_arm_flashless_system_timer", .file = "src/system_timer.zig" },
```
tact1m4n3, line 31: For consistency

examples/raspberrypi/rp2xxx/build.zig
```diff
@@ -25,11 +25,14 @@ pub fn build(b: *std.Build) void {
...
-        .{ .target = raspberrypi.pico2_riscv_flashless, .name = "pico2_riscv_flashless_system_timer", .file = "src/system_timer.zig" },
+        .{ .target = raspberrypi.pico2_arm_flashless, .name = "pico2_arm_flashless_system-timer", .file = "src/system_timer.zig" },
```
tact1m4n3, line 34: For consistency

port/raspberrypi/rp2xxx/src/boards/raspberry_pi_pico.zig
```diff
@@ -1,5 +1,7 @@
 pub const xosc_freq = 12_000_000;
 
+pub const bootrom = @import("shared/rp2040_bootrom.zig");
+
 comptime {
```
Grazfather, line 5: what's the purpose of this comptime block?

Grazfather (approved):

port/raspberrypi/rp2xxx/src/boards/raspberry_pi_pico.zig
tact1m4n3: We have to forcefully reference bootrom because it exports the bootloader (on non flashless targets).

Grazfather: Ah OK. I didn't know that that was a way to do that. Does that make the entire of that file comptime? And if so, are the comptime blocks in that file redundant?

tact1m4n3: Afaik zig starts evaluating a file only if it is referenced. If it is referenced (the board module is fortunately referenced by clock config code because of xosc_freq I think), all the comptime blocks are evaluated as well as export functions. The comptime block just references the bootrom file. This will in turn evaluate all comptime blocks inside it and all exported functions.

tact1m4n3: > Does that make the entire of that file comptime? And if so, are the comptime blocks in that file redundant?

It's not executing the content of the file in a comptime context. Basically, in the above comptime block, we just reference a symbol (bootrom).  It's not possible to execute a file at comptime, only blocks and functions.

Grazfather: ok thanks for clarifying

PR #709, opened 2025-10-17
adding some polish

aaronhmiller:
this patch is not big...more in the "returning the shopping cart to the nest" type of contribution. i was doing some fiddling w/ my new RP and found myself needing some additional guidance. adding some guidance to help the next person.

### DIFF
--- a/examples/raspberrypi/rp2xxx/scripts/usb_device_loopback.py
+++ b/examples/raspberrypi/rp2xxx/scripts/usb_device_loopback.py
@@ -19,7 +19,7 @@ try:
 
     # was it found?
     if dev is None:
-        raise ValueError('Device not found. Verify that device is attached, then check vendor and product settings.')
+        raise ValueError("Device not found. Verify that device is attached, then check vendor and product settings.")
 
     print("Device found!")
 
@@ -53,8 +53,10 @@ try:
         intf,
         custom_match=lambda e: usb.util.endpoint_direction(e.bEndpointAddress) == usb.util.ENDPOINT_IN)
 
-    assert inep is not None, "IN endpoint not found"
-    assert outep is not None, "OUT endpoint not found"
+    if inep is None:
+        raise ValueError("IN endpoint not found")
+    if outep is None:
+        raise ValueError("OUT endpoint not found")
 
     print(f"OUT endpoint: 0x{outep.bEndpointAddress:02x}")
     print(f"IN endpoint: 0x{inep.bEndpointAddress:02x}")
@@ -77,7 +79,6 @@ try:
     from_device = inep.read(64, timeout=2000)
     print(f"Received {len(from_device)} bytes")
     print("Device Says: {}".format(''.join([chr(x) for x in from_device])))
-
 except usb.core.USBTimeoutError:
     print("\nTimeout! The device isn't responding.")
 except usb.core.USBError as e:
@@ -96,11 +97,11 @@ finally:
                 dev.attach_kernel_driver(interface)
                 print(f"Kernel driver reattached to interface {interface}")
             except:
-                pass
+                print(f"Could not reattach kernel driver to interface {interface}")
         
         # Dispose of the device object
         usb.util.dispose_resources(dev)
     except:
-        pass
+        print("Attempts to cleanup by releasing interfaces failed.")
     
     print("Done!")


aaronhmiller:
@haydenridd you got me into this thing, it's only fair you take a look at reviewing this smol contrib

haydenridd:
Ran the scripts locally and they seem to work! Would you mind adding a short description of what changed/improved with these changes?

examples/raspberrypi/rp2xxx/scripts/list_devices.py
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env python3
```
Grazfather, line 1: this doesn't need to be rpi specific

examples/raspberrypi/rp2xxx/scripts/usb_device_loopback.py
```diff
@@ -10,39 +10,97 @@
...
-
-print("Device Says: {}".format(''.join([chr(x) for x in from_device])))
+try:
+    # find our device
+    # use list_devices.py for idVendor and idProduct values
+    dev = usb.core.find(idVendor=0x0000, idProduct=0x0001)
+
+    # was it found?
+    if dev is None:
+        raise ValueError('Device not found. Verify that device is attached, then check vendor and product settings.')
```
Grazfather: double quotes

aaronhmiller:
`list_devices.py` is a utility script that helps the user know which values to enter into `usb.core.find` section of the loopback script.

the updates to `usb_device_loopback.py` were made to allow for better error handling as well as more proper cleanup so that it can be run multiple times in succession. Previously, it would get into a corrupted state that I could only resolve by resetting the pico.

@haydenridd this help put the changes in context?

examples/raspberrypi/rp2xxx/scripts/list_devices.py
aaronhmiller: ah! should i remove it from here and put it somewhere else more generic?

examples/raspberrypi/rp2xxx/scripts/usb_device_loopback.py
```diff
@@ -10,39 +10,97 @@
...
+
+    outep = usb.util.find_descriptor(
+        intf,
+        custom_match=lambda e: usb.util.endpoint_direction(e.bEndpointAddress) == usb.util.ENDPOINT_OUT)
+
+    inep = usb.util.find_descriptor(
+        intf,
+        custom_match=lambda e: usb.util.endpoint_direction(e.bEndpointAddress) == usb.util.ENDPOINT_IN)
+
+    assert inep is not None, "IN endpoint not found"
```
Grazfather: don't use asserts for this. Asserts are for when invariants in logic are not met. errors. You can just print an error and exit.

examples/raspberrypi/rp2xxx/scripts/usb_device_loopback.py
```diff
@@ -10,39 +10,97 @@
...
+                print(f"Kernel driver detached from interface {interface}")
+            except:
+                pass
+
+    # Set configuration
+    try:
+        dev.set_configuration()
+        print("Configuration set")
+    except usb.core.USBError as e:
+        print(f"Configuration already set or error: {e}")
```
Grazfather, line 42: do we want to continue here?

examples/raspberrypi/rp2xxx/scripts/usb_device_loopback.py
```diff
@@ -10,39 +10,97 @@
...
+    print(f"\nSending: {test_string}")
+
+    bytes_written = outep.write(test_string, timeout=1000)
+    print(f"Wrote {bytes_written} bytes")
+    
+    time.sleep(0.1)  # Give device a moment to process
+    
+    from_device = inep.read(64, timeout=2000)
+    print(f"Received {len(from_device)} bytes")
+    print("Device Says: {}".format(''.join([chr(x) for x in from_device])))
```
Grazfather, line 81: Can you just do:
```suggestion
    print(f"Device Says: {from_device.decode('utf-8')}"))
```

examples/raspberrypi/rp2xxx/scripts/usb_device_loopback.py
```diff
@@ -10,39 +10,97 @@
...
+            try:
+                dev.attach_kernel_driver(interface)
+                print(f"Kernel driver reattached to interface {interface}")
+            except:
+                pass
+        
+        # Dispose of the device object
+        usb.util.dispose_resources(dev)
+    except:
+        pass
```
Grazfather: probably good to do something here

examples/raspberrypi/rp2xxx/scripts/list_devices.py
Grazfather: If we have a nice place to put it. tool/ in the top of the repo probably would be OK.

examples/raspberrypi/rp2xxx/scripts/usb_device_loopback.py
Grazfather: or raise an exception and handle it below

examples/raspberrypi/rp2xxx/scripts/usb_device_loopback.py
```diff
@@ -10,39 +10,97 @@
...
+        usb.util.release_interface(dev, 1)
+        usb.util.release_interface(dev, 0)
+        
+        # Reattach kernel driver if we detached it
+        for interface in detached_interfaces:
+            try:
+                dev.attach_kernel_driver(interface)
+                print(f"Kernel driver reattached to interface {interface}")
+            except:
+                pass
```
Grazfather: probably good to at least print that we weren't able to attach it. Maybe catch the specific exception type that the attach method raises when it fails.

examples/raspberrypi/rp2xxx/scripts/usb_device_loopback.py
aaronhmiller: tried the suggestion and it errors as follows:
```
Traceback (most recent call last):
  File "/Users/aaron.miller/Locals/github/zig-learning/microzig/examples/raspberrypi/rp2xxx/scripts/./usb_device_loopback.py", line 70, in <module>
    print(f"Device Says: {from_device.decode('utf-8')}")
                          ^^^^^^^^^^^^^^^^^^
AttributeError: 'array.array' object has no attribute 'decode'
```

examples/raspberrypi/rp2xxx/scripts/list_devices.py
aaronhmiller: i took a look in there and my concern is that as a python script, it'll confuse folks. maybe instead of writing it in python, I find a way to do it in zig and put that in /tools?
for now, it feels more expected to be here with the other red-headed python stepchildren.

Grazfather:
We can port and move it later. No worries

Thanks,

G

examples/raspberrypi/rp2xxx/scripts/usb_device_loopback.py
aaronhmiller: good catch. fixing.

examples/raspberrypi/rp2xxx/scripts/usb_device_loopback.py
aaronhmiller: will remove the asserts, which were there to maintain continuity with the previous version. changing to 
```
if inep is None:
    raise ValueError("IN endpoint not found")
if outep is None:
    raise ValueError("OUT endpoint not found")
 ```
 and this will also make it more consistent with the other parts.

examples/raspberrypi/rp2xxx/scripts/usb_device_loopback.py
aaronhmiller: i tried a continue, but couldn't figure it out. leaving as is for now.

examples/raspberrypi/rp2xxx/scripts/usb_device_loopback.py
aaronhmiller: agree the syntax isn't as clean as we might like, but it works for now and we can fix it some other day.

examples/raspberrypi/rp2xxx/scripts/usb_device_loopback.py
aaronhmiller: added some print output :)

examples/raspberrypi/rp2xxx/scripts/usb_device_loopback.py
aaronhmiller: good idea. added a print message that we can't reattach.

haydenridd (approved):
Agreed with @Grazfather that we should probably move this somewhere else, but you didn't put these files here so not relevant to this PR :)

examples/raspberrypi/rp2xxx/scripts/usb_device_loopback.py
Grazfather: no I mean do you want to error out here or continue running the script (as it's written now)

examples/raspberrypi/rp2xxx/scripts/usb_device_loopback.py
Grazfather: You might be able to do `.tobytes().decode('utf-8')`

examples/raspberrypi/rp2xxx/scripts/usb_device_loopback.py
aaronhmiller: gotcha. leaving as is for now. didn't come across any big problems (yet).

aaronhmiller:
thank you both for the thoughtful input and careful review.

PR #713, opened 2025-10-24
Drivers: AS5600 position sensor

Grazfather:


drivers/sensor/AS5600.zig
```diff
@@ -0,0 +1,175 @@
...
+        // Magnet detected
+        MD: u1 = 0,
+        reserved6: u2 = 0,
+    };
+
+    pub fn init(dev: mdf.base.I2C_Device) Self {
+        return Self{ .dev = dev };
+    }
+
+    pub inline fn read1_raw(self: *const Self, reg: Self.register) !u8 {
```
tact1m4n3: I suggest you don't use inline here. There was some consequence that it actually might prevent llvm from doing optimisations in the caller function. I don't remember exactly. But llvm should do the inlining anyways if thinks it a good choice.

drivers/sensor/AS5600.zig
```diff
@@ -0,0 +1,175 @@
...
+    pub fn read_status(self: *const Self) !Status {
+        const s = try self.read1_raw(register.STATUS);
+        return @bitCast(s & (0b111 << 3));
+    }
+
+    pub fn read_magnitude(self: *const Self) !u16 {
+        return self.read2_raw(register.MAGNITUDE);
+    }
+
+    // TODO: Write burn functions. Scary.
```
tact1m4n3, line 174: 😄

drivers/sensor/AS5600.zig
Grazfather: Yeah fair! trust the compiler

drivers/sensor/AS5600.zig
Grazfather: I have 4 pieces of the hardware so I will be able to test it out eventually :)

PR #720, opened 2025-10-27
Aviron: Return errors on memory/flash access error and instruction decode

Grazfather:
Instead of crashing the simulator when we have a problem with the
executing code, return an error and handle it in the runner. Make the
instruction implementations just transparently pass errors through.

### DIFF
--- a/sim/aviron/src/lib/Cpu.zig
+++ b/sim/aviron/src/lib/Cpu.zig
@@ -77,7 +77,7 @@ pub fn reset(cpu: *Cpu) void {
     cpu.sreg = @bitCast(@as(u8, 0));
 }
 
-pub const RunError = error{ InvalidInstruction, MemoryAccessError };
+pub const RunError = error{ InvalidInstruction, BusError };
 pub const RunResult = enum {
     breakpoint,
     enter_sleep_mode,
@@ -156,7 +156,7 @@ pub fn run(cpu: *Cpu, mileage: ?u64, breakpoint: ?u24) RunError!RunResult {
             std.debug.print("Failed to fetch instruction at PC 0x{X:0>6}\n", .{pc});
             std.debug.print("Error: {}\n", .{err});
             if (cpu.trace) cpu.dump_system_state();
-            return error.MemoryAccessError;
+            return error.BusError;
         };
         const inst = isa.decode(code) catch |err| {
             std.debug.print("\n=== INSTRUCTION DECODE ERROR ===\n", .{});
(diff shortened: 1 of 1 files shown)


sim/aviron/src/lib/Cpu.zig
```diff
@@ -151,7 +151,19 @@ pub fn run(cpu: *Cpu, mileage: ?u64, breakpoint: ?u24) RunError!RunResult {
...
+            std.debug.print("Error: {}\n", .{err});
+            if (cpu.trace) cpu.dump_system_state();
+            return error.MemoryAccessError;
+        }) catch |err| {
+            std.debug.print("\n=== INSTRUCTION DECODE ERROR ===\n", .{});
+            std.debug.print("Failed to decode instruction at PC 0x{X:0>6}\n", .{pc});
+            std.debug.print("Error: {}\n", .{err});
+            if (cpu.trace) cpu.dump_system_state();
+            return error.InvalidInstruction;
+        };
```
ikskuh: Please split this code into two lines, it's hard to read

sim/aviron/src/lib/Cpu.zig
```diff
@@ -187,9 +199,21 @@ pub fn run(cpu: *Cpu, mileage: ?u64, breakpoint: ?u24) RunError!RunResult {
...
                     } else {
-                        @field(instructions, @tagName(tag))(cpu, info);
+                        @field(instructions, @tagName(tag))(cpu, info) catch |err| {
+                            std.debug.print("\n=== MEMORY ACCESS ERROR ===\n", .{});
+                            std.debug.print("Failed executing instruction at PC 0x{X:0>6}: {f}\n", .{ pc, fmt_instruction(inst) });
+                            std.debug.print("Error: {}\n", .{err});
+                            if (cpu.trace) cpu.dump_system_state();
+                            return error.MemoryAccessError;
+                        };
                     }
```
ikskuh: This codeblock can also be simplified by removing the duplication:
```zig
const exec_result = if (@TypeOf(info) == void)
    @field(instructions, @tagName(tag))(cpu)
else
    @field(instructions, @tagName(tag))(cpu, info);

exec_result catch |err| { … };
```

sim/aviron/src/lib/bus.zig
```diff
@@ -395,10 +386,78 @@ test "MemoryMapping: basic read/write across segments" {
...
+    const Mapping = MemoryMapping(BusType);
+
+    // Only map 0x0020-0x005F (64 bytes at 0x20)
+    const segs = [_]Mapping.Segment{
+        .{ .at = 0x0020, .size = 0x40, .backend = sram_dev },
+    };
+
+    var gpa = std.heap.GeneralPurposeAllocator(.{}){};
+    defer _ = gpa.deinit();
+    const alloc = gpa.allocator();
```
ikskuh: `std.testing.allocator` solves this

sim/aviron/src/lib/Cpu.zig
Grazfather: Much nicer. Thanks!

PR #726, opened 2025-11-07
rp2xxx: Enable fpu and dcp in core 1 entrypoint

tact1m4n3:
PS: I haven't actually tested that it works, but it should 😄

### DIFF
--- a/drivers/framework.zig
+++ b/drivers/framework.zig
@@ -43,6 +43,7 @@ pub const led = struct {
 };
 
 pub const sensor = struct {
+    pub const AS5600 = @import("sensor/AS5600.zig").AS5600;
     pub const ICM_20948 = @import("sensor/ICM-20948.zig").ICM_20948;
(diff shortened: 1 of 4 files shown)


port/raspberrypi/rp2xxx/src/hal.zig
```diff
@@ -155,6 +129,39 @@ pub fn init_sequence(comptime clock_cfg: clocks.config.Global) void {
...
+        if (microzig.options.hal.enable_fpu) {
+            if (is_fpu_used) {
+                // enable lazy state preservation
+                microzig.cpu.peripherals.fpu.FPCCR.modify(.{
+                    .ASPEN = 1,
+                    .LSPEN = 1,
+                });
+
+                // enable the FPU
+                cpacr |= 0xF << 20;
```
Grazfather: why not just `.modify`

PR #727, opened 2025-11-09
Cooperative multitasking

piotrfila:
Hello!

I was nerd-sniped by the discord server into writing an async proof of concept.
It's heavily inspired by zig's upcoming Io interface.
Futures and awaiting are not yet implemented, so all tasks just run indefinitely.

Is this within scope of Microzig? Or do you think it's a better idea to split this into another repo?

Comments and PRs are very welcome.

### DIFF
--- a/port/raspberrypi/rp2xxx/src/hal/Io.zig
+++ b/port/raspberrypi/rp2xxx/src/hal/Io.zig
@@ -1,0 +1,31 @@
+const std = @import("std");
+const microzig = @import("microzig");
+const rp2xxx = @import("../hal.zig");
+const Io = microzig.core.Io;
+
+/// See Io.VTable for function descriptions.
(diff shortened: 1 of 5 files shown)


port/raspberrypi/rp2xxx/src/hal/Io.zig
```diff
@@ -0,0 +1,303 @@
...
+fn to_stack_units(size: usize) usize {
+    return std.math.divCeil(
+        usize,
+        size,
+        @sizeOf(StackUint),
+    ) catch unreachable;
+}
+
+/// Information about when a task should be resumed.
+/// This could me made into an interface.
```
piotrfila: Alternatively, waiting for peripherals could be implemented by generic read address, mask and compare.

PR #728, opened 2025-11-15
Attempt at enabling TinyUSB with microzig

Kytezign:
# Introduction
Figured I'd share my progress so far in enabling tinyusb with microzig.
This compiles tinyusb + pico-sdk headers into a static library and links to the firmware.  
The short of it is that this seems to work for RP2040 from my limited testing.

Anyway, I don't think this is ready to be pulled in but I am curious to get some feedback from people who might know what they are doing, I embarked on this as a learning exercise and stumbled through it.

Usage example here:  https://github.com/Kytezign/rp2xxx_tusb_lib_build

Please let me know your thoughts.  

# Structural
## Tinyusb submodule
I've added a tinyusb module into the microzig project and added it as a new dependency.  That tinyusb sub-project is a build script with a few dependencies beyond just tinyusb including the pico-sdk and newlib (for libc headers pico-sdk & tinyUSB expect). 

The submodule pulls in a specific pico-sdk and related changeset if tinyusb (based on what the pico-sdk points to as a git-submodule).  I've added a function into microzig build that links a library to the firmware being built and adds the relevant tinyusb headers to the root module for usage. 

## microzig to tinyusb interface
In addition to that microzig submodule, there is also a new file in the RP2040 hal area which exports a few functions needed by tinyusb.  A lot of it is related to debugging tinyusb (perhaps not needed when debug logging is disabled).  

In the end I've done pretty limited testing - only CDC & MCS so far.   But they've been pretty solid. 

# Other thoughts
There are a lot of issues currently I'm sure, but the ones I'm most keenly aware of are:
- Generally probably needs to be cleaned up. 
- Probably not the most zig focused way to do this - builds a lot of c code and relies on the translate-C feature to bring in the API.

## next steps
- pico sdk should be lazy dependency
- create basic, easy to use, cdc interface with magic baud rate setting that will reboot

### DIFF
--- a/modules/tinyusb/build.zig
+++ b/modules/tinyusb/build.zig
@@ -6,6 +6,10 @@ const Build = std.Build;
 pub fn build(b: *Build) !void {
     addTinyUSBLib(b);
 }
+const tusb_h_contents =
+    \\#include_next "tusb.h"
+    \\#define TU_ATTR_FAST_FUNC       __attribute__((section(".ram_text")))
+;
 
 pub fn addTinyUSBLib(b: *Build) void {
     const module = b.addModule("tinyusb", .{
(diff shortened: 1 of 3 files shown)


Kytezign:
Due to not a lot of interest from the group and my own reservations about this approach, I'm closing this.  
I've moved over to a standalone package for microzig <--> tinyusb: https://github.com/Kytezign/zig_tusb_lib 

Example usage:  https://github.com/Kytezign/example_microzig_tusb_logging

PR #734, opened 2025-11-25
rp2xxx: Update readme to describe boot process and flashless targets

Grazfather:
I found my old notes I was writing when I originally added support for flashless targets.

I cleaned it up and gave it a once over. Hopefully it's helpful.

### DIFF
--- a/port/raspberrypi/rp2xxx/README.md
+++ b/port/raspberrypi/rp2xxx/README.md
@@ -32,10 +32,9 @@ HAL and register definitions for the RP2040 and RP2350 chips.
     - This linker builds the bootrom and creates an import, for the later build to `@embedFile`,
       pad and slap the CRC on.
 - *rp2040.ld*:
-  - Includes `.boot2` section.
-    - Which is the embedded bootloader file, embedded in _bootrom.zig_.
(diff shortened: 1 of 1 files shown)


port/raspberrypi/rp2xxx/README.md
```diff
@@ -2,6 +2,140 @@
...
+RAM image linker scripts (e.g., `ld/rp2350/arm_ram_image_sections.ld`) add:
+
+**`.bootmeta` section** (RP2350 only):
+- Contains image definition block for bootrom
+- Specifies image type, CPU architecture, security settings
+- For ARM RAM images, includes custom entry point to override default vector table expectation
+
+**`.ram_vectors` section** (NOLOAD):
+- Space for runtime-modifiable vector table
+- Not initialized from flash
```
tact1m4n3: This section isn't used anymore. I changed the code to use .data or .rodata (on riscv) and on cortex_m it is placed in data because ram_vector_table is declared using var. We can probably take it out. Is there a good reason to use a custom section?

tact1m4n3: #735

Grazfather: Happy to update! This was just from what I gleaned from reading it, and it makes sense it's out of date.

PR #735, opened 2025-11-25
rp2xxx: Cleanup linker scripts

tact1m4n3:
Remove unused `.ram_vectors` and `.core_vectors` sections


tact1m4n3:
A while back when I did the ram vector table refactoring.

PR #736, opened 2025-11-25
rp2xxx: PIO: Fix sideset having no way to account for optional bit

Grazfather:
If sideset is optional, then SIDESET_COUNT needs to account for that extra bit.

Because that option was not available to `sm_set_pin_mappings`, I had to plumb it through.

### DIFF
--- a/port/raspberrypi/rp2xxx/src/hal/pio/common.zig
+++ b/port/raspberrypi/rp2xxx/src/hal/pio/common.zig
@@ -517,11 +517,9 @@ pub fn PioImpl(EnumType: type, chip: Chip) type {
             program: Program,
             options: LoadAndStartProgramOptions(chip),
         ) !void {
+            // We don't account for the optional bit here because it's not known to the pin_mappings
             const expected_side_set_pins = if (program.side_set) |side_set|
-                if (side_set.optional)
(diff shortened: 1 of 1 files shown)


port/raspberrypi/rp2xxx/src/hal/pio/common.zig
```diff
@@ -273,7 +273,7 @@ pub fn PioImpl(EnumType: type, chip: Chip) type {
             });
         }
 
-        pub fn sm_set_pin_mappings(self: EnumType, sm: StateMachine, options: PinMappingOptions) !void {
+        pub fn sm_set_pin_mappings(self: EnumType, sm: StateMachine, options: PinMappingOptions, side_set_optional: bool) !void {
```
Grazfather, line 276: We could make it s/t the sideset config part of `PinMappingOptions` has an extra field, and maybe we can then take that field away from the ExecOptions

Grazfather: Actually, that would just make `sm_set_exec_options` more annoying, since we'd have to plumb it _there_ instead.

mattnite (approved):

PR #742, opened 2025-11-29
ch32v: Add time keeping via systick

Grazfather:
* Add timekeeping via SysTick
* Add delay functions
* Add LANA_TNY board
* Add ws2812 example

### DIFF
--- a/examples/wch/ch32v/build.zig
+++ b/examples/wch/ch32v/build.zig
@@ -31,6 +31,7 @@ pub fn build(b: *std.Build) void {
         .{ .target = mb.ports.ch32v.chips.ch32v203x6, .name = "blinky_ch32v203", .file = "src/blinky.zig" },
         .{ .target = mb.ports.ch32v.chips.ch32v203x6, .name = "blinky_systick_ch32v203", .file = "src/blinky_systick.zig" },
         .{ .target = mb.ports.ch32v.boards.ch32v203.suzuduino_uno_v1b, .name = "suzuduino_blinky", .file = "src/board_blinky.zig" },
+        .{ .target = mb.ports.ch32v.boards.ch32v203.lana_tny, .name = "lana_tny_ws2812", .file = "src/ws2812.zig" },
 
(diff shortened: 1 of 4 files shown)


port/wch/ch32v/src/hals/time.zig
```diff
@@ -0,0 +1,119 @@
...
+///
+/// NOTE: This must be called AFTER configuring the system clock to the final frequency.
+pub fn init() void {
+    // Reset configuration
+    PFIC.STK_CTLR.raw = 0;
+
+    // Reset the count register
+    PFIC.STK_CNTL.raw = 0;
+
+    // Configure SysTick to trigger every 1ms
```
Grazfather, line 30: can we/should we up this? `sleep_us` loses granularity otherwise

port/wch/ch32v/src/hals/time.zig
```diff
@@ -0,0 +1,171 @@
...
+    if (@hasField(cpu.csr, "mcountinhibit")) {
+        const ci = cpu.csr.mcountinhibit.read();
+        if ((ci & 0x1) != 0) cpu.csr.mcountinhibit.write(ci & ~@as(u32, 1));
+    }
+
+    // Probe whether the cycle counter advances at all; if not, fall back.
+    const probe0: u32 = cpu.csr.cycle.read();
+    asm volatile ("" ::: .{ .memory = true });
+    const probe1: u32 = cpu.csr.cycle.read();
+    if (probe0 == probe1) {
```
Grazfather, line 127: In practice on my board we use this fallback. Not sure why. Would love if someone else could verify (remove fallback and see if the delay hangs). If wch broke the CSR stuff we can just remove it and default to this cycle counting implementation.

port/wch/ch32v/src/hals/time.zig
Grazfather: Moved to using a cycle-counting implementation for sleep

PR #745, opened 2025-12-02
Improve I2C device compatibility for STM32F303

mathk:
Improvement to the STM32F303 I2C.

I would like to have some timeout management and better error handling in the I2C before moving on.

### DIFF
--- a/port/stmicro/stm32/src/hals/STM32F303/i2c.zig
+++ b/port/stmicro/stm32/src/hals/STM32F303/i2c.zig
@@ -6,48 +6,16 @@ const I2CType = emus_type.I2CType;
 const I2C_t = microzig.chip.types.peripherals.i2c_v2.I2C;
 const peripherals = microzig.chip.peripherals;
 const hal = microzig.hal;
+const mdb = microzig.drivers.base;
 
-pub const AddressMode = enum {
-    @"7bits",
-    @"10bits",
-};
-
-pub const Address = struct {
-    mode: AddressMode = .@"7bits",
-    addr: u10,
-
-    pub fn new(addr: u7) Address {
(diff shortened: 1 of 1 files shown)


Grazfather:
Can you please break this up into two PRs? One for the HAL and one for the generic driver?

mathk:
> Can you please break this up into two PRs? One for the HAL and one for the generic driver?

Split done. I also include  a common folder for stm32 hal since the I2C implementation is specific to the register version generated by regz (from the embassy project). so in the future one could reuse the same implementation if the chip contain the peripheral with same version. (i2c_v2 for instance).

I am planning to do the same for uart actually.

port/stmicro/stm32/src/hals/STM32F303/i2c.zig
```diff
@@ -62,25 +30,11 @@ const TimmingSpec_Standard = .{
...
-    }
-}
-
 fn get_regs(comptime instance: I2CType) *volatile I2C_t {
     return @field(microzig.chip.peripherals, @tagName(instance));
 }
-pub const I2C = struct {
+
+const I2C = struct {
     regs: *volatile I2C_t,
```
Grazfather, line 38: I get that this was already here, but I wonder why the stm32 HALs uniquely have the I2C object be a struct with the address rather than an enum. In the others `get_regs()` is called on the time, but it's inline and comptime resolved. We can do this later, but unless there's a good reason for one over the other, I think that we should try to make them consistent across chips.

Grazfather: We can do this later, though.

Grazfather (approved):

mattnite (approved):

PR #746, opened 2025-12-02
Add HTS221 sensor driver (temperature only)

mathk:


drivers/sensor/HTS221.zig
```diff
@@ -0,0 +1,217 @@
...
+//! * HTS221: https://www.st.com/resource/en/datasheet/hts221.pdf
+//!
+const std = @import("std");
+const mdf_base = @import("../framework.zig").base;
+
+const assert = std.debug.assert;
+
+// Simpler version of MMIO to ease register setting.
+// I guess this would be a nice abstraction to have in the framework base
+fn Reg(comptime PackedT: type) type {
```
Grazfather: Yeah this could be moved somewhere. If you move it we should file an issue to update other drivers to use it.

drivers/sensor/HTS221.zig
```diff
@@ -0,0 +1,217 @@
...
+pub const HTS221 = struct {
+    const address: mdf_base.I2C_Device.Address = @enumFromInt(0b1011111);
+    const RegsAddr = enum(u8) {
+        WHO_AM_I = 0x0F,
+        AV_CONF = 0x10,
+        CTRL_REG1 = 0x20,
+        T_OUT = 0x2A,
+        STATUS_REG = 0x27,
+
+        // Callibration register
```
Grazfather: typo Calibration

drivers/sensor/HTS221.zig
```diff
@@ -0,0 +1,217 @@
...
+
+    const AV_CONF = Reg(packed struct(u8) {
+        AVGH: AVGH,
+        AVGT: AVGT,
+        reserved0: u2,
+    });
+
+    dev: *mdf_base.I2C_Device,
+
+    // Needs to be initialize with calibration value.
```
Grazfather:
```suggestion
    // Needs to be initialized with calibration value.
```
typo

drivers/sensor/HTS221.zig
```diff
@@ -0,0 +1,217 @@
...
+    t_slop: f32 = 0.0,
+    t_b0: f32 = 0.0,
+
+    // internal read buffer
+    read_buffer: [2]u8 = [2]u8{ 0, 0 },
```
Grazfather:
```suggestion
    read_buffer: [2]u8 = .{ 0, 0 },
```

Grazfather: Or could do `read_buffer = [_]u8{ 0, 0 }`

drivers/sensor/HTS221.zig
```diff
@@ -0,0 +1,217 @@
...
+        const t1_deg = @as(f32, @floatFromInt(@as(i10, @bitCast(@as(u10, @intCast(self.read_buffer[0])))) | t1msb)) / 8.0;
+
+        try self.dev.write_then_read(address, &[_]u8{RegsAddr.T0_OUT.auto_v()}, &self.read_buffer);
+        const T0 = @as(f32, @floatFromInt(@as(i16, @intCast(self.read_buffer[0])) | (@as(i16, @intCast(self.read_buffer[1])) << 8)));
+
+        try self.dev.write_then_read(address, &[_]u8{RegsAddr.T1_OUT.auto_v()}, &self.read_buffer);
+        const T1 = @as(f32, @floatFromInt(@as(i16, @intCast(self.read_buffer[0])) | (@as(i16, @intCast(self.read_buffer[1])) << 8)));
+
+        self.t_slop = ((t1_deg - t0_deg) / (T1 - T0));
+        self.t_b0 = t0_deg - T0 * self.t_slop;
```
Grazfather, line 143: can you please break up long (>100char) lines?

drivers/sensor/HTS221.zig
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    fn setup_output_rate(self: *@This(), rate: ODR) !void {
```
Grazfather: We can say that these are InterfaceErrors

drivers/sensor/HTS221.zig
```diff
@@ -0,0 +1,217 @@
...
+        try self.dev.write_then_read(address, &[_]u8{RegsAddr.CTRL_REG1.v()}, self.read_buffer[0..1]);
+        var reg = CTRL_REG1.from(self.read_buffer[0]);
+        reg.modify(.{ .ODR = rate });
+        try self.dev.write(address, &[_]u8{ RegsAddr.CTRL_REG1.v(), reg.raw });
+    }
+
+    fn setup_average(self: *@This(), avgt: AVGT, avgh: AVGH) !void {
+        try self.dev.write_then_read(address, &[_]u8{RegsAddr.AV_CONF.v()}, self.read_buffer[0..1]);
+        var reg = AV_CONF.from(self.read_buffer[0]);
+        reg.modify(.{
```
Grazfather: This is the only time you actually set more than one field with `modify`. We can probably just use simple structs, maybe an extra PR to add this reg type

drivers/sensor/HTS221.zig
```diff
@@ -0,0 +1,217 @@
...
+    dev: *mdf_base.I2C_Device,
+
+    // Needs to be initialize with calibration value.
+    t_slop: f32 = 0.0,
+    t_b0: f32 = 0.0,
+
+    // internal read buffer
+    read_buffer: [2]u8 = [2]u8{ 0, 0 },
+
+    fn init_temp_callibration(self: *@This()) !void {
```
Grazfather: typo 
```suggestion
    fn init_temp_calibration(self: *@This()) !void {
```

examples/stmicro/stm32/src/stm32f303/hts221.zig
```diff
@@ -0,0 +1,58 @@
...
+            .PIN6 = .{ .mode = .{ .alternate_function = .{
+                .afr = .AF4,
+            } } },
+            .PIN7 = .{ .mode = .{ .alternate_function = .{
+                .afr = .AF4,
+            } } },
+        },
+    }).apply();
+
+    // // I2C1 is the magnetometer on pin PB6/7
```
Grazfather:
```suggestion
    // I2C1 is the magnetometer on pin PB6/7
```

examples/stmicro/stm32/src/stm32f303/hts221.zig
```diff
@@ -0,0 +1,58 @@
...
+pub const microzig_options: microzig.Options = .{
+    .logFn = microzig.hal.uart.log,
+};
+
+pub noinline fn init() void {
+    microzig.board.init();
+    microzig.board.init_log();
+}
+
+pub noinline fn main() !void {
```
Grazfather, line 23: what was the inline issue? Do we need to fix other examples?

mathk: The issue is due to the FPU. and release mode. If the main is inline into microzig_main the function will start with some vpush instruction before you can enable fpu  which rise a fault. I guess a better approach would be to enable FPU at the very beginning and even maybe decide that at build time with some options.

drivers/sensor/HTS221.zig
mathk: > Or could do `read_buffer = [_]u8{ 0, 0 }`

I would go with your first suggestion since this would confuse the compiler with tuple.

drivers/sensor/HTS221.zig
mathk: I would suggest for consistency to always use the Regs abstraction. Even when the solution could be accomplish with struct alone. (It also hide those bitCast). Wdyt?

Grazfather: How does it hide bitcasts? Whether you use the field or modify, you have to truncate or cast to get the types to work.

I agree we should always use it if we have it for consistency, but I am questioning whether we want to add it at all at this point since it's only actually slightly nicer here.

If we wanted to go wacky, we could make the type generator take a read and write function, so that modify _actually_ did a rmw, but I think that that's over the top.

examples/stmicro/stm32/src/stm32f303/hts221.zig
Grazfather: Yes that absolutely should be cleaned up.

drivers/sensor/HTS221.zig
```diff
@@ -0,0 +1,231 @@
...
+
+// Simpler version of MMIO to ease register setting.
+// I guess this would be a nice abstraction to have in the framework base
+fn Reg(comptime PackedT: type) type {
+    const size = @bitSizeOf(PackedT);
+
+    const IntT = std.meta.Int(.unsigned, size);
+
+    if (@sizeOf(PackedT) != (size / 8))
+        @compileError(std.fmt.comptimePrint("IntT and PackedT must have the same size!, they are {} and {} bytes respectively", .{ size / 8, @sizeOf(PackedT) }));
```
Grazfather: (if we keep this), split lines please

examples/stmicro/stm32/src/stm32f303/hts221.zig
```diff
@@ -0,0 +1,57 @@
+const std = @import("std");
+const microzig = @import("microzig");
+const stm32 = microzig.hal;
+const HTS221 = microzig.drivers.sensor.HTS221;
+
+fn delay() void {
```
Grazfather, line 6: do we not have a sleep or delay function for this hal?

examples/stmicro/stm32/src/stm32f303/hts221.zig
```diff
@@ -0,0 +1,57 @@
...
+    });
+    try hts221.turn_on();
+
+    while (true) {
+        delay();
+
+        if (try hts221.temperature_ready()) {
+            const temp = try hts221.read_temperature();
+
+            std.log.info("Reading temp sensor {} ", .{temp});
```
Grazfather: would be nice to add units

examples/stmicro/stm32/src/stm32f303/hts221.zig
mathk: STM32F103 yes but not in STM32F303, I guess we need to be more align across STM32 hal

examples/stmicro/stm32/src/stm32f303/hts221.zig
Grazfather: @tact1m4n3 is working on this

examples/stmicro/stm32/src/stm32f303/hts221.zig
Grazfather: for later. just wanted to make sure.

PR #747, opened 2025-12-04
Add HAL for the STM32L47X and move peripherals to common folder

mathk:


port/stmicro/stm32/src/generate.zig
```diff
@@ -259,6 +259,14 @@ fn generate_chips_file(
                 \\
             );
         }
+        if (std.mem.startsWith(u8, chip_file.name, "STM32L47")) {
```
mathk, line 262: I would be happy to change that but not sure how.

port/stmicro/stm32/src/hals/STM32F303/enums.zig
```diff
@@ -1,7 +1,7 @@
 pub const UARTType = enum(usize) {
-    UART1,
```
mathk, line 2: I think it would be better to generate those enums from regz.

port/stmicro/stm32/src/generate.zig
mattnite: Same here, there might be some sort of family field from embassy that we could use in the future, but this is good for now.

mattnite:
Great work!

PR #749, opened 2025-12-04
 cortex_m: Enable fpu on startup

tact1m4n3:


core/src/cpus/cortex_m.zig
```diff
@@ -694,7 +717,7 @@ pub const startup_logic = struct {
             enable_fault_irq();
         }
 
-        microzig_main();
+        @call(.never_inline, microzig_main, .{});
```
mathk, line 771: NP: instead can we use the noinline  from the `microzig.zig` ? it won't be that big of an issue for other  chip ?

tact1m4n3: Imo it's ok that calls inside microzig_main get inlined because the FPU is already enabled when execution gets there.

core/src/cpus/cortex_m.zig
```diff
@@ -684,6 +703,10 @@ pub const startup_logic = struct {
                 : .{ .memory = true, .r0 = true, .r1 = true });
         }
 
+        if (fpu_present and builtin.abi.float() == .hard) {
+            enable_fpu();
```
mathk, line 741: we could maybe clean the hal that have this at init ?

core/src/cpus/cortex_m.zig
mathk: this would bubble up the hf store to the startup function and you end up with the same issue IIUC.

tact1m4n3: Sorry, I don't understand what you are referring to. Can you elaborate?

core/src/cpus/cortex_m.zig
tact1m4n3: I've cleaned up RP2xxx and STM32F303. Are there any others?

core/src/cpus/cortex_m.zig
```diff
@@ -643,6 +643,24 @@ pub const atomic = struct {
...
 };
 
+/// Enables the FPU.
+///
+/// NOTE: This function is automatically called on cpu startup if the cpu has
+/// an fpu and hard float is enabled for the target. HAL's also call this in
+/// the startup of other cores.
+pub inline fn enable_fpu() void {
+    // Taken from the rust crate cortex-m-rt.
+    asm volatile (
```
tact1m4n3, line 678: I've implemented this using raw assembly since we might be missing some register definitions for FPU in some cortex-m variants and also it allows being embedded in naked functions.

core/src/cpus/cortex_m.zig
```diff
@@ -970,10 +992,13 @@ const scb_base = scs_base + core.scb_base_offset;
...
 const fpu_base = scs_base + 0x0F34;
 
-const properties = microzig.chip.properties;
 // TODO: will have to standardize this with regz code generation
-const mpu_present = @hasDecl(properties, "cpu.mpuPresent") and std.mem.eql(u8, properties.@"cpu.mpuPresent", "true");
-const fpu_present = @hasDecl(properties, "cpu.fpuPresent") and std.mem.eql(u8, properties.@"cpu.fpuPresent", "true");
+const mpu_present = @hasDecl(microzig.chip, "properties") and
+    @hasDecl(microzig.chip.properties, "cpu.mpuPresent") and
+    std.mem.eql(u8, microzig.chip.properties.@"cpu.mpuPresent", "true");
+const fpu_present = @hasDecl(microzig.chip, "properties") and
```
tact1m4n3, line 1051: Sometimes code generated by regz doesn't include a properties declaration => if you used in the past fpu_present or mpu_present you got a compilation error.

core/src/cpus/cortex_m.zig
```diff
@@ -684,6 +702,10 @@ pub const startup_logic = struct {
                 : .{ .memory = true, .r0 = true, .r1 = true });
         }
 
+        if (builtin.abi.float() == .hard) {
+            enable_fpu();
```
mattnite, line 741: I like that this change makes microzig have more "batteries included" by default, but let's allow the user to disable the FPU at build time, with a flag named `disable_fpu`.

mattnite (changes requested):
Looks good to me once that flag is done, and the conversation with @mathk is resolved.

core/src/cpus/cortex_m.zig
tact1m4n3: I can add this but isn't it a footgun? You could achieve the same by switching to soft float in the rp2350_arm target in build.zig and you don't risk crashing.

core/src/cpus/cortex_m.zig
mathk: Ah sorry I think I was not clear. I though of flagging the microzig_main as noinline for every CPU not only arm. Just to simplify this @call other arch could suffer the same issue. It might also deserve a small explanation why we don't want microzig_main not to be inlined ? Feel free to discard this comment if you don't think it worth the effort.

core/src/cpus/cortex_m.zig
mathk: I don't see any others

core/src/cpus/cortex_m.zig
Grazfather: Definitely worth a comment to explain.

tact1m4n3: > I though of flagging the microzig_main as noinline for every CPU not only arm.

How can this be achieved? Idk of a way to make calls to microzig_main never inline (since it's always referenced as extern). But I will mark this PR as draft for now. I want to do a bit of refactoring.

tact1m4n3: I added the comment. Is it ok?

core/src/cpus/cortex_m.zig
tact1m4n3: I added the option

core/src/cpus/cortex_m.zig
```diff
@@ -33,8 +33,25 @@ pub const InterruptOptions = microzig.utilities.GenerateInterruptOptions(&.{
...
+    ///
+    /// NOTE: This option is enabled in debug builds by default.
+    verbose_unhandled_irq: bool = (builtin.mode == .Debug),
+
+    /// If true, the FPU will be enabled in the startup code.
+    ///
+    /// NOTE: This option is enabled by default if hard float is enabled.
+    /// NOTE: Not supported on cortex_m0, cortex_m0plus and cortex_m3.
+    enable_fpu: bool = builtin.abi.float() == .hard,
+};
```
tact1m4n3, line 54: I made common CPU variant between cortex m flavours to be easier to add options. I also added compilation errors if for instance you try to use a ram vector table on cortex m0 or if you try to enable the FPU on the cortex ms that don't support it.

tact1m4n3: Let me know what you think of this

tools/regz/src/patch.zig
```diff
@@ -23,6 +23,12 @@ pub const Patch = union(enum) {
         device_name: []const u8,
         arch: Arch,
     },
+    set_device_property: struct {
+        device_name: []const u8,
+        key: []const u8,
+        value: []const u8,
+        description: ?[]const u8 = null,
+    },
```
tact1m4n3, line 31: @mattnite what do you think of this?

tact1m4n3:
Only one last thing remains: standardising fpu and mpu present chip properties (there is a long standing TODO about this in cortex_m.zig)

tools/regz/src/atdf.zig
```diff
@@ -121,11 +121,21 @@ fn load_param(ctx: *Context, node: xml.Node, device_id: DeviceID) !void {
         "caption",
     });
 
+    const property_name_map: std.StaticStringMap([]const u8) = .initComptime(.{
```
tact1m4n3, line 124: Is it ok to hardcode these here? Same applies for svd

tools/regz/src/Database.zig
```diff
@@ -435,6 +435,20 @@ fn gen_sql_table_impl(comptime name: []const u8, comptime T: type) ![]const u8 {
...
+            if (first) {
+                first = false;
+            } else {
+                try fbs.writeAll(", ");
+            }
+            try fbs.writeAll(col_name);
+        }
+        try fbs.writeAll(")");
+    }
+
```
tact1m4n3, line 451: unique constraints used to be ignored. Hope I implemented this right. I don't have much experience with sql :)

tools/regz/src/atdf.zig
tact1m4n3: These should probably be documented. Currently we just rename the properties. Another option which I like is to duplicate them with the microzig standard name (keep __MPU_PRESENT but also export mpu_present). But this makes set_device_property a bit weird. For this to work we should add some sort of property aliasing at least for the database (where if you change one, you change the other).

tact1m4n3:
> Only one last thing remains: standardising fpu and mpu present chip properties (there is a long standing TODO about this in cortex_m.zig)

Is this needed though? A different approach would be to check both fields (__FPU_PRESENT and @"cpu.fpuPresent") in cortex_m.zig when we assign fpu_present.

mattnite:
@tacit1m4n3 I'm not sure where `__FPU_PRESENT` is from, maybe embassy? But I do know `cpu.fpuPresent` is part of the CMSIS/SVD spec so what you've outlined is good.

core/src/cpus/cortex_m.zig
mattnite: Great comment, I was unsure why this was an issue but that makes sense to me!

tools/regz/src/Database.zig
mattnite: Lmao me neither, but it looks pretty good to me

tools/regz/src/atdf.zig
mattnite: I'm fine with this for now, but honestly we should standardize these for regz/microzig soon. Similar to how we transform register schemas to a common structure, we can do the same for some specific properties, and leave the rest as they come.

tools/regz/src/patch.zig
mattnite: 10/10, I promise some patch improvements are coming soon -- this works great with them.

tact1m4n3:
> @tact1m4n3 I'm not sure where __FPU_PRESENT is from

I got a CI compile error from an atsam chip that it didn't have an fpu available. Except it did but the property was called  `__FPU_PRESENT = "1"` :) So I figured that is the convention for atdf specs (is it?). This is why I added the __FPU_PRESENT -> fpu_present mapping in atdf.zig

core/src/cpus/cortex_m.zig
```diff
@@ -643,6 +660,32 @@ pub const atomic = struct {
     }
 };
 
+/// Enables the FPU.
+///
+/// NOTE: This function is automatically called on cpu startup if the cpu has
+/// an fpu and hard float is enabled. HAL's also call this in the startup of
```
Grazfather: nit HALs

core/src/cpus/cortex_m.zig
Grazfather: no clobbers? I get there's a dsb and isb, but couldn't the compile still use cached registers, especially since this is inline?

Grazfather (approved):
I didn't look at the Regz part, and I don't fully understand the nuance of how or when fpu instructions will sneak in earlier, but this looks good.

core/src/cpus/cortex_m.zig
tact1m4n3: Oops, forgot about those :)

Grazfather: That's what code reviews are for :)

PR #750, opened 2025-12-04
esp: fix I2C writes

andreacoradi:
Before this small fix, I2C writes would incorrectly repeat the leading byte of the payload while excluding the last one.
By changing the order of the I2C commands, I was able to correctly write to my I2C device.
While consulting the TRM, I also noticed that calling `self.update_config();` was unnecessary in that context.

### DIFF
--- a/examples/espressif/esp/src/i2c_bus_scan.zig
+++ b/examples/espressif/esp/src/i2c_bus_scan.zig
@@ -7,7 +7,7 @@ const i2c = esp.i2c;
 const gpio = esp.gpio;
 const peripherals = microzig.chip.peripherals;
 
-const i2c0 = i2c.instance.num(0);
+var i2c0 = i2c.instance.num(0);
 
 const usb_serial_jtag = esp.usb_serial_jtag;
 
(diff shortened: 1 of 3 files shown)


port/espressif/esp/src/hal/i2c.zig
```diff
@@ -428,44 +436,13 @@ pub const I2C = enum(u1) {
...
-                    .ack_check_en = true,
-                    .length = 1,
-                } });
-                try self.add_cmd(cmd_start_idx, .{ .write = .{
-                    .ack_exp = .ack,
-                    .ack_check_en = true,
-                    .length = 1,
-                } });
-            }
+            try self.add_cmd(cmd_start_idx, .{ .write = .{
```
Grazfather, line 439: Do you understand enough of the esp-idf to have an idea why it was wacky like this? I basically just started with a port without understand fully why it split writes into extra commands

andreacoradi: So not with esp-idf, but looking at the esp-hal they just do it like this:
```rust
            if cfg!(esp32) && !(start || stop) {
            // NOT US
            } else {
                add_cmd(
                    cmd_iterator,
                    Command::Write {
                        ack_exp: Ack::Ack,
                        ack_check_en: true,
                        length: write_len as u8,
                    },
                )?;
            }

```

And from my testing it works as intended, so I'm not really sure.
Also, TRM just mentions to create one single write command with length N+1 when starting transmission, and length M if it's in-between.
So I would trust that more.

PR #754, opened 2025-12-06
feat: add ds18b20 driver

kaans:
Add a driver for the DS18B20 temperature sensor.

All commands except the Search Rom [0xF0] and Alarm Search [0xEC] have been implemented and successfully tested.
The resolution can be configured and is respected when converting the temperature value to Celsius.

The driver supports single drop configurations (i.e. only one sensor is connected) and multi drop configuraitons (i.e. multiple sensors are connected). If the functions are called without a target/rom code, the sensor is selected automatically if only one sensor is connected. If more than one sensor is connected, the rom code of the target sensor must be specified to address each sensor individually.

Parasitic power mode is not supported by the driver itself.

(I am new to zig, please let me know where I can improve the code)

### DIFF
--- a/drivers/sensor/DS18B20.zig
+++ b/drivers/sensor/DS18B20.zig
@@ -75,13 +75,18 @@ pub const DS18B20 = struct {
         external,
     };
 
+    pub const Error = error{
+        device_not_present,
+        rom_crc_mismatch,
+    };
+
     pub const RomCode = struct {
         value: [8]u8,
     };
 
-    pub const Error = error{
-        device_not_present,
-        rom_crc_mismatch,
+    pub const Config = struct {
+        alarms: Alarms,
+        resolution: Resolution,
     };
 
     pub fn init(pin: mdf.base.Digital_IO, clock_dev: mdf.base.Clock_Device) !Self {
@@ -182,7 +187,7 @@ pub const DS18B20 = struct {
     /// An error is returned if the device is not present.
     /// If target is null, the command is sent to all devices on the bus.
     /// Returns the alarms (TH and TL registers) and resolution.
-    pub fn read_config(self: *const Self, args: struct { target: ?RomCode = null }) !struct { alarms: Alarms, resolution: Resolution } {
+    pub fn read_config(self: *const Self, args: struct { target: ?RomCode = null }) !Config {
         if (try self.reset() == false) return Error.device_not_present;
 
         try self.select_target(args.target);
@@ -194,7 +199,7 @@ pub const DS18B20 = struct {
         const tl = try self.read_byte(); // TL register
         const config = try self.read_byte();
 
-        const resolution_bits: u2 = @intCast((config >> 5) & 0x03);
+        const resolution_bits: u2 = @truncate(config >> 5);
 
         return .{
             .alarms = .{ .th = th, .tl = tl },


drivers/sensor/DS18B20.zig
```diff
@@ -0,0 +1,390 @@
...
+    /// If alarms or resolution are null, the current value is preserved.
+    /// An error is returned if the device is not present.
+    /// If target is null, the command is sent to all devices on the bus.
+    /// The configuration can be saved to EEPROM using save_config_to_eeprom()
+    /// and recalled using recall_config_from_eeprom().
+    /// The alarms consist of two bytes: TH and TL registers. They can be used
+    /// to trigger an alarm when the temperature exceeds the set limits or to
+    /// store user data if the alarm functionality is not used.
+    pub fn write_config(self: *const Self, args: struct { alarms: ?Alarms = null, resolution: ?Resolution = null, target: ?RomCode = null }) !void {
+        if (!(try self.reset())) return error.DeviceNotFound;
```
tact1m4n3, line 149: Detecting the device on each command might not be ideal (performance wise). The other drivers usually do this kind of checking in init and then assume the device is there.

drivers/sensor/DS18B20.zig
```diff
@@ -0,0 +1,390 @@
...
+    /// Writes the configuration to the device.
+    /// If alarms or resolution are null, the current value is preserved.
+    /// An error is returned if the device is not present.
+    /// If target is null, the command is sent to all devices on the bus.
+    /// The configuration can be saved to EEPROM using save_config_to_eeprom()
+    /// and recalled using recall_config_from_eeprom().
+    /// The alarms consist of two bytes: TH and TL registers. They can be used
+    /// to trigger an alarm when the temperature exceeds the set limits or to
+    /// store user data if the alarm functionality is not used.
+    pub fn write_config(self: *const Self, args: struct { alarms: ?Alarms = null, resolution: ?Resolution = null, target: ?RomCode = null }) !void {
```
tact1m4n3, line 148: Having args as structs isn't the zig way of doing it. `target` could be passed in init as an arg and then it can be stored in the struct. For the other args (configuration parameters) I suggest you make a named struct `Config` or smth. Also the config can be taken as a param and applied in init (after device detection). You can still keep this function if the user wants to apply a new config.

drivers/sensor/DS18B20.zig
```diff
@@ -0,0 +1,390 @@
...
+        try self.write_byte(th);
+        try self.write_byte(tl);
+        try self.write_byte(new_config);
+    }
+
+    /// Reads the configuration from the device.
+    /// An error is returned if the device is not present.
+    /// If target is null, the command is sent to all devices on the bus.
+    /// Returns the alarms (TH and TL registers) and resolution.
+    pub fn read_config(self: *const Self, args: struct { target: ?RomCode = null }) !struct { alarms: Alarms, resolution: Resolution } {
```
tact1m4n3: If you store the target in the struct, then you don't need this arg. For the return type, you can use the same struct `Config` as for write_config.

drivers/sensor/DS18B20.zig
```diff
@@ -0,0 +1,390 @@
...
+    const CMD_SKIP_ROM = 0xCC;
+    // const CMD_ALARM_SEARCH = 0xEC; // not implemented
+
+    // function commands
+    const CMD_CONVERT_T = 0x44;
+    const CMD_WRITE_SCRATCHPAD = 0x4E;
+    const CMD_READ_SCRATCHPAD = 0xBE;
+    const CMD_COPY_SCRATCHPAD = 0x48;
+    const CMD_REACLL_SCRATCHPAD = 0xB8;
+    const CMD_READ_POWER_SUPPLY = 0xB4;
```
tact1m4n3: Maybe this can be an enum `Command`. To retrieve the number you can use `@intFromEnum`
```zig
const Command = enum(u8) {
    convert_t = 0x44,
    write_scratchpad = 0x4E,
    // ...
}
```

drivers/sensor/DS18B20.zig
```diff
@@ -0,0 +1,390 @@
...
+    clock_dev: mdf.base.Clock_Device,
+
+    // timing constants
+    const T_RESET = 480;
+    const T_PRESENCE_WAIT = 70;
+    const T_SLOT = 70;
+    const T_WRITE_1 = 10;
+    const T_WRITE_0 = 60;
+    const T_READ_SAMPLE = 10;
+    const T_RECOVER = 1;
```
tact1m4n3: I assume these are in microseconds? You could postfix these with `_US` for clarity (eg: `T_RESET_US`)

drivers/sensor/DS18B20.zig
```diff
@@ -0,0 +1,390 @@
...
+        }
+
+        fn fromInt(value: u2) Resolution {
+            return switch (value) {
+                0 => .HalfDegree9,
+                1 => .QuarterDegree10,
+                2 => .EightDegree11,
+                3 => .SixteenthDegree12,
+            };
+        }
```
tact1m4n3, line 70: This isn't needed. You can use `@enumFromInt`

drivers/sensor/DS18B20.zig
```diff
@@ -0,0 +1,390 @@
...
+        External,
+    };
+
+    pub const RomCode = struct {
+        value: [8]u8,
+    };
+
+    pub const Error = error{
+        DeviceNotPresent,
+        RomCrcMismatch,
```
tact1m4n3: enum tags usually use snake_case

drivers/sensor/DS18B20.zig
```diff
@@ -0,0 +1,390 @@
...
+    /// initiate_temperature_conversion().
+    /// If resolution is provided, it is used to calculate the temperature.
+    /// Otherwise, the resolution is read from the sensor.
+    /// Returns temperature in Celsius.
+    /// An error is returned if the device is not present.
+    pub fn read_temperature(self: *const Self, args: struct { target: ?RomCode = null, resolution: ?Resolution = null }) !f32 {
+        if (try self.reset() == false) return Error.DeviceNotPresent;
+
+        try self.select_target(args.target);
+        try self.write_byte(CMD_READ_SCRATCHPAD);
```
tact1m4n3: If you make an enum for commands, you can make a function that writes a command and use it like this `try self.write_command(.read_scratchpad)`

tact1m4n3:
To fix the formatting CI error you can run `zig fmt [+ the file you want to format]`

kaans:
> Nice work! I added some comments with some things you can improve.

Thanks a lot for the detailed review! I will go through your points and resolve them.

drivers/sensor/DS18B20.zig
kaans: Good point, they are in microseconds

drivers/sensor/DS18B20.zig
kaans: When creating the driver, I had a multi-sensor setup in mind. Thus, it is necessary to be able to address an individual sensor for each write or read operation. When passing the target in the init function, the whole driver can only communicate with one sensor. Creating multiple instances of the driver, e.g. one for each sensor, could be problematic because they use the same gpio pin but may be called form different threads, thus, leading to a collision on the bus. How would you account for that, or how would you pass the target for each function individually?

kaans: I chose the anonymous struct because it allows me to set defaults for optional parameters and I only have to set the parameters I actually want to pass (e.g. not pass the target if I want to either address all sensors or a single sensor). Also, I would not need the struct otherwise except for passing the parameters. But still, I am new to zig, so I am rather curious what would be the advantage of having a named struct like Config? Are there other ways to pass optional arguments to functions with default values?

To me it made sense to call the functions like this: `try ds18b20.write_config(.{ .resolution = .SixteenthDegree12 })`

drivers/sensor/DS18B20.zig
kaans: The reset is a mandatory part of the communication protocol with the DS18B20. It must be issued before any new communication and additionally allows the master to sense whether the addressed device(s) is/are present. 

Taken from the [datasheet page 10, "Initialization"](https://www.analog.com/media/en/technical-documentation/data-sheets/ds18b20.pdf):

`All transactions on the 1-Wire bus begin with an initialization sequence. The initialization sequence consists of a reset pulse transmitted by the bus master followed by presence pulse(s) transmitted by the slave(s). The presence pulse lets the bus master know that slave devices (such as the DS18B20) are on the bus and are ready to operate`

drivers/sensor/DS18B20.zig
kaans: See answers to `write_config`.

drivers/sensor/DS18B20.zig
Grazfather: Can you add a comment explaining this (you can do it more tersely, even just saying that they datasheet says to.

drivers/sensor/DS18B20.zig
Grazfather: Nonetheless, could you define a type for this to return instead of defining it inline?

drivers/sensor/DS18B20.zig
```diff
@@ -0,0 +1,392 @@
...
+        _ = try self.read_byte(); // temp MSB
+        const th = try self.read_byte(); // TH register
+        const tl = try self.read_byte(); // TL register
+        const config = try self.read_byte();
+
+        const resolution_bits: u2 = @intCast((config >> 5) & 0x03);
+
+        return .{
+            .alarms = .{ .th = th, .tl = tl },
+            .resolution = Resolution.fromInt(resolution_bits),
```
Grazfather:
```suggestion
            .resolution = .fromInt(resolution_bits),
```

drivers/sensor/DS18B20.zig
```diff
@@ -0,0 +1,392 @@
...
+        try self.select_target(args.target);
+        try self.write_command(.read_scratchpad);
+
+        _ = try self.read_byte(); // temp LSB
+        _ = try self.read_byte(); // temp MSB
+        const th = try self.read_byte(); // TH register
+        const tl = try self.read_byte(); // TL register
+        const config = try self.read_byte();
+
+        const resolution_bits: u2 = @intCast((config >> 5) & 0x03);
```
Grazfather: could just `@truncate`

examples/raspberrypi/rp2xxx/src/ds18b20.zig
```diff
@@ -0,0 +1,49 @@
...
+    var ds18b20_gpio = rp2xxx.drivers.GPIO_Device.init(pins.ds18b20);
+    const clock_device = rp2xxx.drivers.clock_device();
+
+    const ds18b20 = try DS18B20.init(ds18b20_gpio.digital_io(), clock_device);
+
+    // set desired resolution
+    try ds18b20.write_config(.{ .resolution = .sixteenth_degree_12 });
+
+    // read rom code of connected sensor (assuming only one sensor is connected)
+    // the tom code can be used to address individual devices in a multi drop environment
```
Grazfather:
```suggestion
    // the rom code can be used to address individual devices in a multi drop environment
```

kaans:
Thanks for your comments. I am still not sure if there is a better way to pass the config values to the functions with optional parameters. Creating a struct for the parameters for each function feels unnecessary, but still I might not know about a better mechanism zig offers.

drivers/sensor/DS18B20.zig
tact1m4n3: > The reset is a mandatory part of the communication protocol with the DS18B20. It must be issued before any new communication and additionally allows the master to sense whether the addressed device(s) is/are present.
> 
> Taken from the [datasheet page 10, "Initialization"](https://www.analog.com/media/en/technical-documentation/data-sheets/ds18b20.pdf):
> 
> `All transactions on the 1-Wire bus begin with an initialization sequence. The initialization sequence consists of a reset pulse transmitted by the bus master followed by presence pulse(s) transmitted by the slave(s). The presence pulse lets the bus master know that slave devices (such as the DS18B20) are on the bus and are ready to operate`

Didn't know that. It's ok then :)

drivers/sensor/DS18B20.zig
tact1m4n3: I see your point. What do you think of having `set_target`? And functions would use the last set target (or the one set in init). Zig doesn't have default values for arguments so structs with defaults is the only way to accomplish it.

tact1m4n3: My suggestion would be to have a set_target function, this way you don't take the parameter each time. But otherwise status quo seems ok. 

> what would be the advantage of having a named struct like Config?

Only that you don't repeat yourself and maybe the user can use that type. If you don't use that type multiple times and there is no reason for the user to use it it's fine to inline it.

tact1m4n3: For return values you usually want to define it anyways because the user might want to store it somewhere

tact1m4n3:
It feels a bit strange that you take the optional target parameter for every function. An alternative would be to remove it (this way most functions have no parameter) and instead you can have a `target` field. In zig, fields are public always so you don't require a setter if the only it does is set a field.

```zig
driver.target = something;
driver.write_config(.{ config but no target }); // can be the same struct returned from read_config
driver.initiate_temperature_conversion();
// etc
```

PR #757, opened 2025-12-08
regz: Standard format for chip properties

tact1m4n3:
* Rename `chip.properties` to `chip.raw_properties`
* Type of `chip.properties` is defined in `regz/src/properties.zig` which is both imported by regz and copied into the final chip file
* A side effect of this and my previous PR is that if you try to enable hard float on a target that doesn't have an FPU (like rp2040) you get a nice compile error. We can also do something like this in the future for VTOR on cortex_m0plus that, afaik, is optional. If VTOR is not available, you shouldn't be able to compile a ram image or set the ram_vector_table option.

## Generated code for RP2350
```zig
pub const Properties = struct {
    has_vtor: ?bool = null,
    has_mpu: ?bool = null,
    has_fpu: ?bool = null,
    interrupt_priority_bits: ?u8 = null, // this was called nvic_prio_bits but I tried to make it sound more platform independent
};

pub const Interrupt = struct {
    name: [:0]const u8,
    index: i16,
    description: ?[:0]const u8,
};

pub const properties: Properties = .{
    .has_vtor = true,
    .has_mpu = true,
    .has_fpu = true,
    .interrupt_priority_bits = 4,
};

pub const raw_properties = struct {
    pub const @"cpu.deviceNumInterrupts" = "52";
    pub const @"cpu.dspPresent" = "1";
    pub const @"cpu.endian" = "little";
    pub const @"cpu.fpuPresent" = "true";
    pub const @"cpu.mpuPresent" = "true";
    pub const @"cpu.name" = "CM33";
    pub const @"cpu.nvicPrioBits" = "4";
    pub const @"cpu.revision" = "r1p0";
    pub const @"cpu.sauNumRegions" = "8";
    pub const @"cpu.vendorSystickConfig" = "false";
    pub const @"cpu.vtorPresent" = "1";
    pub const license =
        \\
        \\        Copyright (c) 2024 Raspberry Pi Ltd.
        \\
        \\        SPDX-License-Identifier: BSD-3-Clause
        \\    
    ;
};
```

### DIFF
--- a/tools/regz/src/svd.zig
+++ b/tools/regz/src/svd.zig
@@ -138,11 +138,7 @@ pub fn load_into_db(db: *Database, doc: xml.Doc) !void {
 
         for (optional_properties) |property| {
             if (cpu.get_value(property)) |value| {
-                const property_name = try std.mem.join(
-                    arena.allocator(),
-                    ".",
-                    &.{ "cpu", property }
-                );
+                const property_name = try std.mem.join(arena.allocator(), ".", &.{ "cpu", property });
                 try db.add_device_property(device_id, .{
                     .key = property_name,
                     .value = value,


core/src/cpus/cortex_m.zig
```diff
@@ -737,23 +737,13 @@ pub const startup_logic = struct {
                 : .{ .memory = true, .r0 = true, .r1 = true });
         }
 
-        if (fpu_present and microzig.options.cpu.enable_fpu) {
+        if (microzig.options.cpu.enable_fpu and has_fpu) {
```
tact1m4n3, line 740: If `enable_fpu` is true, but it is uncertain if the chip has an fpu or not (has_fpu is null), this will automatically raise a compile error. Though we don't want this error if `microzig.options.cpu.enable_fpu` is false. This is why I changed the order.

Grazfather: other than with lazy evaluation, I don't see how the order matters (though this makes it a bit more clear)

mattnite (approved):
Hell yeah

PR #760, opened 2025-12-11
Rework USB driver

piotrfila:
This PR is my (yet another) shot at a better USB driver implementation.

Key changes:
- Multi-function (composite) devices possible
- Descriptors are automatically constructed at compile time by drivers
- More type safety around descriptors
- Supporting other USB controllers as well as multiple drivers per controller (eg. interrupted/polled) and writing new drivers should be much easier
- Removed a lot of global variables

This is a succesor to #658, but this time I tried to change as little controller logic as possible to avoid bugs.
As last time, I'm open to suggestions.

There is still much work to be done in this area, but this PR is already big. Nice additions would include:
- Better way of constructing HID descriptors
- Using the new types in stm32 examples
- More and better examples (eg. actual HID keyboard/mouse/gamepad, mass storage)

### DIFF
--- a/core/src/core/usb.zig
+++ b/core/src/core/usb.zig
@@ -73,9 +73,9 @@ pub fn DeviceController(config: Config) type {
     std.debug.assert(config.configurations.len == 1);
 
     return struct {
+        const config0 = config.configurations[0];
         const driver_fields = @typeInfo(config0.Drivers).@"struct".fields;
         const DriverEnum = std.meta.FieldEnum(config0.Drivers);
-        const config0 = config.configurations[0];
         const config_descriptor = blk: {
             var num_interfaces = 0;
             var num_strings = 4;
@@ -212,7 +212,7 @@ pub fn DeviceController(config: Config) type {
             }
         }
 
-        /// Called by the device implementation when a packet has been received.
+        /// Called by the device implementation when a packet has been sent or received.
         pub fn on_buffer(self: *@This(), device_itf: *DeviceInterface, ep: types.Endpoint, buffer: []u8) void {
             if (config.debug) log.info("buff status", .{});
 


arkadiuszwojcik:
@piotrfila I will try take a look on this over the weekend. Going back to `memcpy` issue from previous thread, what is your plan for this? I think other usb libraries expose `memcpy` operation as part of driver interface. That would require change in `FIFO` to accept custom memcpy as dependency right?

piotrfila:
I plan to adress that in a separate PR, this one is already pretty big. I just wanted to get the descriptors out of the way first.

Despite the large diffs in the core/usb.zig and rp2xxx/hal/usb.zig the logic is largely unchanged there.

Do you mean the FIFO in the cdc driver? I think that could be retired entirely in favor of std reader/writer, but I have yet to write any code relating to that.

core/src/core/usb.zig
```diff
@@ -1,680 +1,406 @@
...
-            return S.started;
-        }
+pub const ack: []const u8 = "";
+pub const nak: ?[]const u8 = null;
+
+/// USB Device interface
+/// Any device implementation used with DeviceController must implement those functions
+pub const DeviceInterface = struct {
+    pub const VTable = struct {
+        start_tx: *const fn (this: *DeviceInterface, ep_num: types.Endpoint.Num, buffer: []const u8) void,
```
arkadiuszwojcik: By convention in zig and this codebase we should use 'self' intead of 'this' for name of first parameter. Please change that in other places as well.

core/src/core/usb.zig
```diff
@@ -1,680 +1,406 @@
...
-            const DeviceRequestProcessor = struct {
-                fn process_setup_request(setup: *const types.SetupPacket) !void {
-                    switch (setup.request_type.type) {
-                        .Class => {
-                            //const itfIndex = setup.index & 0x00ff;
-                            std.log.info("Device.Class", .{});
+    return struct {
+        const driver_fields = @typeInfo(config0.Drivers).@"struct".fields;
+        const DriverEnum = std.meta.FieldEnum(config0.Drivers);
+        const config0 = config.configurations[0];
```
arkadiuszwojcik: For compiler order might not be important, but can this line be moved to top of the struct for better top-down read clarity?

core/src/core/usb.zig
```diff
@@ -1,680 +1,406 @@
...
+            }
+        }
 
-                    if (BosConfig.try_get_desc_as(types.ConfigurationDescriptor, curr_bos_cfg)) |_| {
-                        curr_bos_cfg = BosConfig.get_desc_next(curr_bos_cfg);
-                    } else {
-                        // TODO - error
-                        return;
+        /// Called by the device implementation when a packet has been received.
+        pub fn on_buffer(this: *@This(), device_itf: *DeviceInterface, ep: types.Endpoint, buffer: []u8) void {
```
arkadiuszwojcik: how about `on_buffer_rx` name?

piotrfila: Not sure how I missed it writing the comment, but in the current implementation this also gets called after a packet has been sent, so it's not just rx. I'll fix the comment.

core/src/core/usb.zig
piotrfila: it's the result of calling @This() so it obviously should be called this 🙃
jokes aside I totally agree with the argument for consistency

core/src/core/usb.zig
piotrfila: totally agree, mb

piotrfila: totally agree, mb

arkadiuszwojcik:
I made some quick tests, and current behavior is not stable both on RP2040 and RP2350. Here is picture from my terminal and RP2040 test. It also freezes after some time (In my case after 33). For RP2350 output is even more messier. How does it look on your side?

usb_driver

piotrfila:
Someone reported it in the discord as well. That bug is also present in the main branch (to be specific the main branch does not work at all - you have to go a few commits back). I'm not quite sure where the bug is, but I suspect the fifo implementation. I'd rather just rewrite the cdc driver and usb device code in a separate PR.

arkadiuszwojcik:
Sorry, I haven’t checked the code from main for a long time. Yes, it looks like the code doesn’t work at all.

@mattnite @Grazfather, what do you think about it? This code is not 100% correct yet, but it works better than the current code in master (some past PR had to break it). The solution is also much cleaner than the current USB driver code, so it’s a welcome improvement and a step in the right direction.

port/raspberrypi/rp2xxx/src/hal/usb.zig
```diff
@@ -243,19 +299,22 @@ pub fn F(comptime config: UsbConfig) type {
...
             // doesn't bother.
             // TODO: assert!(buffer.len() <= 64);
             // You should only be calling this on IN endpoints.
             // TODO: assert!(UsbDir::of_endpoint_addr(ep.descriptor.endpoint_address) == UsbDir::In);
 
-            const ep = hardware_endpoint_get_by_address(ep_addr);
+            const ep = self.hardware_endpoint_get_by_address(.in(ep_num));
             // wait for controller to give processor ownership of the buffer before writing it.
-            while (ep.buffer_control.?.read().AVAILABLE_0 == 1) {}
+            // while (ep.buffer_control.?.read().AVAILABLE_0 == 1) {}
```
Kytezign, line 317: Curious why this is commented out.  From my previous testing with large, fast writes this could be the source of jumbled data.  The controller needs to give back ownership of the buffer before we should write it again or weird things could happen.

arkadiuszwojcik: @Kytezign you are right. I just tested this PR with this line uncommented and everything works fine on RP2040 and RP2350. @piotrfila can you fix it please.

piotrfila: Dunno, that line was introduced by one of the recent commits that broke things and makes no logical sense to me. It's not okay to just hang execution, we may need to answer on other endpoints. Unsurprisingly, uncommenting this hangs the pico and makes the cdc example stop working.

In the first place, I think the intention was to only call start_tx when you know there is an available buffer. AFAIK the code never worked this way though.

Kytezign: Ya, that's fair point.  Eventually if the buffer is not being emptied by the host we either have to drop data or stall.  And there are likely cases where the host is never going to empty the buffer so stall is not a good option.
I agree, it probably needs to be handle before we get to the start_tx.

At start_tx it's already too late.  Maybe it's better to drop the new data is Available_0 = 1 at this point?

arkadiuszwojcik: First of all would be good to know why this line when uncommented fixes CDC example on my PC and breaks things on your machine and vice versa. Are you using Linux? I am using Windows in my tests.

piotrfila: After some investigation, the only place when data is sent when the buffer is not available is inside of CdcClassDriver's write_flush() function. I still blame the circular buffer, I may try to remove it this weekend.

mattnite:
@arkadiuszwojcik @piotrfila  that works for me, I don't know enough about USB yet to review this properly. I'm happy to merge improvements that are imperfect.

mattnite (approved):

port/raspberrypi/rp2xxx/src/hal/usb.zig
piotrfila: @arkadiuszwojcik Yup, I'm on Linux and adding the line breaks the example for me. This would point towards usb code being the issue rather than the circular buffer... I'll investigate within a couple days.

piotrfila: @arkadiuszwojcik Yup, I'm on Linux and adding the line breaks the example for me. This would point towards usb code being the issue rather than the circular buffer... I'll investigate within a couple days.

PR #764, opened 2025-12-14
esp32c3: correctly use END or STOP for I2C reads and writes

andreacoradi:
This fixes a problem about UnknownAborts (actually Error.ExecutionIncomplete)  on 100KHz I2C.
By looking at the TRM there's a clear distinction between single and multi command sequences, and the HAL now reflects that.
In particular, END command should only be given if we expect more commands to follow, and STOP only when all is done.

### DIFF
--- a/port/espressif/esp/src/hal/i2c.zig
+++ b/port/espressif/esp/src/hal/i2c.zig
@@ -541,6 +541,7 @@ pub const I2C = struct {
         self.reset_fifo();
         self.reset_command_list();
 
+        // See TRM sections 28.5.5.1 and 28.5.8.1 to understand the command structure
         var cmd_idx: usize = 0;
         if (start)
             try self.add_cmd(&cmd_idx, Command.start);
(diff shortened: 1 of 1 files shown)


port/espressif/esp/src/hal/i2c.zig
```diff
@@ -546,29 +538,27 @@ pub const I2C = struct {
...
         var cmd_idx: usize = 0;
-
         if (start)
             try self.add_cmd(&cmd_idx, Command.start);
 
         try self.setup_read(addr, buffer, start, will_continue, &cmd_idx);
 
-        try self.add_cmd(&cmd_idx, Command.end);
+        if (stop) {
+            try self.add_cmd(&cmd_idx, Command.stop);
```
Grazfather: These might be nicer as

```zig
try self.add_cmd(&cmd_idx, if (stop) Command.stop else Command.end);
```

Grazfather (approved):
I can't test for a few weeks, but looks good. If you could add comments referencing the part of the TRM where you figure this out, that could help us out in the future.

port/espressif/esp/src/hal/i2c.zig
andreacoradi: Yep, you are right

andreacoradi:
> I can't test for a few weeks, but looks good. If you could add comments referencing the part of the TRM where you figure this out, that could help us out in the future.

I saw this mostly from the examples in the TRM: section 28.5.1.1 and 28.5.4.1 for writes, 28.5.5.1 and 28.5.8.1 for reads.
EDIT: oh you meant to add them in code. brb

PR #763, opened 2025-12-13
Fix STM32 example compilation issue

mathk:

PR #765, opened 2025-12-14
cortex m3: Define MMFSR,BFSR,UFSR for better error handling

Grazfather:
Needed to compile stm32 target. I don't know how CI was passing without this.


Grazfather:
Oops. Handled by #763

PR #768, opened 2025-12-15
Add systick low level abstraction and add clock device for stm32

mathk:


port/stmicro/stm32/src/hals/common/systick.zig
```diff
@@ -0,0 +1,85 @@
...
+const MAX_COUNTER = 0xFFFFFF;
+
+// For now we need a clock source that is less than 16Mhz so
+// the counter could hold value for an interupt every second.
+// We will choose the best clock source for that.
+// HAL should provide both value.
+// See: https://developer.arm.com/documentation/dui0497/a/cortex-m0-peripherals/optional-system-timer--systick/systick-control-and-status-register
+pub fn init() SystickError!void {
+    const use_processor_clk = if (hal.get_sys_clk() > MAX_COUNTER) false else true;
+    const sys_freq = if (use_processor_clk) hal.get_sys_clk() else hal.get_systick_clk();
```
Grazfather, line 23: so this would require the hals have these two functions. Probably not a terrible requirement for cortex chips, but at the moment they don't. We could start a discussion (on discord, but also file an issue). This PR could maybe have this part live in the stm32 hal until we get buy-in for the other HALs.

mathk: I have move it to the STM32 hal for now

examples/stmicro/stm32/src/stm32f303/hts221.zig
```diff
@@ -1,26 +1,26 @@
...
-    while (i < 8_000_000) {
-        microzig.cpu.nop();
-        i += 1;
-    }
-}
-
 pub const microzig_options: microzig.Options = .{
     .logFn = microzig.hal.uart.log,
+    .interrupts = .{
+        .SysTick = .{ .c = systick.SYSTICK_handler },
```
Grazfather: nitpicky, but maybe better to call this `SysTick_handler` to match the interrupt field

mathk: Done

PR #769, opened 2025-12-15
Add dsp extension to rp2350 (arm)

piotrfila:
This extension is present, as per section 3.7.2 of the datasheet.
I also ran some instructions from this extension with inline assembly and no illegal instruction exception was generated.


mattnite:
Great addition!

Grazfather (approved):

PR #770, opened 2025-12-16
Make STM32 HAL and board more align and push more peripheral to common

mathk:
folder


port/stmicro/stm32/src/boards/STM32F3DISCOVERY.zig
```diff
@@ -4,29 +4,21 @@ pub const microzig = @import("microzig");
 
 pub const hal = microzig.hal;
 pub const rcc = hal.rcc;
-pub const pins = hal.pins;
+pub const hal_pins = hal.pins;
```
Grazfather: why this change?

port/stmicro/stm32/src/boards/STM32F3DISCOVERY.zig
```diff
@@ -53,3 +45,19 @@ pub fn init_log() void {
         microzig.hal.uart.init_logger(.USART1, logger);
     }
 }
+
+pub fn i2c1() hal.i2c.I2C_Device {
+    _ = (hal.pins.GlobalConfiguration{
```
Grazfather, line 49: still using `hal.pins` instead of `hal_pins` here. I would rather we either keep `const pins` or use `hal.pins`.

port/stmicro/stm32/src/boards/STM32L476DISCOVERY.zig
```diff
@@ -191,3 +200,36 @@ pub const Lcd = struct {
         return .{ .current_ram = .{} };
     }
 };
+
+pub fn i2c1() hal.i2c.I2C_Device {
+    _ = (hal.pins.GlobalConfiguration{
```
Grazfather, line 205: same concern about hal_pins

port/stmicro/stm32/src/hals/STM32F103.zig
```diff
@@ -0,0 +1,27 @@
...
+pub const timer = @import("./common/timer_v1.zig");
+pub const usb = @import("./STM32F103/usb.zig");
+pub const adc = @import("./STM32F103/adc.zig");
+pub const crc = @import("./STM32F103/crc.zig");
+pub const power = @import("./STM32F103/power.zig");
+pub const backup = @import("./STM32F103/backup.zig");
+pub const rtc = @import("./STM32F103/rtc.zig");
+pub const dma = @import("./STM32F103/DMA.zig");
+pub const time = @import("./STM32F103/time.zig");
+const util = @import("./common//util.zig");
```
Grazfather:
```suggestion
const util = @import("./common/util.zig");
```

port/stmicro/stm32/src/hals/STM32F103.zig
```diff
@@ -0,0 +1,27 @@
...
+
+//temporary solution
```
Grazfather, line 21: what is the longer term solution? please add TODO

port/stmicro/stm32/src/hals/common/i2c_v2.zig
```diff
@@ -0,0 +1,264 @@
...
+    fn comput_low_time(t_presc: f32) ConfigError!u8 {
+        const real_scll = ((TimingSpec_Standard.t_low - TimingSpec_Standard.t_min_af - 2 * t_presc) / t_presc) - 1;
+        if (real_scll > 256.0) {
+            std.log.info("Clock low time is too short consider lowering the i2cclk frequency", .{});
+            return ConfigError.PCLKOverflow;
+        }
+        return @as(u8, @intFromFloat(@round(real_scll)));
+    }
+
+    fn comput_high_time(t_presc: f32) ConfigError!u8 {
```
Grazfather: This file was largely just copied over, right? skipping reviewing this, but noticed this typo in the function name. MIght as well fix it.

port/stmicro/stm32/src/hals/common/i2c_v2.zig
```diff
@@ -0,0 +1,264 @@
...
+        if (real_scdel > 15.0) {
+            std.log.info("Setup time is too short consider lowering the i2cclk frequency", .{});
+            return ConfigError.PCLKOverflow;
+        }
+        const scdel = @as(u4, @intFromFloat(@round(real_scdel)));
+        // TODO check fi max value is less the 1 and return an error
+        return if (scdel == 0) 1 else scdel;
+    }
+
+    fn comput_low_time(t_presc: f32) ConfigError!u8 {
```
Grazfather: compute is consistently spelled wrong in here :)

port/stmicro/stm32/src/hals/STM32F103.zig
mathk: TBH I have no idea. It was copied from the previous hal.zig file. Maybe we want a way to optin instead of wiring  all the time this handler?

RecursiveError: > what is the longer term solution? please add TODO

The tim API allows you to choose any GPTimer to be a clockwall, but for that, it needs to have a default handlers for all interrupts from those timers (which may or may not exist).

It's possible to completely remove it if time counting with SysStick works well, since both depend on interrupts anyway.

mathk: The Systick clock have not  set the handler by default. It need to be optin using microzig_options or by other mean...

mattnite:
@RecursiveError are the changes to clock here compatible with the stuff you're doing?

RecursiveError:
@mattnite Yes, it’s basically the same approach I used on the F1: a global variable holding the clock values.

That said, there’s really no way for this to be “incompatible”. The clock tree is fully self-contained and already computes, at comptime, all the information needed to configure the clocks correctly. At that point it’s just a matter of checking flags and configuration options — the HAL is free to do pretty much whatever it wants on top of that.

There’s no shared RCC configuration yet, but that’s not really an issue either. If you look at the clock tree tests in [lib.zig,](https://github.com/RecursiveError/microzig/blob/c60b7cd8e87404030ea8151e17a6969f01963e47/port/stmicro/stm32/stm32-clocks/lib.zig#L21924) you’ll see that peripherals tend to have basically the same names across different clock trees anyway.

PR #771, opened 2025-12-19
Added a few helper APIs for the RP2040 pwm

Kytezign:
Added a few APIs that were useful to me using the PWM


port/raspberrypi/rp2xxx/src/hal/pwm.zig
```diff
@@ -57,6 +63,16 @@ pub const Slice = enum(u32) {
     }
 };
 
+/// returns PWM associated with the pin number
```
Grazfather:
```suggestion
/// Return PWM associated with the pin number.
```

PR #773, opened 2025-12-22
Don't distinguish between Regz generated code for "standalone"

### LINKED ISSUE #740: regz: depends on microzig when generating standalone
The generated `VectorTable` struct depends on `microzig.interrupt.Handler` and `microzig.interrupt.unhandled`.

```zig
pub const VectorTable = extern struct {
    const Handler = microzig.interrupt.Handler;
    const unhandled = microzig.interrupt.unhandled;
    // ...
};
```

mattnite:
Fixes #740.

We had a flag to make Regz generate code "for MicroZig". The point was to be able to output code that isn't tightly coupled to MicroZig, however the coupling is quite weak.

Instead I've removed that edge case so the imports for generated code always looks the same. To help those trying to use Regz output for a standalone project, I wrote a quick guide on what to implement for the `microzig` import.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -465,7 +465,6 @@ pub fn MicroBuild(port_select: PortSelect) type {
                 .atdf, .svd => |file| blk: {
                     const regz_run = b.addRunArtifact(regz_exe);
 
-                    regz_run.addArg("--microzig");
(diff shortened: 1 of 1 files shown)


Grazfather:
Nice!

PR #775, opened 2025-12-23
Make public app declarations available from root

### LINKED ISSUE #465: Make the user's app root module of the firmware
Currently the root module of the firmware is `core/start.zig`. This has one upside and one downside. The upside is that we can export the startup logic without requiring the user to do anything. However this can be solved by requiring the user to call a function at comptime like this:
```zig

comptime {
    microzig.cpu.export_startup_logic(); // would export startup logic
}

pub fn main() {

}
```
The downside is that the user loses the ability to control the `root` module, which is often used to configure other dependencies (even std). This has been worked around for `std` but for other deps there is currently no solution that I am aware of.

If the user's module is the root module, then you can specify all the configuration you want directly, like you would do for a normal zig program.

mattnite:
Fixes #465.

One of the weaknesses of MicroZig's build system is that packages that use root declarations for comptime-configuration don't work. We make the application a package to the real root of an executable, and break out useful stdlib options (namely logging) through `microzig_options`. This is also notable because MicroZig changes the defaults for logging because logging will cause compiler errors from not having an operating system.

I want to avoid boiler-plate like the following:

```zig

comptime {
    microzig.cpu.export_startup_logic(); // would export startup logic
}

pub fn main() {
    // ...
}
```

So this patch creates a code generation step in our build system that takes the existing `startup.zig` code that currently makes up the root of the executable and appends the public declarations from the application. This gives us everything we want, no boiler plate, and packages will now be able to use comptime-configuration.

The special handling of the Standard Library logging function is temporary. This needs to be fixed upstream so that the regular logging function is a no-op on freestanding targets.

### DIFF
--- a/core/src/generate_root.zig
+++ b/core/src/generate_root.zig
@@ -20,8 +20,7 @@ pub fn main() !void {
     try writer.writeAll(@embedFile("start.zig"));
 
     inline for (@typeInfo(app).@"struct".decls) |decl| {
-        if (!std.mem.eql(u8, "main", decl.name) and
-            !std.mem.eql(u8, "panic", decl.name) and
+        if (!std.mem.eql(u8, "panic", decl.name) and
             !std.mem.eql(u8, "std_options", decl.name))
         {
             try writer.print("pub const {f} = app.{f};\n", .{ std.zig.fmtId(decl.name), std.zig.fmtId(decl.name) });
--- a/core/src/start.zig
+++ b/core/src/start.zig
@@ -59,8 +59,8 @@ export fn microzig_main() noreturn {
     if (!@hasDecl(app, "main"))
         @compileError("The root source file must provide a public function main!");
 
-    const main = @field(app, "main");
-    const info: std.builtin.Type = @typeInfo(@TypeOf(main));
+    const app_main = @field(app, "main");
+    const info: std.builtin.Type = @typeInfo(@TypeOf(app_main));
 
     const invalid_main_msg = "main must be either 'pub fn main() void' or 'pub fn main() !void'.";
     if (info != .@"fn" or info.@"fn".params.len > 0)
@@ -81,7 +81,7 @@ export fn microzig_main() noreturn {
         microzig.hal.init();
 
     if (@typeInfo(return_type) == .error_union) {
-        main() catch |err| {
+        app_main() catch |err| {
             // Although here we could use @errorReturnTrace similar to
             // `std.start` and just dump the trace (without panic), the user
             // might not use logging and have the panic handler just blink an
@@ -109,7 +109,7 @@ export fn microzig_main() noreturn {
             }
         };
     } else {
-        main();
+        app_main();
     }
 
     // Main returned, just hang around here a bit.


core/src/start.zig
```diff
@@ -10,10 +10,25 @@ pub const panic = if (!@hasDecl(app, "panic")) microzig.panic else app.panic;
...
+        std.Options{};
+
+    if (options.logFn != std.log.defaultLog)
+        @compileError("It seems that you're trying to change the stdlib's logFn. Please set this in the microzig_options, we require this so that embedded executables don't give compile errors by default.");
+
+    if (options.log_level != std.log.default_level)
+        @compileError("It seems that you're trying to change the stdlib's log_level. Please set this in the microzig_options.");
+
+    if (options.log_scope_levels.len > 0)
+        @compileError("It seems that you're trying to change the stdlib's log_scop_levels. Please set this in the microzig_options.");
```
tact1m4n3: typo: log_scop_levels

mattnite:
Oh man that took me a minute to rotate things in my head. I don't think that's possible since Microsoft is always specially built for each executable. If I'm wrong here please explain it in more detail (I also just woke up so that may be a factor). I'm going to merge though, because if I'm wrong it should be able to be done in an incremental manner.

tact1m4n3:
Sure. This is something for the future anyways. I don't have a clear idea about how it would work either, but it would require some serious changes to the build system.

core/src/generate_root.zig
```diff
@@ -0,0 +1,32 @@
...
+    defer file.close();
+
+    var buf: [4096]u8 = undefined;
+    var file_writer = file.writer(&buf);
+    const writer = &file_writer.interface;
+
+    try writer.writeAll(@embedFile("start.zig"));
+
+    inline for (@typeInfo(app).@"struct".decls) |decl| {
+        if (!std.mem.eql(u8, "main", decl.name) and
```
Grazfather: why do these three need special handling?

core/src/start.zig
```diff
@@ -10,10 +10,25 @@ pub const panic = if (!@hasDecl(app, "panic")) microzig.panic else app.panic;
...
-    .log_scope_levels = microzig.options.log_scope_levels,
-    .logFn = microzig.options.logFn,
+pub const std_options: std.Options = blk: {
+    var options = if (@hasDecl(app, "std_options"))
+        app.std_options
+    else
+        std.Options{};
+
+    if (options.logFn != std.log.defaultLog)
+        @compileError("It seems that you're trying to change the stdlib's logFn. Please set this in the microzig_options, we require this so that embedded executables don't give compile errors by default.");
```
Grazfather: nit, but can we break up this lines (e.g. by splitting the string and concatenating with `++`?).

core/src/start.zig
```diff
@@ -10,10 +10,25 @@ pub const panic = if (!@hasDecl(app, "panic")) microzig.panic else app.panic;
 // defined. Parts of microzig use the stdlib logging facility and
 // compilations will now fail on freestanding systems that use it but do
 // not explicitly set `root.std_options.logFn`
-pub const std_options: std.Options = .{
-    .log_level = microzig.options.log_level,
-    .log_scope_levels = microzig.options.log_scope_levels,
-    .logFn = microzig.options.logFn,
+pub const std_options: std.Options = blk: {
+    var options = if (@hasDecl(app, "std_options"))
```
Grazfather, line 14: Why do we have `microzig_options` and `std_options`. Why would someone device to set std_options over microzig options? Couldn't we just `@compileError` here if they set `std_options` at all, or is it only the logging stuff that we are concerned with?

mattnite: Microzig options exist for interrupts and a couple other things I think. There are two reasons why there are stdlib options in the Microzig options. First, they needed to be forwarded as public decls on the app wouldn't make it to the public decals of root (this patch fixes that), and secondly the defaults aren't great for embedded. I've started a PR here to fix this and I've got a more detailed explanation there: https://codeberg.org/ziglang/zig/pulls/30576

core/src/start.zig
mattnite: I will do this, been doing a lot of go recently lol

core/src/generate_root.zig
mattnite: Main was because I was worried something might pick it up when we didn't want it to, it probably doesn't need to be there though. Panic and std_options are things we want to change the defaults for. Like having a breakpoint instruction in the panic handler since that's nice for embedded.

PR #781, opened 2025-12-24
Add the ability to create tests which can import microzig.

### LINKED ISSUE #230: Add support for importing MicroZig as a module in unit tests
On a personal project I'm working on, I recently ran into an issue in which either importing MicroZig directly or using something that relies on it inside of a `test` block causes a compilation error. When running using a test step (e.g. `zig build test`), the error is `no module named 'microzig' available within module root`. When running using `zig test`, the error is `no module named 'microzig' available within module test`. The fix for this was adding a line to `build.zig` that imports the MicroZig module within the test step:

`
    const test_run_step = b.step("test", "Run unit tests");

    const driver_tests = b.addTest(.{
        .name = "driver_tests",
        .root_source_file = b.path("source/drivers.zig"),
        .target = b.graph.host,
    });
    driver_tests.root_module.addImport("microzig", firmware.modules.microzig); // adding this line fixed the issue

    const driver_test_run = b.addRunArtifact(driver_tests);
    test_run_step.dependOn(&driver_test_run.step);
`
Since this is slightly unintuitive, it'd be nice to have direct support for this from within the MicroZig framework.

mattnite:
Fixes #230

This is really great, but I need some more eyes on whether this is how we want to do things. It seems that we have the limitation that our tests can no longer live within the same file as our executable. Using the modules directly from the Firmware object have linking errors because we are overloading the linker script (no _start symbol). And if I try to make a new module targetting the host using the same path, we get an error that multiple modules aren't allowed to have the same path.

There might be a way around this that I'm not sure of. Otherwise, perhaps having to separate tests into their own files is an acceptable workaround? I imagine even if we magically fixed this, there would be many cases of linker errors, because of code depending on the embedded linkerscripts.

Another pattern we might prefer is requiring any tests in the embedded executable to be run on the hardware itself. that would mean this patch should be providing a new test runner, and if a user wanted to write unit tests that ran on their PC, then they would have to put them elsewhere, which sounds reasonable to me.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -821,6 +821,83 @@ pub fn MicroBuild(port_select: PortSelect) type {
             pub fn add_object_file(fw: *Firmware, source: LazyPath) void {
                 fw.artifact.addObjectFile(source);
             }
+
+            pub const TestOptions = struct {
+                root_source_file: LazyPath,
+
+                // create module options
+                optimize: std.builtin.OptimizeMode = .Debug,
+                link_libc: ?bool = null,
+                link_libcpp: ?bool = null,
+                single_threaded: ?bool = null,
+                strip: ?bool = null,
+                unwind_tables: ?std.builtin.UnwindTables = null,
+                dwarf_format: ?std.dwarf.Format = null,
+                code_model: std.builtin.CodeModel = .default,
+                stack_protector: ?bool = null,
+                stack_check: ?bool = null,
+                sanitize_c: ?std.zig.SanitizeC = null,
+                sanitize_thread: ?bool = null,
+                fuzz: ?bool = null,
+                valgrind: ?bool = null,
+                pic: ?bool = null,
+                red_zone: ?bool = null,
+                omit_frame_pointer: ?bool = null,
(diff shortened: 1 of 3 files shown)


tact1m4n3:
> And if I try to make a new module targetting the host using the same path, we get an error that multiple modules aren't allowed to have the same path.

What do you mean by this?

Imo, we could provide a custom embedded test runner with microzig (I made a poc a while ago with this and it works pretty well - with error traces and everything). 

For pc ran tests, I think this patch is a fine solution. Is there a reason why you would want to import the full embedded microzig module on pc tests? An alternative could be to make a dummy one just to silent the microzig import error.

mattnite:
I mean that I'm getting a build system error when I try to use the same path as the executable to make a module for the tests.

I'm starting to like the test runner angle. I may have assumed the issue creator was trying to make host-run unit tests, but maybe they were also trying this.

I'll get back to the drawing board tomorrow, I think I've got a better way to do this now.

mattnite:
Closing for now, waiting until after we resolve the root app issue.

PR #795, opened 2025-12-28
fix: STM32 example was broken for f1 as fpuPresent props was missing

mathk:
I think we should enable debug build in the CI  and remove chip with small amount of flash.

### DIFF
--- a/tools/regz/src/embassy.zig
+++ b/tools/regz/src/embassy.zig
@@ -440,6 +440,7 @@ pub fn load_into_db(db: *Database, path: []const u8) !void {
         // are differences between cores that's something the user will have
         // to keep track of.
 
+        var has_fpu = false;
         for (core.interrupts) |interrupt| {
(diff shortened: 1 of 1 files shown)


tools/regz/src/embassy.zig
```diff
@@ -440,20 +440,30 @@ pub fn load_into_db(db: *Database, path: []const u8) !void {
...
+                .key = "cpu.fpuPresent",
+                .value = "true",
+            });
+        } else {
+            try db.add_device_property(device_id, .{
+                .key = "cpu.fpuPresent",
+                .value = "false",
+            });
+        }
+
```
tact1m4n3: nit: you could rewrite this as
```zig
    try db.add_device_property(device_id, .{
        .key = "cpu.fpuPresent",
        .value = if (has_fpu) "true" else "false",
    });
```

tact1m4n3 (approved):

mattnite (approved):

PR #810, opened 2025-12-30
ci: Build/test sorcerer

Grazfather:


.github/workflows/ci.yml
```diff
@@ -98,6 +98,27 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Setup Zig
+        uses: mlugg/setup-zig@v2
+        with:
+          version: ${{ env.ZIG_VERSION }}
+      - name: Build Sorcerer
+        run: zig build
+        working-directory: tools/sorcerer
+      - name: Test Sorcerer
```
Grazfather, line 120: No tests yet exist, but that's fine. This'll do something when they're added.

mattnite: Hey @Grazfather, I think you need to install sdl3 on ubuntu, it's either that or maybe we're running out of memory.

mattnite:
Hey @Grazfather I think you either need to install sdl3 on ubuntu, or maybe the build is OOMing

Grazfather:
@mattnite I tried, no luck. Just removed ubuntu for now.

mattnite (approved):

PR #828, opened 2026-01-04
MSP430 Support

### LINKED ISSUE #813: Add MSP430 port
We now have [texasinstruments-data](https://github.com/ZigEmbeddedGroup/texasinstruments-data) to contain raw XML information about MSP430 microcontrollers and other TI families. All we need is for someone to add a parser to regz to import the information and we can have an MSP430 port.

mattnite:
This patch adds  two chips and boards, the MSP430F5529, MSP430G2553, and thier Launch Pad boards. My first microcontroller class was taught with an MSP430F5529, so this is very full circle for me. I have these boards on order and will test before merging.

There's also a couple more details I need finalized before the merge:

- [x] Generate register code from targetdb
- [x] Investigate correct compiler and ubsan rt configurations
- [x] write blinky example
- [x] Implement memset, memcpy, and abort in assembly
- [x] Set stack on startup
- [x] Create an interrupt vector (fine if it's hardcoded at first)
- [x] Map vector table to memory
- [x] Add PORT hal
- [x] Add watchdog hal
- [x] Add external crystals to board

Fixes #813

### DIFF
--- a/port/texasinstruments/msp430/src/generate.zig
+++ b/port/texasinstruments/msp430/src/generate.zig
@@ -32,8 +32,11 @@ pub fn main() !void {
 fn generate_chips_file(
     allocator: std.mem.Allocator,
     writer: *std.Io.Writer,
-    //chip_files: []const std.json.Parsed(ChipFile),
 ) !void {
     _ = allocator;
     _ = writer;
+
+    // TODO: https://github.com/ZigEmbeddedGroup/microzig/issues/839
+    // Implement Chips.zig generation for MSP430
+    @panic("TODO");
 }


port/texasinstruments/msp430/ld/MSP430.ld
```diff
@@ -0,0 +1,50 @@
...
+  {
+      microzig_heap_start = .;
+      . = ORIGIN(ram0) + LENGTH(ram0);
+      microzig_heap_end   = .;
+  } > ram0
+
+  .isr_vector :
+  {
+    KEEP(*(.isr_vector*))
+  } > isr_vector
```
tact1m4n3, line 6: If this is the only modification to the autogenerated linker script, maybe you can append it after the autogenerated one (similar to how it works on rp2xxx)?

core/src/cpus/msp430.zig
```diff
@@ -0,0 +1,179 @@
...
+        table: [63]interrupt.Handler = [_]interrupt.Handler{undefined} ** 63,
+        RESET: interrupt.Handler,
+    }
+else if (std.mem.eql(u8, config.cpu_name, "MSP430"))
+    extern struct {
+        table: [15]interrupt.Handler = [_]interrupt.Handler{undefined} ** 15,
+        RESET: interrupt.Handler,
+    }
+else
+    @compileError("Chip not supported yet");
```
Grazfather: could format a string here with the cpu_name to make this error slightly easier to debug.

examples/texasinstruments/msp430/build.zig
```diff
@@ -0,0 +1,37 @@
...
+
+    const mz_dep = b.dependency("microzig", .{});
+    const mb = MicroBuild.init(b, mz_dep) orelse return;
+
+    const targets = &.{
+        mb.ports.msp430.boards.@"MSP-EXP430F5529LP",
+        mb.ports.msp430.boards.@"MSP-EXP430G2ET",
+    };
+
+    inline for (targets) |target| {
```
Grazfather, line 19: I know we only have two examples now, but it would be nice to universally support the `example` config to filter what to build.

examples/texasinstruments/msp430/build.zig
```diff
@@ -0,0 +1,37 @@
...
+});
+
+pub fn build(b: *std.Build) void {
+    const optimize = b.standardOptimizeOption(.{});
+
+    const mz_dep = b.dependency("microzig", .{});
+    const mb = MicroBuild.init(b, mz_dep) orelse return;
+
+    const targets = &.{
+        mb.ports.msp430.boards.@"MSP-EXP430F5529LP",
```
Grazfather: shouldn't we name these launchpads?

port/texasinstruments/msp430/src/hal/gpio.zig
```diff
@@ -0,0 +1,72 @@
...
+    output,
+};
+
+pub const Pin = struct {
+    port: u3,
+    num: u3,
+
+    fn get_reg(comptime periph_name: []const u8, comptime register_name: []const u8) *volatile u8 {
+        if (!@hasDecl(peripherals, periph_name))
+            @panic("missing peripheral");
```
Grazfather: could also comptime print a nice string to make this easier to debug.

port/texasinstruments/msp430/src/hal.zig
```diff
@@ -0,0 +1,4 @@
+const microzig = @import("microzig");
+
+pub const gpio = @import("hal/gpio.zig");
+pub const watchdog = @import("hal/watchdog.zig");
```
Grazfather, line 4: Could follow other boards and add a test block under hear that assigns all of these, even if they don't (yet) have tests

Grazfather (approved):
Skipped over the regz/sorcerer stuff. Just had a few ideas for debugging improvements, all optional.

core/src/cpus/msp430.zig
mattnite: no real need to, I plan on generating all of this soon, this is just for the short term

examples/texasinstruments/msp430/build.zig
mattnite: great idea! I'll cut a ticket to add it to utilities

port/texasinstruments/msp430/ld/MSP430.ld
mattnite: I'll try to revert, but for some reason this was the only way I would see the vector show up in my objdumps, it's possible that I missed something

examples/texasinstruments/msp430/build.zig
mattnite: I'm a little stumped on this one, if you search for specific launchpads like "MSP430F5529" launchpad, you get these crytic serial numbers, and they're literally at the top of the webpage, with only a long description as something else to call them: https://www.ti.com/tool/MSP-EXP430G2ET. So I figured, why not call them by the name that will be easiest to search?

But honestly, maybe something like `launch_pad_msp430g2553`?

mattnite: actually wait yeah, we have a `url` field, I'm making the name readable

port/texasinstruments/msp430/ld/MSP430.ld
mattnite: okay it works!

examples/texasinstruments/msp430/build.zig
Grazfather: I actually had searched them too. TI isn't the best at naming!

PR #831, opened 2026-01-04
generate_linker_script: align .data LMA to 4-byte boundary

Grazfather:
The startup code for riscv uses word-aligned loads (The `lw` instruction) to copy `.data` from flash to RAM. When the `.text` section ended at an unaligned address, `.data`'s flash location (LMA) became misaligned, causing hardware faults during startup when copying the `.data` section into RAM.

This was triggered by certain code combinations that caused `.text` to end at addresses like 0x1bc1 instead of 0x1bc4. The unaligned load would fault before any user code executed.

```
riscv64-unknown-elf-objdump -h zig-out/firmware/lana_tny_dma.elf | grep data -A2
  2 .data         00000100  20000000  00001bc1  00003000  2**0
```

Fixed by adding `. = ALIGN(4);` at the end of .text section to ensure the next section's load address in flash is properly aligned for word loads.

### DIFF
--- a/tools/generate_linker_script.zig
+++ b/tools/generate_linker_script.zig
@@ -197,6 +197,7 @@ pub fn main() !void {
             \\
             \\  .data :
             \\  {
+            \\    . = ALIGN(4);
             \\    microzig_data_start = .;
             \\    *(.sdata*)
             \\    *(.data*)
@@ -220,14 +221,17 @@ pub fn main() !void {
         }
 
         try writer.interface.print(
+            \\    . = ALIGN(4);
             \\    microzig_data_end = .;
             \\  }} > {s}
             \\
             \\  .bss (NOLOAD):
             \\  {{
+            \\    . = ALIGN(4);
             \\    microzig_bss_start = .;
             \\    *(.sbss*)
             \\    *(.bss*)
+            \\    . = ALIGN(4);
             \\    microzig_bss_end = .;
             \\  }} > {s}
             \\


tools/generate_linker_script.zig
```diff
@@ -173,6 +173,13 @@ pub fn main() !void {
             );
         }
 
+        // Ensure that we are still aligned so .data is aligned both in SRAM and on flash for the
+        // copy routine.
+        try writer.interface.writeAll(
+            \\    . = ALIGN(4);
+            \\
+        );
+
```
tact1m4n3: I realized while working on the wifi esp code that I added no alignment constraints to linker scripts. We should definitely do so. The way I saw this being done (in esp stuff and other places) is this:
```ld
.data : ALIGN(4)
{
  . = ALIGN(4);
  microzig_data_start = .;
  *(.sdata*)
  *(.data*)
  . = ALIGN(4);
  microzig_data_end = .;
}
```
And the same for bss. Though, I am no expert in linker scripts (why I skipped the alignment in the first place:) )

Grazfather: I sure am not one either. Should we add that align before each section? e.g. could .sdata force .data to be unaligned?

tact1m4n3: > e.g. could .sdata force .data to be unaligned?

Idk if individual sections have alignment constraints.
Here is a fragment of a linker script from esp32c3 from esp-rs. They don't seem to add alignment in between sections.
```ld
.data : ALIGN(4)
{
  _data_start = ABSOLUTE(.);
  . = ALIGN (4);
  *(.sdata .sdata.* .sdata2 .sdata2.*);
  *(.data .data.*);
  *(.data1)
  _data_end = ABSOLUTE(.);
  . = ALIGN(4);
} > RWDATA

/* LMA of .data */
_sidata = LOADADDR(.data);
```

tools/generate_linker_script.zig
```diff
@@ -221,6 +222,7 @@ pub fn main() !void {
 
         try writer.interface.print(
             \\    microzig_data_end = .;
+            \\    . = ALIGN(4);
```
tact1m4n3, line 224: Should we put ALIGN(4) before `microzig_data_end`?

tools/generate_linker_script.zig
```diff
@@ -197,6 +197,7 @@ pub fn main() !void {
             \\
             \\  .data :
```
tact1m4n3, line 198: Also an ALIGN(4) here (after `:`)? This is what I've seen been used but idk if it adds anything extra.

tact1m4n3:
Should we add the same for bss?

tools/generate_linker_script.zig
Grazfather: I am following the lead of the code you posted for esp-rs, where the align is after `_data_end`. I am sure it doesn't matter.

tools/generate_linker_script.zig
Grazfather: I have no idea what this would do. If anything I figure it should go just before it so that the data start symbol and the .data section are both aligned and at the same address.

tools/generate_linker_script.zig
tact1m4n3: I asked both gemini and chat gpt about this and they recommended to put it before :) It does make sense though what they say: microzig_data_end sees the address before the alignment so if the end of the data section is not aligned, the symbol would not be aligned as well.

Grazfather: yeah that makes sense, I just didn't think an end section alignment would matter. I can change it, though.

tact1m4n3: I might be faster if both are aligned since you can use u32's to do the copying

PR #835, opened 2026-01-05
Fix CircularBuffer reporting wrong readable/writable length

Copper280z:
The CircularBuffer struct was previously reporting the wrong readable/writable length, which allowed the user to overwrite valid parts of the buffer. This is the cause of the rp2xxx usb-cdc example failing to print the whole string.

This adds a test which previously failed, now passes, along with the actual fix, which is just to set the .full member variable to true in the write function when the buffer is full.


Copper280z:
My comment about this fixing the rp2xxx cdc example is incorrect, the example still doesn't work with this change.

PR #838, opened 2026-01-07
ch32v: Add SPI HAL with DMA support

Grazfather:
## Summary

Adds a complete SPI HAL implementation for CH32V (specifically CH32V203) with optional DMA support.

**Key Features:**
- Master-only mode (covers 95% of use cases)
- Blocking I/O (always available)
- Optional DMA support with threshold-based automatic selection
- Vectored I/O for zero-copy operations (required for display drivers)
- Compile-time DMA channel validation
- All 4 SPI modes (CPOL/CPHA combinations)
- Configurable baud rate, bit order, data size
- `SPI_Datagram_Device` wrapper for integration with display drivers

**DMA Integration:**
- Threshold-based selection: transfers ≥16 bytes (configurable) use DMA, smaller use polling
- Compile-time channel validation prevents invalid configurations
- Full-duplex DMA (both TX and RX channels required)
- Channel mappings:
  - SPI1_TX: DMA1 Ch3
  - SPI1_RX: DMA1 Ch2
  - SPI2_TX: DMA1 Ch5

**Examples Included:**
1. **spi_loopback.zig**  - Tests all 4 SPI modes with DMA enabled, includes 32-byte pattern to trigger DMA path
2. **spi_flash_w25q.zig**  - W25Q128 flash operations (JEDEC ID, erase, write, read, verify):

    ```bash
    [0.006000] info: W25Q128 SPI Flash Test
    [0.009000] info: ======================
    [0.013000] info:
    [0.015000] info: Initializing SPI with DMA support...
    [0.019000] info: SPI initialized at 4 MHz with DMA
    [0.024000] info:
    [0.026000] info: Test 1: Reading JEDEC ID...
    [0.030000] info:   JEDEC ID: 0xEF4018
    [0.033000] info:   Device: Winbond W25Q128 - VERIFIED ✓
    [0.039000] info:
    [0.040000] info: Test 2: Reading Status Register (polling)...
    [0.046000] info:   Status: 0x00
    [0.049000] info:   BUSY: false
    [0.052000] info:   WEL:  false
    [0.055000] info:
    [0.056000] info: Test 3: Erase/Write/Read Cycle...
    [0.061000] info:   Erasing sector at 0x001000...
    [0.091000] info:   Sector erased
    [0.094000] info:   Writing 256 bytes...
    [0.100000] info:   Page written
    [0.102000] info:   Reading 256 bytes (DMA)...
    [0.108000] info:   Page read
    [0.111000] info:   Verifying data...
[...]

PR #842, opened 2026-01-10
drivers: display: Add ls0xx display driver

Grazfather:
Tested on my wch branch on top of #838

### DIFF
--- a/drivers/display/sharp_memory_lcd.zig
+++ b/drivers/display/sharp_memory_lcd.zig
@@ -17,7 +17,7 @@
 //!
 //! Example usage:
 //! ```zig
-//! const Display = Sharp_Memory_LCD(.{
+//! const Display = SharpMemory_LCD(.{
 //!     .width = 160,
 //!     .height = 68,
 //!     .vcom_mode = .none,  // nice!view doesn't need VCOM
@@ -57,7 +57,6 @@ pub const Config = struct {
     Digital_IO: ?type = null,
 };
 
-// TODO: Use an enum
 /// Sharp Memory LCD command bytes
 const Cmd = enum(u8) {
     Write = 0x01, // Write line data
@@ -79,7 +78,7 @@ pub const Color = enum(u1) {
 };
 
 /// Create a Sharp Memory LCD driver instance
-pub fn Sharp_Memory_LCD(comptime config: Config) type {
+pub fn SharpMemory_LCD(comptime config: Config) type {
     // Validate config at compile time
     if (config.width % 8 != 0) {
         @compileError("LCD width must be a multiple of 8");
--- a/drivers/framework.zig
+++ b/drivers/framework.zig
@@ -16,7 +16,7 @@ pub const display = struct {
     pub const ST7735 = st77xx.ST7735;
     pub const ST7789 = st77xx.ST7789;
     pub const HD44780 = hd44780.HD44780;
-    pub const Sharp_Memory_LCD = sharp_memory_lcd.Sharp_Memory_LCD;
+    pub const SharpMemory_LCD = sharp_memory_lcd.SharpMemory_LCD;
 
     // Export color types:
     pub const colors = @import("display/colors.zig");


drivers/display/sharp_memory_lcd.zig
```diff
@@ -0,0 +1,411 @@
...
+
+                try self.dd.write(&.{ cmd, 0x00 });
+            }
+            // No-op for displays that don't need VCOM
+        }
+    };
+}
+
+/// Generate a framebuffer type for the specified dimensions
+fn Framebuffer_Type(comptime width: comptime_int, comptime height: comptime_int) type {
```
tact1m4n3, line 244: For types such as framebuffers, colors etc. it would be nice to have a common implementation (in the future)

tact1m4n3, line 244: also comptime for arguments isn't necessary because:
1. the function returns a type which means it is only valid at comptime
2. comptime_int can't exist at runtime

drivers/display/sharp_memory_lcd.zig
```diff
@@ -0,0 +1,411 @@
...
+        /// Bytes per line
+        pub const bytes_per_line = width / 8;
+
+        /// Total framebuffer size
+        pub const size_bytes = bytes_per_line * height;
+
+        /// Pixel data in line-major order (matches Sharp display RAM organization)
+        /// Layout: [Line 0: bytes_per_line bytes][Line 1: bytes_per_line bytes]...
+        /// Within each byte: LSB-first (bit 0 = leftmost pixel)
+        pixel_data: [size_bytes]u8,
```
tact1m4n3, line 257: You could maybe use `std.StaticBitSet` for this

drivers/display/sharp_memory_lcd.zig
```diff
@@ -0,0 +1,411 @@
...
+
+    return struct {
+        const Self = @This();
+
+        /// Datagram device for SPI communication
+        dd: config.Datagram_Device,
+        /// Optional display enable pin
+        disp_pin: if (config.Digital_IO) |T| T else void,
+        /// VCOM state (compile-time conditional, zero cost when disabled)
+        vcom_state: if (config.vcom_mode == .software) bool else void,
```
tact1m4n3, line 98: Nice!

drivers/display/sharp_memory_lcd.zig
Grazfather: I agree. Seems like a simple problem, the difficulty is in coming up with an interface that works. #849

Grazfather: Yeah I generally go heavy with `comptime` just to make it clear to the reader but this is over the top :)

drivers/display/sharp_memory_lcd.zig
Grazfather: The issue is that if you don't use it, you need to pass `{}` as an argument, unfortunately.

drivers/display/sharp_memory_lcd.zig
vesim987: std.StaticBitSet by default is using array of `usize`, so endianess will affect it.

drivers/display/sharp_memory_lcd.zig
```diff
@@ -0,0 +1,411 @@
...
+
+        /// Write a single line to the display
+        /// line_num: 0-based line number (will be converted to 1-based for Sharp protocol)
+        /// data: pixel data for the line (must be bytes_per_line bytes)
+        pub fn write_line(self: *Self, line_num: usize, data: *const [bytes_per_line]u8) !void {
+            if (line_num >= config.height) {
+                return error.LineOutOfBounds;
+            }
+            if (data.len != bytes_per_line) {
+                return error.InvalidLineLength;
```
vesim987: This will never trigger

drivers/display/sharp_memory_lcd.zig
```diff
@@ -0,0 +1,411 @@
...
+            // Write each line: [line_addr][line_data][dummy_byte]
+            for (0..num_lines) |i| {
+                const line_num = start_line + i;
+                const line_byte: u8 = @intCast(line_num + 1); // 1-based addressing
+                const line_start = i * bytes_per_line;
+                const line_end = line_start + bytes_per_line;
+                const line_data = data[line_start..line_end];
+
+                // Write: [line_address][pixel_data...][dummy_byte]
+                try self.dd.writev(&.{
```
vesim987: idk if it is worth the effort, but you know the height at comptime, so with some magic you can do everything in a single `writev` call

drivers/display/sharp_memory_lcd.zig
Grazfather: Oops. I meant to fix this but forgot and merged. Will followup with a cleanup PR.

PR #843, opened 2026-01-10
rp2xxx: Fix pins apply semantics

Grazfather:
- Make `pins.apply()` match other HALs, where it returns the pins
  - Update examples as necessary
  - Update docs
- Fix some errors in the adafruit_metro_rp2350 pin config.
  - We were never building this, so the errors weren't caught
- Comment out most of the pin config: With board_blinky it writes all that config, which is both a waste, but also makes the branch evaluator blow up.
- Move `blinky_core1` out of rp2040_only: I tested it on rp2350
- Add adafruit_feather_rp2350 board
- Add `board_blinky` example, which is like `blinky`, but it uses the board's configured GPIO named `led` to toggle.

### DIFF
--- a/examples/raspberrypi/rp2xxx/src/allocator.zig
+++ b/examples/raspberrypi/rp2xxx/src/allocator.zig
@@ -28,7 +28,6 @@ const pin_config = hal.pins.GlobalConfiguration{
     },
 };
 
-
 // ---- UART Configuration --------------------------------
 
 const baud_rate = 115200;
--- a/examples/raspberrypi/rp2xxx/src/rp2350_only/always_on_timer.zig
+++ b/examples/raspberrypi/rp2xxx/src/rp2350_only/always_on_timer.zig
@@ -16,7 +16,6 @@ const pin_config = hal.pins.GlobalConfiguration{
     .GPIO0 = .{ .name = "gpio0", .function = .UART0_TX },
 };
 
-
 pub const microzig_options = microzig.Options{
     .log_level = .debug,
     .logFn = uart.log,
--- a/examples/raspberrypi/rp2xxx/src/rp2350_only/random_data.zig
+++ b/examples/raspberrypi/rp2xxx/src/rp2350_only/random_data.zig
@@ -15,7 +15,6 @@ const pin_config = hal.pins.GlobalConfiguration{
     .GPIO0 = .{ .name = "gpio0", .function = .UART0_TX },
 };
 
-
 pub const microzig_options = microzig.Options{
     .log_level = .debug,
     .logFn = uart.log,


examples/raspberrypi/rp2xxx/src/rp2040_only/rtc.zig
```diff
@@ -33,7 +33,7 @@ fn rtc_isr() callconv(.c) void {
 }
 
 pub fn main() !void {
-    const pins = pin_config.pins();
+    const pins = pin_config.apply();
```
Grazfather, line 36: This one was weird. It was never applying the pin config?

examples/raspberrypi/rp2xxx/src/blinky.zig
```diff
@@ -4,16 +4,14 @@ const rp2xxx = microzig.hal;
 const time = rp2xxx.time;
 
 const pin_config: rp2xxx.pins.GlobalConfiguration = .{
-    .GPIO25 = .{
+    .GPIO7 = .{
```
Grazfather: oops will revert

port/raspberrypi/rp2xxx/src/boards/adafruit_metro_rp2350.zig
```diff
@@ -9,184 +9,184 @@ const hal = microzig.hal;
...
+    // },
+    //
+    // .GPIO19 = .{
+    //     .name = "hstx_d0n",
+    //     .function = .HSTX,
+    // },
+    //
+    // .GPIO20 = .{
+    //     .name = "sda",
+    //     .function = .I2C0_SDA,
```
Grazfather, line 54: Many of these did not set the function correctly

port/raspberrypi/rp2xxx/src/boards/raspberry_pi_pico.zig
```diff
@@ -1,7 +1,18 @@
...
+const hal = microzig.hal;
+const pins = hal.pins;
+
 pub const bootrom = @import("shared/rp2040_bootrom.zig");
 
 comptime {
     _ = bootrom;
 }
+
+pub const pin_config = pins.GlobalConfiguration{
```
tact1m4n3, line 13: This would only be valid for the non wifi version of the pico. Maybe we can have different boards for wifi and non wifi version of the pico in the future

port/raspberrypi/rp2xxx/src/boards/raspberry_pi_pico2.zig
```diff
@@ -1 +1,12 @@
 pub const xosc_freq = 12_000_000;
+
+const microzig = @import("microzig");
+const hal = microzig.hal;
+const pins = hal.pins;
+
+pub const pin_config = pins.GlobalConfiguration{
```
tact1m4n3, line 7: this too

tact1m4n3 (approved):

port/raspberrypi/rp2xxx/src/boards/raspberry_pi_pico.zig
Grazfather: Yes I was thinking exactly that.

Grazfather: There's no real 'risk' at the moment because these configs have to be applied manually by the app.

PR #846, opened 2026-01-10
extract network module from pico WiFi examples

ianic:
Integrating lwip requires adding foundationlibc and lwip dependencies. Putting lwipopts.h and arch/cc.h to the include path, setting options and implementing some functions required by lwip.

To simplify that I extracted everything that is not dependent on the platform into network module. That simplifies using it in applications outside of the rpi/examples/net folder. And opens the path to use lwip on other platforms.

Using WiFi in rpi pico (2)w projects is now simplified to adding dependency to the `build.zig.zon`:
```zig
    .dependencies = .{
        .microzig = .{
            .path = "../../microzig/",
        },
        .net = .{
            .path = "../../microzig/modules/network/",
        },
    },
```
and adding module to the application in `build.zig`:
```zig
const std = @import("std");

const MicroBuild = microzig.MicroBuild(.{
    .rp2xxx = true,
});

pub fn build(b: *std.Build) void {
    const optimize = b.standardOptimizeOption(.{});

    const mz_dep = b.dependency("microzig", .{});
    const target = mb.ports.rp2xxx.boards.raspberrypi.pico2_arm;

    const net_dep = b.dependency("net", .{
        .target = b.resolveTargetQuery(target.zig_target),
        .optimize = optimize,
        .lwip_mem_size = 32 * 1024,
        .lwip_pbuf_pool_size = 32,
    });
    const net_mod = net_dep.module("net");

    const firmware = mb.add_firmware(.{
        .name = app,
        .target = target,
        .optimize = optimize,
        .root_source_file = b.path("src/main.zig"),
        .imports = &.{
            .{ .name = "net", .module = net_mod },
        },
    });
    mb.install_firmware(firmware, .{});
}
```

Lwip has a lot of configurable options. Currently in the build system I opened two of them for configuration: `lwip_mem_size` and `lwip_pbuf_pool_size`. That controls memory allocations; static packet buffers pool size and maximum of heap allocation.

There are two platform dependent functions required by lwip: `lwip_sys_now` and `lwip_rand`. For rpi they are implemented in `hal/drivers.zig`.

### DIFF
--- a/modules/network/link/build.zig.zon
+++ b/modules/network/link/build.zig.zon
@@ -6,6 +6,6 @@
     .paths = .{
         "build.zig",
         "build.zig.zon",
-        "src",
+        "link.zig",
     },
 }


modules/network/build.zig
```diff
@@ -0,0 +1,69 @@
...
+        // Assuming 1500 MTU of layer 3 payload, lwip will add 14 bytes of ethernet
+        // header resulting in max 1514 bytes. PBUF_LINK_ENCAPSULATION_HLEN reserves
+        // 22 bytes *at the front of each packet buffer*. lwip will write ethernet
+        // header at position 22, and leave leading 22 bytes for use in the wifi
+        // link interface.
+        //
+        lwip_mod.addCMacro("MEM_ALIGNMENT", "4"); // 4 bytes, u32 alignment
+        lwip_mod.addCMacro("PBUF_POOL_BUFSIZE", "1540"); // 1500 MTU + headers (14 + 22 + 4)
+        lwip_mod.addCMacro("PBUF_LINK_HLEN", "14"); // ethernet header len
+        lwip_mod.addCMacro("PBUF_LINK_ENCAPSULATION_HLEN", "22"); // WiFi chip header space
```
Grazfather: Isn't this chip dependent? Could we force a client (who knows what chip they are using) to provide this value?

port/raspberrypi/rp2xxx/src/hal/drivers.zig
```diff
@@ -535,3 +535,28 @@ pub const WiFi = struct {
...
+
+/// Time since boot in milliseconds.
+export fn lwip_sys_now() u32 {
+    const ts = hal.time.get_time_since_boot();
+    return @truncate(ts.to_us() / 1000);
+}
+
+var rng: ?hal.rand.Ascon = null;
+
+export fn lwip_rand() u32 {
```
Grazfather: Do we not need to enforce the c calling convention for these?

modules/network/build.zig
ianic: `PBUF_LINK_ENCAPSULATION_HLEN` is cyw43 specific. That is the only chip we currently support. Adding support for some other chip will probably need more configuration. 
We can add options like the one we already have: `lwip_mem_size`.
I didn`t prepare for that right now.

port/raspberrypi/rp2xxx/src/hal/drivers.zig
ianic: It seems that exported functions already have callconv(.c)

piotrfila: Yup, this is what the zig documentation says:
https://ziglang.org/documentation/0.15.2/#Functions
I think it is better to specify it explicitly anyway though.

Grazfather: new to me! but it makes sense.

modules/network/build.zig
Grazfather: This is fine, but you changed the comment from something that mentioned cyw43. If this is specific to that chip, please make that explicit.

examples/raspberrypi/rp2xxx/build.zig
```diff
@@ -145,26 +145,17 @@ pub fn build(b: *std.Build) void {
...
         if (std.mem.indexOf(u8, example.name, "_net-") != null) {
-            const target = b.resolveTargetQuery(firmware.target.zig_target);
-            const foundation_dep = b.dependency("foundationlibc", .{
-                .target = target,
+            const net_dep = b.dependency("net", .{
+                .target = b.resolveTargetQuery(firmware.target.zig_target),
                 .optimize = optimize,
+                // lwip options:
+                //.lwip_mem_size = 32 * 1024,
+                //.lwip_pbuf_pool_size = 32,
```
Grazfather: leftovers?

modules/network/src/root.zig
```diff
@@ -26,13 +26,25 @@ pub const Interface = struct {
...
-        recv: *const fn (*anyopaque, []u8) anyerror!?struct { usize, usize },
-        send: *const fn (*anyopaque, []u8) anyerror!void,
-        ready: *const fn (*anyopaque) bool,
-    },
+        vtable: struct {
+            recv: *const fn (*anyopaque, []u8) anyerror!?struct { usize, usize },
+            send: *const fn (*anyopaque, []u8) anyerror!void,
+        },
+
+        pub fn adapt(any: anytype) Link {
```
Grazfather: Why do you have an adapt method instead of having _cyw43439.zig_ just import the interface and return an instance of it?

modules/network/src/root.zig
```diff
@@ -187,15 +198,24 @@ pub const Interface = struct {
...
+        log.debug("memp tcp seg:        {}", .{stats.memp[lwip.MEMP_TCP_SEG].*});
+        log.debug("memp reassdata:      {}", .{stats.memp[lwip.MEMP_REASSDATA].*});
+        log.debug("memp frag buf:       {}", .{stats.memp[lwip.MEMP_FRAG_PBUF].*});
+        log.debug("memp igmp group:     {}", .{stats.memp[lwip.MEMP_IGMP_GROUP].*});
+        log.debug("memp sys timeout:    {}", .{stats.memp[lwip.MEMP_SYS_TIMEOUT].*});
+        log.debug("memp nd6 queue:      {}", .{stats.memp[lwip.MEMP_ND6_QUEUE].*});
+        log.debug("memp ip6 reassdata:  {}", .{stats.memp[lwip.MEMP_IP6_REASSDATA].*});
+        log.debug("memp mld6 group:     {}", .{stats.memp[lwip.MEMP_MLD6_GROUP].*});
+        log.debug("memp pbuf:           {}", .{stats.memp[lwip.MEMP_PBUF].*});
+        //log.debug("memp max:            {}", .{stats.memp[lwip.MEMP_MAX].*});
```
Grazfather: leftover?

modules/network/src/root.zig
ianic: They are in different projects. Importing interface will require adding dependency on the network module to the drivers project.

drivers/wireless/cyw43439.zig
```diff
@@ -63,11 +63,6 @@ pub fn send_zc(ptr: *anyopaque, bytes: []u8) anyerror!void {
     try self.wifi.send_zc(bytes);
 }
 
-pub fn ready(ptr: *anyopaque) bool {
```
Grazfather, line 66: Was this not used?

modules/network/src/root.zig
Grazfather: But they are for the same interface, right? As in, if you change the interface there, you have to make the change here. That implies a dependency anyway. Could a module shared between the two be created that defines this interface that both sides implement?

port/raspberrypi/rp2xxx/src/hal/drivers.zig
```diff
@@ -535,3 +535,28 @@ pub const WiFi = struct {
...
+        .RP2040 => brk: {
+            if (rng == null) {
+                rng = .init();
+            }
+            var val: u32 = 0;
+            rng.?.fill(std.mem.asBytes(&val));
+            break :brk val;
+        },
+    };
+}
```
tact1m4n3: Personally, I wouldn't include this in the hal since the user might use a different network stack entirely. I think a little more boilerplate here is acceptable.

drivers/wireless/cyw43439.zig
ianic: It was, I removed it [here](https://github.com/ZigEmbeddedGroup/microzig/pull/846/commits/75894a2d47a8b753921154fef6ef9d3e5dac5589) and start signaling not readiness by errors. This way interface is smaller and we can signal different errors; no space int the link buffer OutOfMemory vs link down.

modules/network/src/root.zig
ianic: I made a separate module for the link interface and drivers project now imports that tiny module. Adapting is not needed any more, we have typed interfaces.

modules/network/build.zig
ianic: I made `PBUF_POOL_BUFSIZE` and `PBUF_LINK_ENCAPSULATION_HLEN` configurable and move configuration to the rpi examples build. This makes network module not cyw43 specific.

port/raspberrypi/rp2xxx/src/hal/drivers.zig
ianic: Makes sense, removed.

modules/network/src/root.zig
Grazfather: Cool. I personally like it this way more, but if you like your old way, let me know, we can figure it out.

modules/network/link/build.zig.zon
```diff
@@ -0,0 +1,11 @@
+.{
+    .name = .link,
+    .version = "0.0.0",
+    .fingerprint = 0x36ac99f11956c760,
+    .minimum_zig_version = "0.15.2",
+    .paths = .{
+        "build.zig",
+        "build.zig.zon",
+        "src",
```
tact1m4n3: Make sure to add `link.zig` here instead of src (this solves the ci error)

ianic: Ups, thanks.

mattnite (approved):

PR #848, opened 2026-01-10
usb: Allow partial packet writes, add error checking and documentation

piotrfila:
Changes:
- `usb_start_tx` -> `ep_writev`, now returns number of bytes written
- `usb_start_rx` -> `ep_readv` and `ep_listen`: This makes it easier to make interrupt-safe drivers
- More error checking when constructing descriptors
- First attempt at documenting the usb driver comptime interface
- The drivers now assume ACKs are handled automatically. This is the case for the rp2xxx usb controller but I have no idea if other controllers also handle it that way.
- New driver for rp2xxxx family that supports rebooting and logs through picotool
- HID driver implementation

### DIFF
--- a/core/src/core/usb.zig
+++ b/core/src/core/usb.zig
@@ -484,9 +484,8 @@ pub fn DeviceController(config: Config, driver_args: config.DriverArgs()) type {
                         // device_itf.ep_listen(.ep0, 0);
                         self.tx_slice = null;
 
-                        // None of the drivers so far are using the ACK phase
-                        // if (self.driver_last) |drv|
-                        //     self.driver_class_control(device_itf, drv, .Ack, &self.setup_packet);
+                        // W could call an ACK phase callback here,
+                        // but none of the drivers so far need it.
                     }
                 }
             } else if (comptime ep == types.Endpoint.out(.ep0))


port/raspberrypi/rp2xxx/src/hal/usb.zig
```diff
@@ -274,38 +246,33 @@ pub fn Polled(
...
             const ep = self.hardware_endpoint_get_by_address(.in(ep_num));
-            // Wait for controller to give processor ownership of the buffer before writing it.
-            // This is technically not neccessary, but the usb cdc driver is bugged.
-            while (bufctrl_ptr.read().AVAILABLE_0 == 1) {}
 
-            const len = buffer.len;
+            const len = @min(data[0].len, ep.data_buffer.len);
             switch (chip) {
-                .RP2040 => @memcpy(ep.data_buffer[0..len], buffer[0..len]),
+                .RP2040 => @memcpy(ep.data_buffer[0..len], data[0][0..len]),
```
Copper280z: Is the assumption that the caller will only ever provide a single buffer to write? If so, it's probably worth a comment at the vtable definition and an assert.

piotrfila: I simply forgot to implement that :sweat_smile:

core/src/core/usb/drivers/cdc.zig
```diff
@@ -31,188 +57,206 @@ pub const Options = struct {
...
+            };
         }
 
-        pub fn transfer(self: *@This(), ep: types.Endpoint, data: []u8) void {
-            if (ep == types.Endpoint.out(self.ep_out)) {
-                self.rx.write(data) catch {};
-                self.prep_out_transaction();
-            }
+        pub fn on_rx(self: *@This(), ep_num: EP_Num) void {
+            assert(.ep0 == @atomicLoad(EP_Num, &self.ep_out, .seq_cst));
```
Copper280z: Why assert .ep0? The rx endpoint should never be zero?

core/src/core/usb.zig
```diff
@@ -77,54 +158,57 @@ pub fn DeviceController(config: Config) type {
         const driver_fields = @typeInfo(config0.Drivers).@"struct".fields;
         const DriverEnum = std.meta.FieldEnum(config0.Drivers);
         const config_descriptor = blk: {
-            var num_interfaces = 0;
-            var num_strings = 4;
-            var num_ep_in = 1;
-            var num_ep_out = 1;
+            const max_psize = config.max_supported_packet_size;
+            assert(max_psize == 8 or max_psize == 16 or max_psize == 32 or max_psize == 64);
```
Copper280z: Add 512 bytes for USB-HS.

core/src/core/usb/drivers/cdc.zig
```diff
@@ -31,188 +57,206 @@ pub const Options = struct {
...
-        pub fn write_flush(self: *@This()) usize {
-            if (self.tx.get_readable_len() == 0) {
-                return 0;
-            }
-            const len = self.tx.read(&self.epin_buf);
-            self.device.start_tx(self.ep_in, self.epin_buf[0..len]);
+        pub fn write(self: *@This(), data: []const u8) usize {
+            const len = @min(self.tx_data.len - self.tx_end, data.len);
+
+            if (len == 0) return 0;
```
Copper280z: Don't we want to flush if the buffer is full, ie tx_end=tx_data.len?

core/src/core/usb.zig
piotrfila: Do you know if ep0 also can also have a maximum packet size over 64? Or is it just for ep1-15?

Copper280z: I think endpoint 0 is always a max of 64 for HS, but I haven’t checked the spec.

core/src/core/usb/drivers/cdc.zig
piotrfila: Yup, nice catch!

core/src/core/usb.zig
piotrfila: Ok, for now I'll limit it to 64 just in case

core/src/core/usb/drivers/cdc.zig
piotrfila: It was meant as a flag value, but I agree it's kind of messy to do it that way.
I rewrote that part of the code, I think it's easier to follow now.

port/raspberrypi/rp2xxx/src/hal/usb.zig
```diff
@@ -435,3 +433,61 @@ pub fn Polled(
...
+                        @enumFromInt(0x00),
+                        @enumFromInt(0x01),
+                    ),
+                    .interface_s = alloc.string(interface_str),
+                } } };
+            }
+        };
+
+        pub fn init(self: *@This(), desc: *const Descriptor, device: *usb.DeviceInterface, data: []u8) void {
+            assert(data.len == 0);
```
Grazfather: Why do you want to be provided an empty slice?

examples/raspberrypi/rp2xxx/src/usb_hid.zig
```diff
@@ -62,37 +133,59 @@ pub fn panic(message: []const u8, _: ?*std.builtin.StackTrace, _: ?usize) noretu
...
+                    )
+                else
+                    drivers.keyboard.send_report(&.empty));
+            } else {
                 old = new;
-                led.toggle();
+                idx = 0;
             }
+
+            if (drivers.keyboard.receive_repeort()) |report|
```
Grazfather: Not yours I guess, but typo here `receive_repeort`

port/raspberrypi/rp2xxx/src/hal/usb.zig
```diff
@@ -358,52 +358,49 @@ pub fn Polled(
...
+            peripherals.USB.ADDR_ENDP.write(.{ .ADDRESS = addr });
+        }
 
-            peripherals.USB.ADDR_ENDP.modify(.{ .ADDRESS = addr });
+        fn dpram_memcpy(dst: []u8, src: []const u8) void {
+            assert(dst.len == src.len);
+            switch (chip) {
+                .RP2040 => @memcpy(dst, src),
+                .RP2350 => {
+                    // Could be optimized for aligned data, for now just copy
```
Grazfather, line 376: Can you remind us here why you can't use memcpy on rp2350?

core/src/core/usb/types.zig
```diff
@@ -133,26 +399,67 @@ pub const SetupPacket = extern struct {
     /// conflict.
     request: u8,
```
Grazfather, line 400: Can you use the enum here? Make it non-exhaustive to support extra types

core/src/core/usb/drivers/EchoExample.zig
```diff
@@ -0,0 +1,115 @@
+///! This is an example driver that sends any data received on ep2 to ep1.const std = @import("std");
```
Grazfather:
```suggestion
///! This is an example driver that sends any data received on ep2 to ep1.
```

core/src/core/usb/drivers/EchoExample.zig
```diff
@@ -0,0 +1,115 @@
...
+const std = @import("std");
+const usb = @import("../../usb.zig");
+const assert = std.debug.assert;
+const EP_Num = usb.types.Endpoint.Num;
+const log = std.log.scoped(.usb_echo);
+
+/// The descriptors need to have the same memory layout as the sent data.
```
Grazfather, line 7: This is a note not a doc string, please either convert to `//` or also explain this struct.

core/src/core/usb/drivers/EchoExample.zig
```diff
@@ -0,0 +1,115 @@
...
+    self.tx_ready.store(true, .seq_cst);
+}
+
+pub fn on_rx(self: *@This(), ep_rx: EP_Num) void {
+    var buf: [64]u8 = undefined;
+    // Read incoming packet into a local buffer
+    const len_rx = self.device.ep_readv(ep_rx, &.{&buf});
+    log.info("Received on {t}: {s}", .{ ep_rx, buf[0..len_rx] });
+    // Check if we can transmit
+
```
Grazfather:
```suggestion

    // Check if we can transmit
```

core/src/core/usb/types.zig
piotrfila: I tried doing it this way, but the enum changes based on usb device class, so each driver would have to cast it back into an u8 just to immediately cast into its own enum. This way the cast only happens once.

port/raspberrypi/rp2xxx/src/hal/usb.zig
piotrfila: Yup, done

port/raspberrypi/rp2xxx/src/hal/usb.zig
piotrfila: The slice is passed unconditionally, whether the driver requests it or not. Maybe changing the function signature based on whether data was requested is a better idea.

core/src/core/usb/drivers/EchoExample.zig
piotrfila: Thanks, fixed

core/src/core/usb/drivers/EchoExample.zig
piotrfila: Thanks, fixed

examples/raspberrypi/rp2xxx/src/usb_hid.zig
piotrfila: Very much mine :laughing: fixed

port/raspberrypi/rp2xxx/src/hal/usb.zig
piotrfila: Fixed, slices are now not passed if they were not requested

core/src/core/usb/drivers/EchoExample.zig
piotrfila: Thanks, fixed

core/src/core/usb/drivers/CDC.zig
```diff
@@ -0,0 +1,264 @@
...
+    itf_data: desc.Interface,
+    ep_out: desc.Endpoint,
+    ep_in: desc.Endpoint,
+
+    const Strings = struct {
+        itf_notifi: []const u8 = "",
+        itf_data: []const u8 = "",
+    };
+
+    pub fn create(
```
Grazfather, line 73: Could you please explain in a comment why you're using this pattern? e.g. why do you need `alloc.next_itf`?

core/src/core/usb/drivers/CDC.zig
```diff
@@ -0,0 +1,264 @@
...
+                },
+                .ep_out = .bulk(alloc.next_ep(.Out), max_supported_packet_size),
+                .ep_in = .bulk(alloc.next_ep(.In), max_supported_packet_size),
+            },
+            .alloc_bytes = 2 * max_supported_packet_size,
+        };
+    }
+};
+
+pub const handlers: usb.DriverHadlers(@This()) = .{
```
Grazfather: Typo Hadlers -> Handlers

core/src/core/usb/drivers/CDC.zig
```diff
@@ -0,0 +1,264 @@
...
+    @memcpy(self.tx_data[self.tx_end..][0..len], data[0..len]);
+    self.tx_end += @intCast(len);
+
+    if (self.tx_end == self.tx_data.len)
+        _ = self.flush();
+
+    return len;
+}
+
+/// Returns true if flush operation succeded.
```
Grazfather:
```suggestion
/// Flush tx buffer
/// Returns true if flush operation succeeded.
```
(nit: Doc strings should describe the function, not just a note about it)
Fixed a typo, feel free to reword as you like, or convert to normal comment

core/src/core/usb/drivers/CDC.zig
```diff
@@ -0,0 +1,264 @@
...
+        self.descriptor.ep_in.endpoint.num,
+        &.{self.tx_data[0..self.tx_end]},
+    ));
+    self.tx_end = 0;
+    return true;
+}
+
+pub fn init(self: *@This(), desc: *const Descriptor, device: *usb.DeviceInterface, data: []u8) void {
+    const len_half = @divExact(data.len, 2);
+    assert(len_half == desc.ep_in.max_packet_size.into());
```
Grazfather, line 207: Mostly inconsequential, but why this way? We can avoid a divide by just using this value, and assert that this * 2 = data.len

core/src/core/usb/drivers/CDC.zig
```diff
@@ -0,0 +1,264 @@
...
+        .tx_data = data[len_half..],
+        .tx_end = 0,
+        .tx_ready = .init(true),
+    };
+    device.ep_listen(desc.ep_out.endpoint.num, @intCast(self.rx_data.len));
+}
+
+pub fn class_request(self: *@This(), setup: *const usb.types.SetupPacket) ?[]const u8 {
+    const mgmt_request: ManagementRequestType = @enumFromInt(setup.request);
+    log.debug("cdc setup: {any} {} {}", .{ mgmt_request, setup.length.into(), setup.value.into() });
```
Grazfather, line 234: this could cause a lot of issues if we want to support using cdc as a logger. Maybe (later?) we can compileError if the config logfn is cdc and the log level in anything usb is enabled.

core/src/core/usb/drivers/CDC.zig
```diff
@@ -0,0 +1,264 @@
...
+
+    return switch (mgmt_request) {
+        .SetLineCoding => usb.ack, // we should handle data phase somehow to read sent line_coding
+        .GetLineCoding => std.mem.asBytes(&self.line_coding),
+        .SetControlLineState => blk: {
+            // const DTR_BIT = 1;
+            // self.is_ready = (setup.value & DTR_BIT) != 0;
+            // self.line_state = @intCast(setup.value & 0xFF);
+            break :blk usb.ack;
+        },
```
Grazfather: These needed? Looks like this could just be `.SetControlLineState => usb.ack`

core/src/core/usb/drivers/CDC.zig
Grazfather: I see you have a pretty good comment for EchoExample. Maybe just copy it here?

core/src/core/usb.zig
```diff
@@ -1,47 +1,156 @@
...
+    return @Type(.{ .@"struct" = .{
+        .layout = layout,
+        .backing_integer = BackingInt,
+        .decls = &.{},
+        .fields = fields,
+        .is_tuple = false,
+    } });
+}
+
+const drivers_alias = drivers;
```
Grazfather: Needed?

core/src/core/usb.zig
```diff
@@ -1,47 +1,156 @@
...
+            ret = ret ++ [1]descriptor.String{.from_str(s)};
+        return ret;
+    }
+};
+
+pub fn DescriptorCreateResult(Descriptor: type) type {
+    return struct {
+        descriptor: Descriptor,
+        /// If not null, a slice of this length will be passed
+        /// to dirver's init function.
```
Grazfather:
```suggestion
        /// to driver's init function.
```

core/src/core/usb.zig
```diff
@@ -50,125 +159,278 @@ pub const DeviceInterface = struct {
     }
 };
 
+pub fn EndpointHandler(Driver: type) type {
+    return fn (*Driver, types.Endpoint.Num) void;
+}
+
+pub fn DriverHadlers(Driver: type) type {
```
Grazfather: Hadlers -> Handlers

core/src/core/usb.zig
```diff
@@ -179,226 +441,187 @@ pub fn DeviceController(config: Config) type {
...
-                        }
+                        // Otherwise, we've just finished sending tx_slice.
+                        // We expect an ensuing status phase where the host
+                        // sends us a zero-byte DATA packet via EP0 OUT.
+                        // device_itf.ep_listen(.ep0, 0);
+                        self.tx_slice = null;
+
+                        // None of the drivers so far are using the ACK phase
+                        // if (self.driver_last) |drv|
+                        //     self.driver_class_control(device_itf, drv, .Ack, &self.setup_packet);
```
Grazfather: dead code?

examples/raspberrypi/rp2xxx/src/usb_hid.zig
```diff
@@ -4,55 +4,126 @@ const microzig = @import("microzig");
...
+    .bcd_device = .v1_00,
+    .serial = "someserial",
+    .max_supported_packet_size = USB_Device.max_supported_packet_size,
+    .configurations = &.{.{
+        .attributes = .{ .self_powered = false },
+        .max_current_ma = 50,
+        .Drivers = struct { keyboard: Keyboard, reset: rp2xxx.usb.ResetDriver(null, 0) },
+    }},
+}, .{.{
+    .keyboard = .{ .itf_string = "Boot Keyboard", .poll_interval = 1 },
```
Grazfather, line 124: why 'boot' keyboard?

core/src/core/usb/drivers/CDC.zig
piotrfila: First half of the buffer is used for rx and the second is used for tx

core/src/core/usb/drivers/CDC.zig
piotrfila: Nice catch! Just now I tried disabling it based on logFn, but that seems to trigger a dependency loop. I think ideally we should have most logs turned off by default, but I don't see an easy way to do that right now.

examples/raspberrypi/rp2xxx/src/usb_hid.zig
piotrfila: The "Boot protocol" specifies how the report should be laid out so that host driver support is easier. Interface strings are not really used in practice, it's just to provide something human-readable (lsusb does not seem to show them though)

core/src/core/usb.zig
piotrfila: I wanted to keep it around to mark when the ack phase callback should be called, but that's better done with a comment, fixed.

core/src/core/usb.zig
piotrfila: Nope, thanks

core/src/core/usb/drivers/CDC.zig
piotrfila: Nope, thanks

core/src/core/usb.zig
piotrfila: Thanks, fixed

core/src/core/usb/drivers/CDC.zig
piotrfila: Thanks, fixed

core/src/core/usb/drivers/CDC.zig
piotrfila: Thanks, fixed

core/src/core/usb/drivers/CDC.zig
piotrfila: I copied part of the comment

core/src/core/usb/drivers/CDC.zig
Grazfather: no worries, we will worry about that when we add a cdc logger

core/src/core/usb/drivers/CDC.zig
Grazfather: yeah, I meant that you could assert that `desc.ep_in.max_packet_size.into() * 2 == data.len`

examples/raspberrypi/rp2xxx/src/usb_hid.zig
Grazfather: TIL

PR #850, opened 2026-01-11
drivers: Overhaul README

Grazfather:
This is mostly AI generated! I gave it a general structure, had it generate the docs, then manually checked its work, re-worded, and reformated.

### DIFF
--- a/drivers/README.md
+++ b/drivers/README.md
@@ -554,7 +554,7 @@ var gpio_pin = hal.gpio.Pin.init(0, 13); // PA13
 
 // Create LED driver specialized for concrete Pin type
 const LED = LED_Driver(.{
-    .Digital_IO = hal.gpio.Pin,  // Concrete type!
+    .Digital_IO = hal.drivers.GPIO_Device, // Concrete type!
     .active_high = true,
 });
 var led = try LED.init(gpio_pin);  // Pass concrete pin directly


drivers/README.md
```diff
@@ -1,26 +1,1266 @@
...
+
+```zig
+const hal = microzig.hal;
+
+// Get GPIO pin (concrete type, no vtable)
+var gpio_pin = hal.gpio.Pin.init(0, 13); // PA13
+
+// Create LED driver specialized for concrete Pin type
+const LED = LED_Driver(.{
+    .Digital_IO = hal.gpio.Pin,  // Concrete type!
```
tact1m4n3: Isn't this supposed to be: `hal.drivers.GPIO_Device`?

drivers/README.md
```diff
@@ -1,26 +1,1266 @@
...
+
+- **Wireless**
   - [ ] CYW43 (WIP)
   - [ ] [SX1276, SX1278](https://github.com/ZigEmbeddedGroup/microzig/issues/248)
-- Stepper
+
+- **Stepper Motors**
   - [x] A4988
   - [x] DRV8825 (Implemented but untested)
+  - [x] ULN2003
```
tact1m4n3: Can we move this to the top?

drivers/README.md
Grazfather: No, the point is that instead of providing the interface, where calls have an extra layer of indirection via the vtable, these configurable drivers let you provide a concrete type directly. As long as this type is method-compatible, then the calls will be made directly. So in this sample driver, if a hal implemented `read`, `write`, `set_bias` and `set_direction`, then it would compile with direct calls to the methods.

drivers/README.md
Grazfather: done

drivers/README.md
tact1m4n3: I was asking if `hal.gpio.Pin` does implement those exact functions? Afaik for hals, in drivers.zig, there is a `GPIO_Device` that implements the driver `Digital_IO` comptime interface

Grazfather: Oh right. `hal.gpio.Pin` _could_ (I never specified which hal :) ). You're right though, we have those wrappers already that do it without vtable

Though there is still some indirection here! Maybe some HALs should actually make their Pins conform to the interface! The calls are static at least, so they wouldn't mess with any prediction.

Fixed.

PR #855, opened 2026-01-15
Migrate Clock_Device to be align with Zig interface

mathk:


port/espressif/esp/src/hal/drivers.zig
```diff
@@ -448,22 +448,21 @@ pub const GPIO_Device = struct {
     }
 };
 
-///
-/// Implementation of a `Clock_Device` that uses the HAL's `time` module.
-///
-pub fn clock_device() Clock_Device {
-    const S = struct {
-        const vtable: Clock_Device.VTable = .{
+const clock: struct {
```
mathk: At some point I would see this in the time file instead of here. For Clock_Device it is not a big deal since we rarely need the base object to implement the vtable functions.

drivers/base/Clock_Device.zig
```diff
@@ -9,61 +9,80 @@ const mdf = @import("../framework.zig");
...
-
 /// Virtual table for the digital i/o functions.
 vtable: *const VTable,
 
-/// API
-pub fn is_reached(td: Clock_Device, time: mdf.time.Absolute) bool {
+/// Make this object using:
+/// `const timeout = clock.make_timeout(Duration.from_us(100))`
+///
+/// This is handy when you need to communicate with peripherale
```
Grazfather, line 18:
```suggestion
/// This is handy when you need to communicate with peripherals
```

drivers/base/Clock_Device.zig
```diff
@@ -9,61 +9,80 @@ const mdf = @import("../framework.zig");
...
     return time.is_reached_by(now);
 }
 
-pub fn make_timeout(td: Clock_Device, timeout: mdf.time.Duration) mdf.time.Absolute {
-    return @as(mdf.time.Absolute, @enumFromInt(td.get_time_since_boot().to_us() + timeout.to_us()));
+/// API
+pub fn make_timeout(td: *Clock_Device, timeout: mdf.time.Duration) Timeout {
+    return .{
+        .clock = td,
+        .time = @as(mdf.time.Absolute, @enumFromInt(td.get_time_since_boot().to_us() + timeout.to_us())),
```
Grazfather, line 42: I think you can do `.from_us(td.get_time_since_boot().to_us()).add_duration(timeout)`

drivers/base/Clock_Device.zig
```diff
@@ -9,61 +9,80 @@ const mdf = @import("../framework.zig");
...
+    };
 }
 
-pub fn make_timeout_us(td: Clock_Device, timeout_us: u64) mdf.time.Absolute {
-    return @as(mdf.time.Absolute, @enumFromInt(td.get_time_since_boot().to_us() + timeout_us));
+pub fn make_timeout_us(td: *Clock_Device, timeout_us: u64) Timeout {
```
Grazfather, line 46: I think that we should remove this.

You can just do
`ct.make_timeout(.from_us(...))`

Grazfather (commented):
Gave it a once over, want to review more closely tomorrow.

drivers/base/Clock_Device.zig
Grazfather: actually just `td.get_time_since_boot().add_duration(timeout)`

drivers/base/Clock_Device.zig
mathk: You mean removing the function? I hesitate to remove as I didn't understand why having both. I thought of keeping it to avoid too many breaking change. But I am happy to remove it.

drivers/base/Clock_Device.zig
```diff
@@ -9,61 +9,80 @@ const mdf = @import("../framework.zig");
...
 const Clock_Device = @This();
 
-/// Pointer to the object implementing the driver.
-///
-/// If the implementation requires no `ptr` pointer,
-/// you can safely use `undefined` here.
-ptr: *anyopaque,
-
 /// Virtual table for the digital i/o functions.
 vtable: *const VTable,
```
piotrfila: Maybe store the vtable inline? It's just two pointers so the pointer does not save much memory but we'd avoid a level of indirection.

drivers/base/Clock_Device.zig
Grazfather: I'm happy to break things. As long as we keep the examples working. We warn right in the readme:
> NOTE: This is in development; breaks in the API are bound to happen.

mathk:
Will close for now as we do not need more than vtable. I might salvage the Timeout struct.

PR #856, opened 2026-01-15
Zig master (0.16.0) Update

mattnite:


tools/esp-image/src/elf2image.zig
```diff
@@ -11,88 +10,173 @@ var elf_file_reader_buf: [1024]u8 = undefined;
...
+    \\--max-rev-full <u16>                       Maximal chip revision (in format: major * 100 + minor)
+    \\--dont-append-digest                       Don't append a SHA256 digest of the entire image after the checksum
+    \\--flash-freq <FlashFreq>                   SPI Flash frequency
+    \\--flash-mode <FlashMode>                   SPI Flash mode
+    \\--flash-size <FlashSize>                   SPI Flash size in megabytes
+    \\--flash-mmu-page-size <FlashMMU_PageSize>  Flash MMU page size
+    \\--use-segments                             Use program headers instead of section headers
+    \\<str>
+    \\
+;
```
tact1m4n3, line 127: Rewrote arg parsing to not depend on clap that isn't up to date with master. We should use the std cli parser when it lands

core/src/failing.zig
```diff
@@ -0,0 +1,931 @@
...
+const Dir = Io.Dir;
+const File = Io.File;
+const CancelProtection = Io.CancelProtection;
+const Timeout = Io.Timeout;
+const Group = Io.Group;
+const Cancelable = Io.Cancelable;
+const AnyFuture = std.Io.AnyFuture;
+const net = Io.net;
+
+// Taken from https://codeberg.org/ziglang/zig/pulls/30691/ until it lands.
```
tact1m4n3: We should switch to `Io.failing` when it lands

der-teufel-programming:
`Io.failing` is available in `0.16.0` so this PR could use that

mattnite:
I know others have been working towards this separately, this was a way back so I'm killing it.

PR #861, opened 2026-01-16
esp: Wi-Fi support

tact1m4n3:


core/src/interrupt.zig
```diff
@@ -122,9 +122,10 @@ pub const CriticalSectionMutex = struct {
 
     /// Unlocks the mutex.
     pub fn unlock(self: *CriticalSectionMutex) void {
-        if (self.critical_section) |cs| {
+        const maybe_cs = self.critical_section;
+        self.critical_section = null;
+        if (maybe_cs) |cs| {
```
tact1m4n3, line 127: Bug fix

examples/espressif/esp/src/tcp_server.zig
```diff
@@ -0,0 +1,224 @@
...
+};
+
+const SERVER_PORT = 3333;
+
+var maybe_netif: ?*lwip.c.netif = null;
+
+var ip_ready_semaphore: rtos.Semaphore = .init(0, 1);
+var ip: lwip.c.ip_addr_t = undefined;
+
+extern fn netconn_new_with_proto_and_callback(t: lwip.c.enum_netconn_type, proto: lwip.c.u8_t, callback: ?*const anyopaque) [*c]lwip.c.struct_netconn;
```
tact1m4n3, line 61: translate c output for lwip has some issues

port/espressif/esp/ld/esp32_c3/direct_boot_sections.ld
```diff
@@ -5,48 +5,88 @@ SECTIONS
...
+    . = ALIGN(16);
     microzig_bss_start = .;
-    *(.bss*)
     *(.sbss*)
+    *(.bss*)
+    . = ALIGN(16);
     microzig_bss_end = .;
   } > DRAM
 
+  .tbss (NOLOAD) :
```
tact1m4n3: Foundation libc adds something as a threadlocal variable. This is a workaround to avoid a linker error (because of the heap section). I plan to add support for thread local variables in the rtos.

port/espressif/esp/src/cpus/esp_riscv.zig
```diff
@@ -226,10 +247,32 @@ pub const interrupt = struct {
...
+                    (@as(u61, INTERRUPT_CORE0.INTR_STATUS_REG_1.raw) << 32),
+            };
+        }
+
+        pub fn is_set(status: Status, source: Source) bool {
+            return status.reg & (@as(u61, 1) << @intFromEnum(source)) != 0;
+        }
+    };
+
+    pub fn expect_handler(comptime int: Interrupt, comptime expected_handler: InterruptHandler) void {
```
tact1m4n3, line 274: This is an interesting pattern to use in other hals as well

port/espressif/esp/src/hal/radio/osi.zig
```diff
@@ -0,0 +1,1217 @@
...
+
+    @export(&malloc, .{ .name = "malloc", .linkage = .weak });
+    @export(&calloc, .{ .name = "calloc", .linkage = .weak });
+    @export(&free, .{ .name = "free", .linkage = .weak });
+    @export(&puts, .{ .name = "puts", .linkage = .weak });
+
+    @export(&gettimeofday, .{ .name = "gettimeofday", .linkage = .weak });
+    @export(&sleep, .{ .name = "sleep", .linkage = .weak });
+    @export(&usleep, .{ .name = "usleep", .linkage = .weak });
+    @export(&vTaskDelay, .{ .name = "vTaskDelay", .linkage = .weak });
```
tact1m4n3, line 182: These can be overwritten by the user. They are provided in case there is no available implementation. But they are required for esp_wifi drivers to work

build.zig
```diff
@@ -45,12 +44,10 @@ const exe_targets: []const std.Target.Query = &.{
 };
 
 pub fn build(b: *Build) void {
-    const optimize = b.standardOptimizeOption(.{});
-
     const generate_linker_script_mod = b.createModule(.{
         .root_source_file = b.path("tools/generate_linker_script.zig"),
         .target = b.graph.host,
-        .optimize = optimize,
+        .optimize = .ReleaseSafe,
```
tact1m4n3: Prevent recompilation of generate_linker_script when the optimization mode changes.

port/espressif/esp/ld/esp32_c3/image_boot_sections.ld
```diff
@@ -42,36 +42,85 @@ SECTIONS
...
+
   .ram_data_dummy (NOLOAD) :
   {
     . = ALIGN(ALIGNOF(.ram.text)) + SIZEOF(.ram.text);
+    . = ALIGN(ALIGNOF(.ram.text.wifi)) + SIZEOF(.ram.text.wifi);
   } > DRAM
 
   .data :
   {
+    . = ALIGN(16);
```
tact1m4n3, line 86: I also encountered some alignment caused crashes.

tact1m4n3:
I managed to crash the formatter :) The problematic file is `esp/cpus/esp_riscv.zig`. I also tested the formatter from master and that crashes as well.

tools/printer/src/root.zig
```diff
@@ -97,7 +97,7 @@ fn output_source_line(
     out_tty_config: std.io.tty.Config,
     src_loc: DebugInfo.ResolvedSourceLocation,
 ) !void {
-    var dir = try std.fs.cwd().openDir(src_loc.dir_path, .{});
+    var dir = std.fs.cwd().openDir(src_loc.dir_path, .{}) catch return;
```
Grazfather, line 100: why?

port/espressif/esp/ld/esp32_c3/image_boot_sections.ld
Grazfather: I get littering the linker with align directives, but do you actually know for sure that these need align 16, or are you just going this high to be safe?

port/espressif/esp/src/hal/rtos.zig
```diff
@@ -0,0 +1,1117 @@
...
+const time = microzig.drivers.time;
+const rtos_options = microzig.options.hal.rtos;
+pub const Priority = rtos_options.Priority;
+
+const get_time_since_boot = @import("time.zig").get_time_since_boot;
+const system = @import("system.zig");
+const systimer = @import("systimer.zig");
+
+// How it works?
+// For simple task to task context switches, only necessary registers are
```
Grazfather: what are 'necessary'?

port/espressif/esp/src/hal/rtos.zig
```diff
@@ -0,0 +1,1117 @@
...
+// saved. But if a higher priority task becomes available during the handling
+// of an interrupt, a task switch is forced by saving the entire state of the
+// task on the stack. What is interresting is that the two context switches are
```
Grazfather:
```suggestion
// task on the stack. What is interesting is that the two context switches are
```

port/espressif/esp/src/hal/rtos.zig
```diff
@@ -0,0 +1,1117 @@
...
+    }
+    make_ready(&idle_task);
+
+    microzig.cpu.interrupt.map(rtos_options.cpu_interrupt.source(), rtos_options.yield_interrupt);
+    microzig.cpu.interrupt.set_type(rtos_options.yield_interrupt, .level);
+    microzig.cpu.interrupt.set_priority(rtos_options.yield_interrupt, .lowest);
+    microzig.cpu.interrupt.enable(rtos_options.yield_interrupt);
+
+    // unit0 is already enabled as it is used by `hal.time`.
+    if (rtos_options.systimer_unit != .unit0) {
```
Grazfather, line 118: is it OK to do this? should it be a compile error?

port/espressif/esp/src/hal/rtos.zig
```diff
@@ -0,0 +1,1117 @@
...
+
+    return .{ current_task, next_task };
+}
+
+pub fn sleep(duration: time.Duration) void {
+    const timeout: TimerTicks = .after(duration);
+    while (!timeout.is_reached()) {
+        yield(.{ .wait = .{
+            .timeout = timeout,
+        } });
```
Grazfather: I am betting zig fmt will complain about this

port/espressif/esp/src/hal/rtos.zig
```diff
@@ -0,0 +1,1117 @@
...
+        sem.current_value += 1;
+        if (sem.current_value > sem.max_value) {
+            sem.current_value = sem.max_value;
+        } else {
+            sem.wait_queue.wake_one();
+        }
+    }
+};
+
+pub const TypeErasedQueue = struct {
```
Grazfather, line 922: this and the normal queue could maybe go into some utility library? Just something about the rtos state would have to be generalized.

If you think it would fit, a TODO is fine by me.

examples/espressif/esp/src/tcp_server.zig
```diff
@@ -0,0 +1,228 @@
...
+        },
+        .radio = .{
+            .wifi = .{
+                .on_packet_received = on_packet_received,
+            },
+        },
+    },
+};
+
+const SSID = "SSID";
```
Grazfather, line 51: Maybe add a comment here and a compile error if it's not set

tools/printer/src/root.zig
tact1m4n3: We should ignore a source location if we can't find the file. We already do this for the file, this also ignores it if the directory that the file should be in doesn't exist.

port/espressif/esp/src/hal/rtos.zig
tact1m4n3: If the user overrides the systimer_unit to be something other than the unit used by `esp.time` which is already enabled, we should enable it. Otherwise, enabling the unit is unnecessary.

port/espressif/esp/src/hal/rtos.zig
tact1m4n3: Only the registers that the compiler wants to save. This happens through the clobbers from the inline asm context_switch.

port/espressif/esp/ld/esp32_c3/image_boot_sections.ld
tact1m4n3: I used 16 because I got a warning from the linker that .data should be aligned to 16 bytes.

port/espressif/esp/src/hal/radio/osi.zig
```diff
@@ -0,0 +1,1275 @@
...
+const get_time_since_boot = @import("../time.zig").get_time_since_boot;
+const timer = @import("timer.zig");
+const wifi = @import("wifi.zig");
+
+const log = std.log.scoped(.esp_radio_osi);
+
+// TODO: config
+const coex_enabled: bool = false;
+
+pub var gpa: std.mem.Allocator = undefined;
```
Grazfather, line 26: nit but GPA implies general purpose allocator. Do they normally call it GPA unless it's actually that (or the debug allocator now)?

port/espressif/esp/src/hal/radio/timer.zig
```diff
@@ -0,0 +1,178 @@
...
+) void {
+    {
+        mutex.lock();
+        defer mutex.unlock();
+
+        if (find(ets_timer)) |tim| {
+            tim.deadline = .init_relative(get_time_since_boot(), duration);
+            tim.periodic = if (repeat) duration else null;
+        } else {
+            log.warn("timer not found based on ets_timer", .{});
```
Grazfather: Can we actually survive these? Should we be panicking if it's game over here rather than failing later on?

port/espressif/esp/src/hal/radio/timer.zig
```diff
@@ -0,0 +1,178 @@
...
+}
+
+pub fn disarm(ets_timer: *c.ets_timer) void {
+    mutex.lock();
+    defer mutex.unlock();
+
+    if (find(ets_timer)) |tim| {
+        tim.deadline = .no_deadline;
+    } else {
+        log.warn("timer not found based on ets_timer", .{});
```
Grazfather: For better debugging, would be nice to print which function failed (arm vs done vs disarm)

examples/espressif/esp/src/tcp_server.zig
tact1m4n3: I'll add a comment. If we add a compile error, wouldn't CI fail?

port/espressif/esp/src/hal/rtos.zig
tact1m4n3: Well this queue sort of comes with the rtos. The implementation is made specifically for it. Though, it would be nice to allow the user to swap the RTOS for wifi.

examples/espressif/esp/src/tcp_server.zig
Grazfather: ah good point.

port/espressif/esp/src/hal/rtos.zig
Grazfather: ah ok, that's fine by me, then

port/espressif/esp/src/hal/rtos.zig
Grazfather: Yeah I mean to say that you should be more specific in the comment. It's defined by the ABI?

port/espressif/esp/src/hal/radio/osi.zig
tact1m4n3: Afaik, it is a good practice to name allocators gpa or arena to imply what kind of allocator do you need to provide. If it is an arena, you expect it to leak memory.

port/espressif/esp/src/hal/rtos.zig
tact1m4n3: Is it better now?

port/espressif/esp/src/hal/radio/osi.zig
Grazfather: Cool. If that's typical, that works.

port/espressif/esp/src/hal/radio/timer.zig
```diff
@@ -0,0 +1,178 @@
...
+
+pub fn done(gpa: std.mem.Allocator, ets_timer: *c.ets_timer) void {
+    mutex.lock();
+    defer mutex.unlock();
+
+    if (find(ets_timer)) |tim| {
+        timer_list.remove(&tim.node);
+        gpa.destroy(tim);
+    } else {
+        log.warn("timer not found based on ets_timer", .{});
```
Grazfather:
```suggestion
        log.warn("done: timer not found based on ets_timer", .{});
```

Grazfather (approved):

PR #862, opened 2026-01-16
Introduce Timeout object for clock device

mathk:


drivers/led/ws2812.zig
```diff
@@ -63,7 +62,7 @@ pub fn WS2812(options: struct {
 
             try self.dev.write(self.buffer[0..i]);
 
-            self.next_write_time = self.clock_dev.make_timeout_us(300);
+            self.next_write_time = self.clock_dev.make_timeout(300);
```
Grazfather: wouldn't this need `.from_us`?

drivers/base/Clock_Device.zig
```diff
@@ -116,17 +130,10 @@ test Test_Device {
     ttd.elapse_time(2);
     try std.testing.expectEqual(2, td.get_time_since_boot().to_us());
 
-    try std.testing.expect(!td.is_reached(@enumFromInt(4)));
```
mattnite, line 119: Why are these assertions deleted?

mathk: I did remove `is_reached` from the public interface. There is now test around the Timeout which is the new public interface.

mattnite: Do not remove it from the public API. The original design is deadline based, this timeout mechanism should just be a "nice to have" @ikskuh could use your opinion here

mathk: The reason I wanted a more encapsulated Absolute was that it can came from different clock source. You can not really mix different Absolute time if they come from different clock device. The Timeout has the advantage that you can't really use it with a different clock (which would make no sens). I am fine if you think this is not an issue.

mattnite: I don't think this is an issue

drivers/base/Clock_Device.zig
```diff
@@ -18,18 +18,32 @@ ptr: *anyopaque,
...
     return time.is_reached_by(now);
 }
 
-pub fn make_timeout(td: Clock_Device, timeout: mdf.time.Duration) mdf.time.Absolute {
-    return @as(mdf.time.Absolute, @enumFromInt(td.get_time_since_boot().to_us() + timeout.to_us()));
-}
-
-pub fn make_timeout_us(td: Clock_Device, timeout_us: u64) mdf.time.Absolute {
-    return @as(mdf.time.Absolute, @enumFromInt(td.get_time_since_boot().to_us() + timeout_us));
+/// API
```
mattnite: Please remove this doc comment, it doesn't mean anything

PR #865, opened 2026-01-18
feat: use riscv atomic immediate instructions for interrupt enable/disable

kibels:
The existing Csr.write(), Csr.set() and Csr.clear() functions only allow the "r" register constraint, so always generate as two instructions e.g.
```asm
li a0, 31
csrs mstatus, a0
```
This commit adds new functions, Csr.write_imm(), Csr.set_imm(), and Csr.clear_imm() which also allow the "i" immediate constraint on inputs, which generates only one instruction.
```asm
csrsi mstatus, 31
```
These are made separate interfaces because unfortunately when using comptime values >31, the compiler isn't smart enough to put them in a register and just gives an error that they exceed the 5-bit immediate field(llvm backend issue, as it also happens with C and C++).
For comptime values >31, the original interfaces must be used instead.

Fortunately, the RISCV ISA is designed such that the interrupt enable/disable bits always fit in the immediate field so they can be set/cleared atomically.

If this is too niche to accept the interface bloat, feel free to reject this PR, just something I was toying around with after noticing these functions were generating more instructions than necessary.

The atomicity might matter for things like critical sections or kernel context switches, but maybe those should use dedicated functions/assembly instead of these general purpose interfaces.

Godbolt where I was testing this out: https://godbolt.org/z/nv1doEnjz

### DIFF
--- a/modules/riscv32-common/src/riscv32_common.zig
+++ b/modules/riscv32-common/src/riscv32_common.zig
@@ -33,11 +33,11 @@ pub const interrupt = struct {
     }
 
     pub fn enable_interrupts() void {
-        csr.mstatus.set(.{ .mie = 1 });
+        csr.mstatus.set_imm(.{ .mie = 1 });
     }
 
     pub fn disable_interrupts() void {
-        csr.mstatus.clear(.{ .mie = 1 });
+        csr.mstatus.clear_imm(.{ .mie = 1 });
     }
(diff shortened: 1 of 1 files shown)


kibels:
Going to think this over more, I don't like increasing the interface surface this much just to work around the compiler not being able to optimize this.

PR #868, opened 2026-01-19
feat: use riscv atomic immediate instructions for interrupt enable/disable

kibels:
The existing Csr.write(), Csr.set() and Csr.clear() functions only allow the "r" register constraint, so always generate as two instructions e.g.
```asm
li a0, 8
csrs mstatus, a0
```
By refining the implementation a bit to detect when the value is comptime known and small enough to fit in the u5 immediate, this change allows the compiler to generate single atomic instructions for small immediates, like when setting the global interrupt enable bits.
```asm
csrsi mstatus, 8
```
The interface is changed slightly, to accept the packed struct T directly for set/clear instead of an anonymous struct. This has some benefits like a known result type, and allows detection of whether the passed in value is comptime known which the get_bits() necessary for .modify() to work obscures.

A before/after of the codegen can be observed in: https://godbolt.org/z/sKeboeoWv

CSRs that are meant to be used with these interfaces should have default field values of 0, so that the set/clear mask can be generated without providing all of the fields.

### DIFF
--- a/port/raspberrypi/rp2xxx/src/cpus/hazard3.zig
+++ b/port/raspberrypi/rp2xxx/src/cpus/hazard3.zig
@@ -447,31 +447,31 @@ pub const csr = struct {
     pub const dpc = riscv32_common.csr.dpc;
 
     pub const meiea = Csr(0xbe0, packed struct {
-        index: u5,
-        reserved0: u11,
-        window: u16,
+        index: u5 = 0,
+        reserved0: u11 = 0,
+        window: u16 = 0,
     });
     pub const meipa = Csr(0xbe1, packed struct {
-        index: u5,
-        reserved0: u11,
-        window: u16,
+        index: u5 = 0,
+        reserved0: u11 = 0,
+        window: u16 = 0,
     });
     pub const meifa = Csr(0xbe2, packed struct {
-        index: u5,
-        reserved0: u11,
-        window: u16,
+        index: u5 = 0,
+        reserved0: u11 = 0,
+        window: u16 = 0,
     });
     pub const meipra = Csr(0xbe3, packed struct {
-        index: u5,
-        reserved0: u11,
-        window: u16,
+        index: u5 = 0,
+        reserved0: u11 = 0,
+        window: u16 = 0,
     });
     pub const meinext = Csr(0xbe4, packed struct {
-        update: u1,
-        reserved0: u1,
-        irq: u9,
-        reserved1: u20,
-        noirq: u1,
+        update: u1 = 0,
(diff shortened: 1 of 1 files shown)


modules/riscv32-common/src/riscv32_common.zig
```diff
@@ -476,6 +539,17 @@ pub const csr = struct {
                     else => @compileError("unsupported type"),
                 };
             }
+
+            inline fn comptimeKnown(x: anytype) bool {
```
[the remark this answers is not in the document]
kibels: renamed to is_comptime to keep it short

modules/riscv32-common/src/riscv32_common.zig
```diff
@@ -136,9 +136,42 @@ pub const csr = struct {
...
+        pub const Xs = enum(u2) {
+            /// Extension unit status
+            off = 0b00,
+            initial = 0b01,
+            clean = 0b10,
+            dirty = 0b11,
+            pub const zero = .off;
+            pub const mask = .dirty;
+        };
+        pub const Fs = Xs;
```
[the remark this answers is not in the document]
kibels: Capitalized all register field abbreviations to keep in line with microzig style

modules/riscv32-common/src/riscv32_common.zig
```diff
@@ -136,9 +136,42 @@ pub const csr = struct {
...
-        mie: u1,
-        reserved1: u28,
+        pub const XS = enum(u2) {
+            /// Extension unit status
+            off = 0b00,
+            initial = 0b01,
+            clean = 0b10,
+            dirty = 0b11,
+            pub const zero = .off;
+            pub const mask = .dirty;
```
Grazfather: What are these for?

Did some intermediate version of this PR need these?

kibels: Just for clarity that it's initialized to .zero in the struct defaults and that mstatus.set/clear({.fs = .mask}) sets/clears the whole field. The code reads a bit strange if you're using .dirty as a mask unless you happen to know dirty = 0b11.

kibels: The typical usage is probably mstatus.write(.{ .fs = .init }) in startup code or mstatus.modify(.{ .fs = .clean }) in the kernel after saving registers so maybe it's unnecessary.

Grazfather: ah, ok I get it

Grazfather:
Can you rebase/merge in main? #864 might need an update since you added a type to mpp

EDIT: Not needed

Issue #830, opened 2026-01-04
Make microzig_main automatically call `init` for board


Grazfather:
From [this comment](https://github.com/ZigEmbeddedGroup/microzig/pull/743#discussion_r2657883063) from @tact1m4n3.

In the nordic and wch hals (maybe others) we have the example apps explicitly call `board.init()`.

Maybe we should add logic to [`microzig_main`](https://github.com/ZigEmbeddedGroup/microzig/blob/af10feccfb2fc5daa6656fad4714e4fb8885edbf/core/src/start.zig#L39) to call it if its present. It already does this for the app and for the hal.

Makes sense to me, my only real concern is whether this call should supercede an init for the chip/cpu, or be called in addition to it.

- If its the former, then the board init should probably call the chip init, which should probably call the cpu init.
  - This grants extra flexibility (A board might opt not to call a chip init, for the reason below)
  - But it's also extra boilerplate (more boards should call chip init, which should call cpu init) and duplicated code (they might opt to manually do some work, which is more places to make sure to update)
- If it's the latter, I think we run the risk of some configs doing extra work (best case) or trashing each other.
  - For example, a chip setting up RCC to set up the clocks, only for a board, with an external clock, to immediately set them up again with an external clock.

mattnite:
Yes, let's add `board.init()`, `app -> board -> hal` is the priority we want for sure.

I immediately thought that we should add chip and cpu as well, however:

- `chip` is _supposed_ to be generated code, you can hand code it yourself, but it really should just help make sense of MMIO and other chips specific info, no real code.
- We effectively have `init` with cpu with the startup code

tact1m4n3:
> app -> board -> hal is the priority

👍 

I would add init functions only for hal and board. For cpu it doesn't make sense since the cpu already provides the entry point. For chip I agree that it is best to not include any real code there.

PR #869, opened 2026-01-19
STM32:Use ClockTree helper for STM32F3

mathk:


examples/stmicro/stm32/src/hts221.zig
```diff
@@ -14,7 +14,11 @@ pub const microzig_options: microzig.Options = .{
 
 pub fn init() void {
     stm32.dma.DMA1_Channel4.enable_interrupt();
-    board.init();
+    if (comptime std.mem.eql(u8, microzig.config.chip_name, "STM32L476VG")) {
+        board.init();
+    } else {
+        board.init(&board.rcc_medium_speed) catch @panic("Board init fail");
```
Grazfather: why don't you just have each board have its own speed configuration? Does this example need a specific speed? I don't like the examples having this chip stuff hardcoded.
#830 will have this function be called automatically, so we can't do this.

Maybe we can expose some option that has board.init select which config to use?

port/stmicro/stm32/src/hals/common/i2c_v2.zig
```diff
@@ -30,13 +30,22 @@ const TimingSpec_Standard = .{
     .t_min_af = 0.05,
 };
 
+const TIMINGR = blk: for (@typeInfo(I2C_Peripherals).@"struct".fields) |field|
+{
+    if (std.mem.eql(u8, "TIMINGR", field.name)) {
+        break :blk field.type;
+    }
+} else @panic("No TIMINGR register");
```
Grazfather: this should be a compileError.

examples/stmicro/stm32/src/hts221.zig
mathk: Yeap option could be a good choice.

mathk: For clock speed I was thinking that this could be a user choice depending on the usage. (low power mode...) The idea is to be able to have comptime value for each ClockTree and that could be used for each peripheral at comptime too for different config computation.

port/stmicro/stm32/src/boards/STM32F3DISCOVERY.zig
```diff
@@ -39,12 +61,12 @@ pub fn init_log() void {
         },
     }).apply();
     uart_logger.init(.{
-        .baud_rate = 115200,
+        .baud_rate = 9600,
```
Grazfather, line 64: why?

port/stmicro/stm32/src/boards/STM32F3DISCOVERY.zig
```diff
@@ -57,5 +79,5 @@ pub fn i2c1() hal.i2c.I2C_Device {
         },
     }).apply();
 
-    return hal.i2c.I2C_Device.init(.I2C1);
+    return try hal.i2c.I2C_Device.init(.I2C1);
```
Grazfather, line 82: not needed, but i guess we do this a lot anyway.

port/stmicro/stm32/src/hals/common/i2c_v2.zig
```diff
@@ -194,9 +189,24 @@ const I2C = struct {
         return i2c.write_blocking_intern(address, false, data);
     }
 
-    pub fn init(comptime instance: I2C_Type) I2C {
+    /// This init should be comptime safe.
```
Grazfather: nit: this is a note, not a doc string. Can you describe what it's doing?

port/stmicro/stm32/src/hals/STM32F303/rcc.zig
```diff
@@ -1,127 +1,308 @@
...
-        return RccErrorConfig.WrongClockSource;
-    }
+    clean_clock();
+    select_clock();
+}
 
-    if (source == .HSE and current_clock.hse == 0) {
-        return RccErrorConfig.SourceClockNotInitialized;
-    }
+fn get_ppll_source() PLLSRC {
```
Grazfather: typo?

port/stmicro/stm32/src/hals/STM32L47X/rcc.zig
```diff
@@ -1,7 +1,10 @@
 const microzig = @import("microzig");
 const enums = @import("../common/enums.zig");
 const util = @import("../common/util.zig");
+const clock_tree = @import("ClockTree").get_mcu_tree(microzig.config.chip_name);
 
+//expose only configurations structs
```
Grazfather, line 6:
```suggestion
// Expose only configurations structs
```

port/stmicro/stm32/src/boards/STM32F3DISCOVERY.zig
mathk: For some unknown reason on my board the VCOM port is only stable at that frequency. I suspect STLink chip to be out of sync with the target chip but no idea how to debug it further.

Grazfather: OK strange. This isn't ideal 'cause we keep all the examples at 115200 but it's OK

PR #875, opened 2026-01-24
esp: RTOS and systimer fixes

tact1m4n3:
This PR aims to fix the deadlocks happening after a while in firmware using the RTOS. Marked as draft until we can confirm this actually fixes the bug.

Problem: The timer interrupt supposed to wake the cpu when a timeout expired didn't trigger. This caused tasks that did any sort of waiting (either sleeping or take a semaphore with a timeout) to never wake up (though waking up from taking from a semaphore still happened when the semaphore was given).

### DIFF
--- a/port/espressif/esp/src/hal/systimer.zig
+++ b/port/espressif/esp/src/hal/systimer.zig
@@ -106,10 +106,9 @@ pub const Unit = enum(u1) {
         OP.modify(.{ .TIMER_UNIT0_UPDATE = 1 });
         while (OP.read().TIMER_UNIT0_VALUE_VALID != 1) {}
 
-        // Read LO, HI, then LO again, check that LO returns the same value.
-        // This accounts for the case when an interrupt may happen between
-        // reading HI and LO values, and this function may get called from the
-        // ISR. In this case, the repeated read will return consistent values.
+        // Read LO, HI, then LO again, check that LO returns the same value. This accounts for the case
+        // when an interrupt may happen between reading HI and LO values, and this function may get
+        // called from the ISR. In this case, the repeated read will return consistent values.
         var lo_start = LO.read().TIMER_UNIT0_VALUE_LO; // 32 bits
         while (true) {
             const lo = lo_start;


tact1m4n3:
Before merging I'd like to wait for confirmation from @andreacoradi that this fix indeed fixed his problem.
I've also ran some tests on my own and I haven't encountered any deadlocks based on this fix.

port/espressif/esp/src/hal/rtos.zig
```diff
@@ -709,14 +712,10 @@ pub const TimerTicks = enum(u52) {
...
-    pub fn is_reached(ticks: TimerTicks) bool {
-        return ticks.is_before(.now());
-    }
-
-    pub fn is_before(a: TimerTicks, b: TimerTicks) bool {
+    pub fn is_reached_by(a: TimerTicks, b: TimerTicks) bool {
         const _a = @intFromEnum(a);
         const _b = @intFromEnum(b);
-        return _a < _b or _b -% _a < _a;
+        return _b -% _a < std.math.maxInt(u51);
```
Grazfather: technically wrong I think? I think you want `<=`

port/espressif/esp/src/hal/rtos.zig
```diff
@@ -801,7 +800,7 @@ pub const Mutex = struct {
 
         while (mutex.locked) |owning_task| {
             if (maybe_timeout_ticks) |timeout_ticks|
-                if (timeout_ticks.is_reached())
+                if (timeout_ticks.is_reached_by(.now()))
```
Grazfather, line 802: why did you remove `is_reached`?

port/espressif/esp/src/hal/systimer.zig
```diff
@@ -106,9 +106,6 @@ pub const Unit = enum(u1) {
         OP.modify(.{ .TIMER_UNIT0_UPDATE = 1 });
         while (OP.read().TIMER_UNIT0_VALUE_VALID != 1) {}
 
-        // Read LO, HI, then LO again, check that LO returns the same value. This accounts for the case
-        // when an interrupt may happen between reading HI and LO values, and this function may get
-        // called from the ISR. In this case, the repeated read will return consistent values.
```
Grazfather, line 111: Why are you getting rid of this? iirc this is copied from pico sdk

tact1m4n3: oh. didn't mean to do this

port/espressif/esp/src/hal/rtos.zig
Grazfather: if you just think it's more explicit, that makes sense to me. Just curious.

tact1m4n3: I made it look more like microzig driver framework's Timeout (which has is_reached_by instead of is_before) and when using decl literals for getting .now() it looks pretty clean. But I can add it back if you want

tact1m4n3: Also `is_reached` hid the fact that you were fetching the time

Grazfather: yep I like more explicit

port/espressif/esp/src/hal/systimer.zig
```diff
@@ -106,9 +106,10 @@ pub const Unit = enum(u1) {
...
+        // Read LO, HI, then LO again, check that LO returns the same value.
+        // This accounts for the case when an interrupt may happen between
+        // reading HI and LO values, and this function may get called from the
+        // ISR. In this case, the repeated read will return consistent values.
```
Grazfather:
```suggestion
        // Read LO, HI, then LO again, check that LO returns the same value. This accounts for the case
        // when an interrupt may happen between reading HI and LO values, and this function may get
        // called from the ISR. In this case, the repeated read will return consistent values.
```
(Just reverting the fix, why take the diff?)

PR #876, opened 2026-01-24
Ch32v cleanup

Grazfather:
- ch32v:
    - Fixed WFE and reset vector
    - switched mepc writes to CSR
    - clarified build order
    - added/explained CSRs, HPE, and mret/mred behavior.
    - Renamed qingkev4-rv32imafc.zig → qingkev4-rv32imacf.zig to match WCH naming.
  - riscv32 Core:
    - Removed unused functions
    - consolidated PMP setup and m2u switch into single asm blocks.
  - STM32 HAL: Cleaned comments and wrapped long lines in F103 i2c.zig and uart.zig.
  - WCH DMA HAL: Renamed DmaRegs → DMA_Regs.
  - Drivers: Sharp Memory LCD misc cleanup.

### DIFF
--- a/core/src/cpus/riscv32.zig
+++ b/core/src/cpus/riscv32.zig
@@ -14,27 +14,6 @@ pub fn wfe() void {
     asm volatile ("csrs 0x810, 0x1");
 }
 
-pub fn pmp_open_all_space() void {
-    asm volatile (
-        \\
-        // Config entry0 addr to all 1s to make the range cover all space
-        \\li x6, 0xffffffff
-        \\csrw pmpaddr0, x6
-        // Config entry0 cfg to make it NAPOT address mode, and R/W/X okay
-        \\li x6, 0x7f
-        \\csrw pmpcfg0, x6
-        ::: .{ .x6 = true });
-}
-
-pub fn switch_m2u_mode() void {
-    asm volatile (
-        \\la x6, 1f
-        \\csrw mepc, x6
-        \\mret
-        \\1:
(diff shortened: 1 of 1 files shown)


core/src/cpus/riscv32.zig
```diff
@@ -15,19 +15,24 @@ pub fn wfe() void {
 }
 
 pub fn pmp_open_all_space() void {
```
Grazfather: Note: We never call this. This is from #21. Maybe we should just remove it?

core/src/cpus/riscv32.zig
```diff
@@ -15,19 +15,24 @@ pub fn wfe() void {
...
+        // Config entry0 addr to all 1s to make the range cover all space
+        \\li x6, 0xffffffff
+        \\csrw pmpaddr0, x6
+        // Config entry0 cfg to make it NAPOT address mode, and R/W/X okay
+        \\li x6, 0x7f
+        \\csrw pmpcfg0, x6
+        ::: .{ .x6 = true });
 }
 
 pub fn switch_m2u_mode() void {
```
Grazfather: Note: We never call this. This is from #21. Maybe we should just remove it?

core/src/cpus/riscv32.zig
mattnite: Kill it

core/src/cpus/riscv32.zig
mattnite: Ditto

PR #878, opened 2026-01-24
ch32v: Fix: Add clobbers to bss/data loops

Grazfather:
I noticed that ch32v examples were broken. `git bisect` got me to #868, but some debugging with @kibels suggested something else was going on.

Looks like those changes just exposed a bug where we weren't clobbering properly, so a subsequent CSR, which was using a register, had had it clobbered.

I haven't exactly tracked why this exposed the issue, since the csrwi instructions (only changed ones) are before the problematic function.

### DIFF
--- a/port/wch/ch32v/src/cpus/main.zig
+++ b/port/wch/ch32v/src/cpus/main.zig
@@ -310,7 +310,7 @@ pub const startup_logic = struct {
             \\    addi a1, a1, 4
             \\    blt a1, a2, clear_bss_loop
             \\clear_bss_done:
-        );
+            ::: .{ .x10 = true, .x11 = true, .x12 = true });
 
         // Copy .data from FLASH to RAM.
         asm volatile (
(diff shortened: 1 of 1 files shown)


Grazfather:
Making this change causes the issue to be exhibited (for me on zig 0.15.2). The clobbers fix the issue.

```patch
diff --git i/port/wch/ch32v/src/cpus/main.zig w/port/wch/ch32v/src/cpus/main.zig
index a477503d..c1192fd2 100644
--- i/port/wch/ch32v/src/cpus/main.zig
+++ w/port/wch/ch32v/src/cpus/main.zig
@@ -236,7 +236,7 @@ pub const startup_logic = struct {
         );

         // NOTE: this can only be called once. Otherwise, we get a linker error for duplicate symbols
-        startup_logic.initialize_system_memories();
+        microzig.utilities.initialize_system_memories(.auto);

         // Configure the CPU.
         switch (cpu_name) {
```

Even though the issue is 'far away', for me this caused the `jal _system_init` to happen AFTER putting `&microzig_main` into a register for the subsequent write to `mepc`.

PR #879, opened 2026-01-25
add zicsr and zifencei for riscv targets

kibels:
Minor correctness fix, @Grazfather clued me in that the elf headers were off so zicsr and zifencei instructions weren't getting disassembled.

e.g. with riscv64-elf-objdump
```
 172:   30551073                .insn   4, 0x30551073
```
becomes
```
 172:   30551073                csrw    mtvec,a0
```


Grazfather:
Good find!

PR #884, opened 2026-01-26
`add_enum_and_apply` patch operation

mattnite:


port/raspberrypi/rp2xxx/patches/rp2040.zon
```diff
@@ -1,6 +1,6 @@
 .{
     .{
-        .add_enum = .{
+        .create_enum_and_apply = .{
```
Grazfather: for consistency's sake, we might want to call this `add_enum_and_apply`.

PR #890, opened 2026-01-27
Sorcerer: Add CLI

Grazfather:
- Don't generate json anymore, instead generate a zig file and use that directly in sorcer[cli]
- Add a cli that can generate all the output files. This is a lot simpler than `regz`, but basically just wraps it.
  - Also provides a way to list all targets!
- Add a README

### DIFF
--- a/tools/sorcerer/src/cli.zig
+++ b/tools/sorcerer/src/cli.zig
@@ -162,10 +162,10 @@ fn print_list_table(allocator: Allocator, port_filter: ?[]const u8) !void {
     defer seen_chips.deinit();
 
     // Print header
-    stdout.print("{s:<24} {s:<24} {s}\n", .{ "CHIP", "PORT", "BOARD" }) catch |err| return handle_write_error(err);
-    stdout.print("{s:-<24} {s:-<24} {s:-<32}\n", .{ "", "", "" }) catch |err| return handle_write_error(err);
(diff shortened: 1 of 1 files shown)


tools/sorcerer/src/cli.zig
```diff
@@ -0,0 +1,427 @@
...
+
+        // Apply filter
+        if (port_filter) |filter| {
+            if (std.mem.indexOf(u8, port_name, filter) == null) {
+                continue;
+            }
+        }
+
+        for (schema.chips) |chip| {
+            const board_name = if (schema.boards.len > 0) schema.boards[0].name else "-";
```
Grazfather: Maybe we stop at chips?

Grazfather:
Pushed a change to only list chips, since that's the granularity at which we generate SVDs, so basically keeping this as an regz-generator and not a 'target lister'

tools/sorcerer/build.zig
```diff
@@ -14,14 +14,9 @@ pub fn build(b: *std.Build) void {
     const mb = MicroBuild.init(b, mz_dep) orelse return;
     const register_schemas = get_register_schemas(b, mb) catch @panic("OOM");
     const write_files = b.addWriteFiles();
-    const register_schema = write_files.add("register_schemas.json", std.json.Stringify.valueAlloc(b.allocator, register_schemas, .{}) catch @panic("OOM"));
```
mattnite, line 17: I still want the register schema data to be a runtime file data instead of compile-time, at least for the GUI.

PR #893, opened 2026-01-31
ch32 USBHD/USBHS driver implementation

Copper280z:
This adds a new driver for the ch32 USBHD/USBHS peripheral, USB_OTG_FS peripheral, and examples for the ch32v307, and ch32v203 boards.


core/src/core/usb.zig
```diff
@@ -274,7 +274,7 @@ pub fn DeviceController(config: Config, driver_args: config.DriverArgs()) type {
             const desc_device: descriptor.Device = .{
                 .bcd_usb = config.bcd_usb,
                 .device_triple = config.device_triple,
-                .max_packet_size0 = @max(config.max_supported_packet_size, 64),
+                .max_packet_size0 = @min(config.max_supported_packet_size, 64),
```
Grazfather, line 259: woof

examples/wch/ch32v/build.zig
```diff
@@ -36,7 +36,9 @@ pub fn build(b: *std.Build) void {
         .{ .target = mb.ports.ch32v.chips.ch32v203x6, .name = "blinky_ch32v203", .file = "src/blinky.zig" },
         .{ .target = mb.ports.ch32v.chips.ch32v203x6, .name = "blinky_systick_ch32v203", .file = "src/blinky_systick.zig" },
         .{ .target = mb.ports.ch32v.boards.ch32v203.nano_ch32v203, .name = "nano_ch32v203_blinky", .file = "src/board_blinky.zig" },
+        .{ .target = mb.ports.ch32v.boards.ch32v203.nano_ch32v203, .name = "nano_ch32v203_usb_cdc", .file = "src/usb_cdc.zig" },
         .{ .target = mb.ports.ch32v.boards.ch32v203.suzuduino_uno_v1b, .name = "suzuduino_blinky", .file = "src/board_blinky.zig" },
+        .{ .target = mb.ports.ch32v.boards.ch32v203.suzuduino_uno_v1b, .name = "suzuduino_usb_cdc", .file = "src/usb_cdc.zig" },
```
Grazfather, line 42: is this one tested?

examples/wch/ch32v/build.zig
```diff
@@ -46,13 +48,15 @@ pub fn build(b: *std.Build) void {
         .{ .target = mb.ports.ch32v.boards.ch32v203.lana_tny, .name = "lana_tny_spi_loopback", .file = "src/spi_loopback.zig" },
         .{ .target = mb.ports.ch32v.boards.ch32v203.lana_tny, .name = "lana_tny_spi_flash_w25q", .file = "src/spi_flash_w25q.zig" },
         .{ .target = mb.ports.ch32v.boards.ch32v203.lana_tny, .name = "lana_tny_sharp_niceview", .file = "src/sharp_niceview.zig" },
+        .{ .target = mb.ports.ch32v.boards.ch32v203.lana_tny, .name = "lana_tny_usb_cdc", .file = "src/usb_cdc.zig" },
```
Grazfather: and this one? I have this so I can test it for you.

port/wch/ch32v/src/hals/ch32v20x.zig
```diff
@@ -9,6 +9,7 @@ pub const i2c = @import("i2c.zig");
 pub const usart = @import("usart.zig");
 pub const spi = @import("spi.zig");
 pub const dma = @import("dma.zig");
+pub const usb = @import("usbfs.zig");
```
Grazfather, line 12: Can you add this in the test block as well

port/wch/ch32v/src/hals/ch32v30x.zig
```diff
@@ -5,18 +5,32 @@ pub const clocks = @import("clocks.zig");
...
 pub const i2c = @import("i2c.zig");
 pub const usart = @import("usart.zig");
+const std = @import("std");
+pub const usb = @import("usbhs.zig");
+pub const usbfs = @import("usbfs.zig");
 
 /// HSI (High Speed Internal) oscillator frequency
 /// This is the fixed internal RC oscillator frequency for CH32V30x
 pub const hsi_frequency: u32 = 8_000_000; // 8 MHz
+pub const hse_frequency: u32 = 8_000_000; // 8 MHz
```
Grazfather: This should be board level, since it's external to the chip and may not be present.

port/wch/ch32v/src/hals/clocks.zig
```diff
@@ -60,9 +62,12 @@ pub fn enable_peripheral_clock(p: peripheral) void {
         // AHB peripherals
         .DMA1 => RCC.AHBPCENR.modify(.{ .DMA1EN = 1 }),
         .DMA2 => RCC.AHBPCENR.modify(.{ .DMA2EN = 1 }),
+        .USBHS => RCC.AHBPCENR.modify(.{ .USBHS_EN = 1 }),
+        .USBOTG => RCC.AHBPCENR.modify(.{ .OTG_EN = 1 }),
 
         // APB2 peripherals (high-speed bus)
-        .USART1 => RCC.APB2PCENR.modify(.{ .USART1EN = 1 }),
+        .USART1,
```
Grazfather: old command reformatted for no reason?

port/wch/ch32v/src/hals/clocks.zig
```diff
@@ -413,6 +418,173 @@ pub fn get_freqs() ClockSpeeds {
...
+}
+// ============================================================================
+// Enable + configure USBHS clocks.
+// ============================================================================
+
+/// This configures RCC_CFGR2 fields for the USBHS PHY PLL reference (if present),
+/// selects whether the USBHS 48MHz clock comes from the system PLL clock or the USB PHY,
+/// and enables the AHB clock gate for USBHS.
+///
+/// Note: The SVD names bit31 as `USBFSSRC`, but the reference manual
```
Grazfather, line 443: Can we patch this? Which SVD? All of them?

examples/wch/ch32v/src/usb_cdc.zig
```diff
@@ -0,0 +1,139 @@
...
+const RCC = microzig.chip.peripherals.RCC;
+const AFIO = microzig.chip.peripherals.AFIO;
+const PFIC = microzig.chip.peripherals.PFIC;
+
+const usart = hal.usart.instance.USART1;
+
+const usart_tx_pin = gpio.Pin.init(0, 9); // PA9
+const func_pin = gpio.Pin.init(0, 10); // PA10
+const tog_pin = gpio.Pin.init(0, 14);
+const mco_pin = gpio.Pin.init(0, 8);
```
Grazfather: Might as well keep with the comments e.g. PA14 and PA8

examples/wch/ch32v/src/usb_cdc.zig
```diff
@@ -0,0 +1,139 @@
...
+    .device_triple = .unspecified,
+    .vendor = .{ .id = 0x2E8A, .str = "MicroZig" },
+    .product = .{ .id = 0x000A, .str = board.product_string },
+    .bcd_device = .v1_00,
+    .serial = "someserial",
+    .max_supported_packet_size = hal.usb.max_packet_size,
+    .configurations = &.{.{
+        .attributes = .{ .self_powered = false },
+        .max_current_ma = 50,
+        .Drivers = struct { serial: USB_Serial },
```
Grazfather, line 42: Could you maybe copy some of the extra comments I added to the cdc example for rpi?

examples/wch/ch32v/src/usb_cdc.zig
```diff
@@ -0,0 +1,139 @@
...
+    }},
+}, .{.{
+    .serial = .{ .itf_notifi = "Board CDC", .itf_data = "Board CDC Data" },
+}});
+
+pub var usb_dev: hal.usb.Polled(
+    .{},
+) = undefined;
```
Grazfather:
```suggestion
pub var usb_dev: hal.usb.Polled(.{}) = undefined;
```

examples/wch/ch32v/src/usb_cdc.zig
```diff
@@ -0,0 +1,139 @@
...
+// NOTE: After each USB chunk transfer, we have to call the USB task so that bus TX events can be handled
+pub fn usb_cdc_write(serial: *USB_Serial, comptime fmt: []const u8, args: anytype) void {
+    const text = std.fmt.bufPrint(&usb_tx_buff, fmt, args) catch &.{};
+    var write_buff = text;
+    while (write_buff.len > 0) {
+        write_buff = write_buff[serial.write(write_buff)..];
+        while (!serial.flush()) {
+            // pins.tog.put(0);
+            usb_dev.poll(false, &usb_controller);
+            // pins.tog.put(1);
```
Grazfather: gets rid of toggle pins completely or uncomment these please

Grazfather (commented):
Reviewed everything except for the HALs themselves (since this is a draft). Will get to those when you think they're ready.

examples/wch/ch32v/build.zig
Copper280z: nano_ch32v203 is tested, suzuduino is not, but the way the peripheral and pins work there's no board specific configuration. It automatically configures the pins, and there's only one option for the pair, so if usb is exposed on the board and there's an external crystal, it should work.

More testing is always better though, so test away.

examples/wch/ch32v/build.zig
Copper280z: Same as above, there's only one set of pins, and the peripheral configures them automatically, but more testing is better, so please do so!

examples/wch/ch32v/src/usb_cdc.zig
Copper280z: These will be removed before I mark it as ready, MCO and toggle are things I was using for testing. MCO is useful for validating PLL stuff with a scope.

port/wch/ch32v/src/hals/clocks.zig
Copper280z: At this point it's unclear which name is correct. I should probably either remove the comment or figure it out.

core/src/core/usb.zig
Copper280z: I could find it in the spec, but this is correct now. For LS and FS devices endpoint 0 can be 8 to 64 bytes, for HS it's always 64.

Unofficial, but useful reference: https://www.beyondlogic.org/usbnutshell/usb4.shtml

port/wch/ch32v/src/hals/ch32v20x.zig
Grazfather: Also, as you noticed: Some variants of this have both USBD and USBHD. Somewhat confusingly you're calling `hal.usb` usbfs here, but for 3xx `usb` is hs and usbfs is ufbfs.

examples/wch/ch32v/build.zig
Grazfather: As mentioned on discord: This device has the x6 bin which is the 'usbd' variant of the usbfs, so it's not (yet) supported.

port/wch/ch32v/src/hals/usbfs.zig
```diff
@@ -0,0 +1,634 @@
...
+    TOG: u1 = 0x0,
+    /// enable automatic toggle after successful transfer completion on endpoint 1/2/3: 0=manual toggle, 1=automatic toggle
+    AUTO: bool = false,
+    padding: u4 = 0,
+});
+
+fn base_addr() usize {
+    return @intFromPtr(Regs);
+}
+fn mmio_u32(off: usize) *volatile Reg_U32 {
```
piotrfila: You can define an array of registers instead of doing manual pointer arithmetic

piotrfila: It would be even better if we had a regz patch for situations like this, rp2xxx's usb would also benefit from it

port/wch/ch32v/src/hals/usbfs.zig
```diff
@@ -0,0 +1,634 @@
...
+
+    const mode_offset: usize = switch (ep) {
+        1, 4 => 0x0c,
+        2, 3 => 0x0d,
+        5, 6 => 0x0e,
+        7 => 0x0f,
+        else => unreachable,
+    };
+    const enable_bits: u8 = switch (ep) {
+        1, 3, 6 => (1 << 6) | (1 << 7),
```
piotrfila, line 129: Nitpick, but maybe use 0b11 << shift?
Also who designed this hardware!?

port/wch/ch32v/src/hals/usbfs.zig
```diff
@@ -0,0 +1,634 @@
...
+    return @as(u4, @intCast(@intFromEnum(ep)));
+}
+// --- USBHD token encodings ---
+const TOKEN_OUT: u2 = 0;
+const TOKEN_SOF: u2 = 1;
+const TOKEN_IN: u2 = 2;
+const TOKEN_SETUP: u2 = 3;
+
+// --- INT_FG raw bits (match USBHS.zig packed order) ---
+const UIF_BUS_RST: u8 = 1 << 0;
```
piotrfila: Is there some reason why you can't use a packed struct?

port/wch/ch32v/src/hals/usbfs.zig
Copper280z: Is an array of volatile pointers to packed structs safe? There was a miscompilation related to this not long ago and it's unclear if it's actually fixed. I can't find the issue on codeberg, but here's the issue on gh. https://github.com/ziglang/zig/issues/21033

port/wch/ch32v/src/hals/usbfs.zig
Copper280z: You can, but this seems more concise for a couple of reasons, the SVD field names are often combinations of device and host functionality (ie `RB_UIF_BUS_RST__RB_UIF_DETECT`). So I could write a new mmio type with good names and point it at the address. Then the use sites get nicer in some places (fewer bitwise operations!), but messier in others ie:
```                const flags = Regs.USB_INT_FG.read();
                if (flags.RB_UIF_HST_SOF == 0 and flags.RB_UIF_SUSPEND == 0 and
                    flags.RB_UIF_FIFO_OV == 0 and flags.RB_UIF_BUS_RST__RB_UIF_DETECT == 0 and
                    flags.RB_UIF_SETUP_ACT == 0 and flags.RB_UIF_TRANSFER == 0)
                    break;
```
instead of
```
                const fg: u8 = Regs.USB_INT_FG.raw & ~UIF_ISO_ACT;
                if (fg == 0) break;
```
If I do both where the call sites are nicer, then I have the existing flag defs and a new mmio definition. 

If you want it like this, I'll change it, but I don't necessarily think it's better. Maybe I should look at it tomorrow, though.

port/wch/ch32v/src/hals/usbfs.zig
Copper280z: This puts the specific bit position in the code as a number, ie it's obvious that bit 6 and 7 are set here. 0b11 << 6 takes just a little more thought to recognize it's both 6 and 7. If you feel strongly, I'll change it, otherwise I'll trust the compiler to condense this from 3 ops to 2.

port/wch/ch32v/src/hals/usbfs.zig
Copper280z: Seems like that miscompilation is either resolved, or I used enough explicit *volatile types to avoid it, but this is done and works on hardware with a release build.

Grazfather (approved):
LGTM. @piotrfila please reply to his comments, we can maybe clean up in a followup PR. Want to get this in.

port/wch/ch32v/src/hals/usbfs.zig
Copper280z: I did implement this, btw

piotrfila:
Yeah, everything's good

PR #894, opened 2026-02-02
rp2xxx: Add scratch sections

vesim987:


arkadiuszwojcik:
@vesim987 wouldn't be a problem for you to also add `__StackOneTop` and `__StackTop` to this PR?

vesim987:
@arkadiuszwojcik i was thinking about doing more things with stack in separate PR, like moving it to SCRATCH_X, using MSPLIM/MPU to set stack limit, and couple of other things.

Grazfather (commented):
Could you add comments somewhere? What are these for?

mattnite:
@vesim987 can you please respond to @Grazfather's question, I would like to know as well.

mattnite:
I don't want to merge this now, so I'm closing it. I've created https://github.com/ZigEmbeddedGroup/microzig/issues/951 because I do want the essence of this, but I think we need a larger discussion first about linkerscripts. This has a lot of duplication and feels a bit too closely coupled with RP2350.

PR #895, opened 2026-02-02
core.Mmio: Respect field access type

piotrfila:
This is my attempt at making Mmio access-type aware. This could catch things like trying to write to a read-only register.


port/nordic/nrf5x/src/hal/clocks.zig
```diff
@@ -63,13 +65,14 @@ pub const lfclk = struct {
...
     pub fn calibrate() void {
+        // Consider patching the svd
         switch (version) {
             .nrf51 => {
                 CLOCK.TASKS_CAL = 1;
                 while (CLOCK.EVENTS_DONE == 0) {}
             },
             .nrf52 => {
-                CLOCK.TASKS_CAL.write_raw(1);
+                CLOCK.TASKS_CAL.raw(1);
```
tact1m4n3: = 1

piotrfila: Thanks, I'm sure I've missed more. It's still early WIP I just wanted to get CI coverage

PR #897, opened 2026-02-07
USB: Add clarifying comments.

Grazfather:
I spent some (a lot of, actually) time trying to understand how our USB stack works.

Hopefully these comments make it easier for the next person.

### DIFF
--- a/core/src/core/usb.zig
+++ b/core/src/core/usb.zig
@@ -386,13 +386,9 @@ pub fn DeviceController(config: Config, driver_args: config.DriverArgs()) type {
             // Create a tuple with the appropriate types
             const ep_handlers_types: [2]type = .{ Tuple(&ep_handler_types[0]), Tuple(&ep_handler_types[1]) };
             var ep_handlers: Tuple(&ep_handlers_types) = undefined;
-            // For IN/OUT
+            // Iterate over all IN and OUT endpoints and bind the handler for any that are set.
             for (&ep_handler_types, &ep_handler_names, &ep_handler_drivers, 0..) |htypes, hnames, hdrivers, dir| {
-                // for each endpoint slot for this direction
                 for (&htypes, &hnames, &hdrivers, 0..) |T, name, drv_id, ep| {
-                    // Bind the handler: If the type isn't void, then get the driver (e.g. .serial),
-                    // then for the handler name get that field in the driver, it's a function,
-                    // assign it as a callback
                     if (T != void)
                         ep_handlers[dir][ep] = @field(driver_fields[drv_id.?].type.handlers, name);
                 }


core/src/core/usb.zig
```diff
@@ -47,6 +48,7 @@ pub fn Struct(
     } });
 }
 
+// What does this do? It lets you iterate through interfaces and endpoints?
```
piotrfila:
```suggestion
// This is used at comptime to give each driver unique endpoint and interface numbers.
// Descriptor strings are deduplicated and also given unique numbers.
```

core/src/core/usb.zig
```diff
@@ -298,24 +312,33 @@ pub fn DeviceController(config: Config, driver_args: config.DriverArgs()) type {
...
                 size += @sizeOf(Descriptors);
 
+                // Collect handler types, names, and drivers, to later be bound to the appropriate
+                // endpoints
                 for (@typeInfo(Descriptors).@"struct".fields, 0..) |fld, desc_num| {
                     const desc = @field(descriptors, fld.name);
 
+                    // Determine which interface numbers this driver owns. If it is an
+                    // InterfaceAssociation, then use the interface count. If it is an Interface,
+                    // then the driver owns just that one interface.
```
piotrfila:
```suggestion
                    // Determine which interface numbers this driver owns from the first descriptor.
                    // If it is of type InterfaceAssociation, then use the interface count.
                    // If it is of type Interface, then the driver owns just that one interface.
                    // Other types are not allowed for the first descriptor.
```

core/src/core/usb.zig
```diff
@@ -356,11 +381,18 @@ pub fn DeviceController(config: Config, driver_args: config.DriverArgs()) type {
...
+            // Create a tuple with the appropriate types
             const ep_handlers_types: [2]type = .{ Tuple(&ep_handler_types[0]), Tuple(&ep_handler_types[1]) };
             var ep_handlers: Tuple(&ep_handlers_types) = undefined;
+            // For IN/OUT
             for (&ep_handler_types, &ep_handler_names, &ep_handler_drivers, 0..) |htypes, hnames, hdrivers, dir| {
+                // for each endpoint slot for this direction
                 for (&htypes, &hnames, &hdrivers, 0..) |T, name, drv_id, ep| {
+                    // Bind the handler: If the type isn't void, then get the driver (e.g. .serial),
+                    // then for the handler name get that field in the driver, it's a function,
+                    // assign it as a callback
```
piotrfila: I don't think T can be void anymore, this check can be removed

core/src/core/usb.zig
Grazfather: Oops! Yes this was an old comments I meant to replace with an answer. Thank you.

core/src/core/usb.zig
Grazfather: It can (I flipped the logic and added a compile log). This loops over all 32 eps, so in the case of the rpi cdc example, it is void every time except for IN 2 and OUT 1.

We could just check drv_id is not null if you think that's cleaner.

piotrfila: I see, my bad

piotrfila:
This looks great, the comments are a lot clearer now

Grazfather:
Thank you!

PR #898, opened 2026-02-08
esp: RTOS optimizations

tact1m4n3:
Some numbers with 200 concurrent tasks:
- Plain cooperative rescheduling of a task until the switch to a different task: 140-150 cycles
- If the task suspends itself (eg: waiting on a mutex without timeout): 126-137 cycles (this only measures the yielding time, not the overhead induced by the mutex)
- If the task sleeps: 268-1700 cycles (1700 is for the 200th task that must be inserted at the end of the sleep_queue since it wakes up last). This is O(n). Also there is an overhead in the tick interrupt when all these tasks are woken up: the max tick interrupt time is close to 4000cycles (O(n) with the number of tasks waking up on the same tick) (otherwise being ~250cycles if no tasks wake).

I also added support for time slicing (switch between tasks of the same priority on the tick interval). This is by default off.

Tasks can now be awaited and the memory freed. I used this to implement cleanup for radio/timer.zig, but for the tasks spawned by the wifi driver it is more complicated so I will leave this for a later PR.

Unit tests for the esp port now run on CI.


port/espressif/esp/src/hal/rtos.zig
```diff
@@ -39,24 +40,64 @@ const systimer = @import("systimer.zig");
 const STACK_ALIGN: std.mem.Alignment = .@"16";
 const EXTRA_STACK_SIZE = @max(@sizeOf(TrapFrame), 32 * @sizeOf(usize));
 
+pub const TickFrequency = enum(u32) {
+    _,
+
+    pub fn from_hz(v: u32) TickFrequency {
+        assert(v < 100_000); // frequency to high
```
Grazfather:
```suggestion
        assert(v < 100_000); // frequency too high
```

port/espressif/esp/src/hal/rtos.zig
```diff
@@ -39,24 +40,64 @@ const systimer = @import("systimer.zig");
...
+    .signedness = .unsigned,
+} })) {
+    idle = 0,
+    lowest = 1,
+    _,
+
+    pub const highest: @This() = @enumFromInt(std.math.maxInt(@typeInfo(@This()).@"enum".tag_type));
+
+    pub fn next_higher(prio: Priority) Priority {
+        return @enumFromInt(@intFromEnum(prio) +| 1);
```
Grazfather, line 95: is this supposed to wrap around? that is the `+|` for?

port/espressif/esp/src/hal/rtos.zig
```diff
@@ -486,116 +574,76 @@ pub const yield_interrupt_handler: microzig.cpu.InterruptHandler = .{
                 \\addi sp, sp, 32*4
                 \\mret
                 :
-                : [schedule_in_isr] "i" (&schedule_in_isr),
+                : [tick_handler] "i" (&tick_handler),
```
Grazfather, line 585: Do we not need all the clobbers? Is this because it's an interrupt handler?

port/espressif/esp/src/hal/rtos.zig
```diff
@@ -630,18 +678,25 @@ pub const Task = struct {
     state: State = .none,
 
     /// Node used for rtos internal lists.
-    node: std.DoublyLinkedList.Node = .{},
+    node: LinkedListNode = .{},
+
+    /// Ticks for when the task will be awaken.
```
Grazfather:
```suggestion
    /// Ticks for when the task will be awoken.
```

Grazfather: Or just 'when the task will wake' of course.

port/espressif/esp/src/hal/rtos.zig
tact1m4n3: Since it is a naked function, I don't think clobbers would make any difference since the compiler has nowhere to save/restore those registers.

port/espressif/esp/src/hal/rtos.zig
tact1m4n3: +| is saturating add which means the addition will "saturate" at the max int value before the overflow. This function isn't used anymore so I'll just remove it.

mattnite (approved):

port/espressif/esp/src/hal/rtos.zig
Grazfather: cool. Today I learned!

PR #908, opened 2026-02-13
ESP: lwIP fixes

andreacoradi:
Before this PR, lwIP caused null pointer dereferences in my setup (see [this](https://savannah.nongnu.org/bugs/?59831)).
It was also possible to use APIs that expected the core mutex to be locked without locking it.
Now it will instead panic, letting you know.
My project is now much more stable, and I haven't encountered any crashes.


tact1m4n3 (approved):
Nice! Thank you! I think the merge to main broke the examples (the assert import in exports.zig was removed). After CI passes we can merge.

andreacoradi:
Oops! Thanks for telling me

PR #910, opened 2026-02-20
rp2xxx: Unique board identifier

ianic:
We had implementation for rp2040, this makes it work also on rp2350.

Difference is explained in pico [sdk](https://github.com/raspberrypi/pico-sdk/blob/a1438dff1d38bd9c65dbd693f0e5db4b9ae91779/src/rp2_common/pico_unique_id/include/pico/unique_id.h), and this is based on pico sdk [unique.c](https://github.com/raspberrypi/pico-sdk/blob/a1438dff1d38bd9c65dbd693f0e5db4b9ae91779/src/rp2_common/pico_unique_id/unique_id.c).

### DIFF
--- a/port/raspberrypi/rp2xxx/src/hal/rom.zig
+++ b/port/raspberrypi/rp2xxx/src/hal/rom.zig
@@ -213,7 +213,7 @@ pub fn reset_to_usb_boot() void {
 
 pub fn get_sys_info(out_buffer: [*]u32, out_buffer_word_size: u32, flags: u32) void {
     switch (chip) {
-        .RP2040 => unreachable,
+        .RP2040 => @compileError("not supported on this chip"),
         .RP2350 => {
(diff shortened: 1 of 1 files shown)


port/raspberrypi/rp2xxx/src/hal/rom.zig
```diff
@@ -213,3 +210,24 @@ pub fn reset_to_usb_boot() void {
         },
     }
 }
+
+pub fn get_sys_info(out_buffer: [*]u32, out_buffer_word_size: u32, flags: u32) void {
```
tact1m4n3, line 214: I suggest you make this function return a compile error on RP2040 (instead of runtime unreachable). Just to prevent accidental calls to it.

ianic: Thanks!

PR #921, opened 2026-03-19
Add support for NRF softdevice

mattnite:
This is untested, once I see examples correctly running on my local hardware then I'll consider merging


tact1m4n3 (commented):
Great patch! I wish I could help test this but I only have a microbit v1 with an older nrf51 chip.

I think there are some issues though with restricted/blocked peripherals. I found a list here: https://docs.nordicsemi.com/bundle/sds_s140/page/SDS/s1xx/sd_resource_reqs/hw_block_interrupt_vector.html
From that list, we use CLOCK and RTC0 in `time.zig` and `clock.zig`. We should switch time stuff to another peripheral.
Also, the RTC interrupt which is set up by the hal should use another priority as 0 (default), 1, 4 are reserved by the softdevice.

After a quick skim of the code, it looks good to me (except for what I specified above).

PR #922, opened 2026-03-20
stm: update embassy data dependency

deanrobertcook:
Updates the stm32 data reference to include this addition to the [AES.ICR register](https://github.com/embassy-rs/stm32-data/pull/754)


Grazfather:
You're using the commit in the data repo, vs. the generated repo, you'll have to use the correct hash. I think it's dd0a8684df63f2956614fa3fb6c26cfff42914c2

Grazfather:
Also, if you could link to a diff between the two, since it looks like we will be pulling in a whole bunch of other stuff, as our commit is rather old.

deanrobertcook:
Apologies, I thought this might have been a simpler change. My mistake. I don't have time to follow through with it so I'll close it for now. Please keep up the good work 😊

Issue #924, opened 2026-03-23
STM32: Bad value returned by systick.get_time_since_boot().to_us()


mihainicolae80:
Hi,

Calling `systick.get_time_since_boot().to_us()` will occasionally return a bad value every 50s or so.

HW: STM32 F3 Discovery board
Using Microzig release 0.15.1
Compiling with zig 0.15.2

main.zig:
```
const std = @import("std");
const microzig = @import("microzig");
const hal = microzig.hal;
const board = microzig.board;
const log = std.log;

pub const microzig_options: microzig.Options = .{
    .logFn = microzig.board.uart_logger.log,
    .cpu = .{
        .ram_vector_table = true,
    },
    .hal = .{ .rcc_clock_config = board.rcc_medium_speed }
};

pub fn init() void
{
    hal.dma.DMA1_Channel4.enable_interrupt();
    board.init();
    board.init_log();

    hal.systick.init() catch {
        @panic("systick init fail");
    };
}

pub fn main() !void
{
    const systick = hal.systick_timer.clock_device() catch {
        @panic("systick_timer clock device");
    };

    var last_us: u64 = systick.get_time_since_boot().to_us();
    while (true)
    {
        const now_us = systick.get_time_since_boot().to_us();

        if (now_us > last_us + 20_000)
        {
            const now2_us = systick.get_time_since_boot().to_us();
            log.info("now_us jump={d}us last_us={d}us now_us={d}us now2_us={d}us", .{now_us - last_us, last_us, now_us, now2_us});
        }
        last_us = now_us;
    }
}
```

build.zig:
```
const std = @import("std");
const microzig = @import("microzig");

const MicroBuild = microzig.MicroBuild(.{
    .stm32 = true,
});

pub fn build(b: *std.Build) void {

    const mz_dep = b.dependency("microzig", .{});
    const mb = MicroBuild.init(b, mz_dep) orelse return;

    const firmware = mb.add_firmware(.{
        .name = "firmware",
        .target = mb.ports.stm32.boards.stm32f3discovery,
        .optimize = .Debug,
        .root_source_file = b.path("src/main.zig"),
    });

    mb.install_firmware(firmware, .{});
    mb.install_firmware(firmware, .{ .format = .elf });

    const flash_argv = [_][] const u8 {
        "arm-none-eabi-gdb.exe",
        "-iex",
        "set auto-load safe-path /",
        "--batch", 
        "--quiet",
        "--command=flash.gdb",
    };
[...]

PR #925, opened 2026-03-27
stm32: fix systick get_time_since_boot() bad value due to non-atomic memory read

mihainicolae80:
Based on issue #924


Grazfather (approved):
Thank you!

PR #935, opened 2026-04-19
update stm32 clock configs

RecursiveError:


mathk (commented):
LGTM, I let @mattnite  do the final approval.

PR #939, opened 2026-04-27
Implement pins_v2.zig for stm32f429 and extend support for stm32l479

### LINKED ISSUE #938: `microzig.hal` is void for stm32f429idiscovery
sample code:
```
const hal = microzig.hal;
pub fn main() !void {
    hal.parse_pin("PG13");
}
```

sample build (adapted from the getting started guide):

```
const std = @import("std");

const MicroBuild = microzig.MicroBuild(.{
    .stm32 = true,
});

pub fn build(b: *std.Build) void {
    const mz_dep = b.dependency("microzig", .{});

    const firmware = mb.add_firmware(.{
        .name = "blinky",
        .target = mb.ports.stm32.boards.stm32f429idiscovery,
        .optimize = .ReleaseSmall,
        .root_source_file = b.path("src/main.zig"),
    });

    // We call this twice to demonstrate that the default binary output for
    // RP2040 is UF2, but we can also output other formats easily
    mb.install_firmware(firmware, .{ });
    mb.install_firmware(firmware, .{ .format = .elf });
}
```

*Expected output:* A working program that does nothing.

*Output:* 
```
src/main.zig:6:8: error: type 'void' has no members
    hal.parse_pin("PG13");
    ~~~^~~~~~~~~~
```

Looking at https://github.com/ZigEmbeddedGroup/microzig/blob/48582bf8be10d024466a97a4dfa28470d5ebee14/port/stmicro/stm32/src/hals/STM32F429.zig it looks like hal should be filled with something, at least more than void. Am I doing something wrong? Is there a missing step that needs to be done to generate the hal?

JAicewizard:
fixes #938 

This mostly makes `comon/pins_v2.zig` generic to the number of GPIO ports available on a chip. 

I have looked into other methods, like having a function that determins the number of ports in `pins_v2.zig` itself, but found that it was very cumbersome to implement and it would be easier to pass it down when importing. Since the only files importing this are from microzig itself, this seems like a reasonable option.


port/stmicro/stm32/src/hals/common/pins_v2.zig
```diff
@@ -40,12 +94,99 @@ pub const Pin = enum {
...
     };
 };
 
-pub const InputGPIO = struct {
-    pin: gpio.Pin,
+const GPIO_Pin = struct {
+    pin: Pin,
+    port: Port,
+
+    inline fn write_pin_config(_gpio: GPIO_Pin, mode: Mode) void {
```
mattnite, line 105: `_gpio` => `p` for pin is more inline with the style.

port/stmicro/stm32/src/hals/common/pins_v2.zig
```diff
@@ -186,14 +330,19 @@ pub fn Pins(comptime config: GlobalConfiguration) type {
...
-pub const Port = enum {
-    GPIOA,
-    GPIOB,
-    GPIOC,
-    GPIOD,
-    GPIOE,
-    GPIOF,
-    GPIOG,
+pub const Port = enum(u8) {
+    GPIOA = if (util.has_port('A')) 0 else undefined,
```
mattnite, line 334: Unfortunately this does not work:

```
/home/runner/work/microzig/microzig/port/stmicro/stm32/src/hals/common/pins_v2.zig:342:5: error: use of undefined value here causes illegal behavior
    GPIOI = if (util.has_port('I')) 8 else undefined,
```

The right way to do this is to create an enum at compile-time based on what ports are available. This also brings the benefit of getting rid of the backing integer.

port/stmicro/stm32/src/hals/common/pins_v2.zig
```diff
@@ -214,21 +363,35 @@ pub const Port = enum {
...
 
 pub const GlobalConfiguration = struct {
-    GPIOA: ?Port.Configuration = null,
-    GPIOB: ?Port.Configuration = null,
-    GPIOC: ?Port.Configuration = null,
-    GPIOD: ?Port.Configuration = null,
-    GPIOE: ?Port.Configuration = null,
-    GPIOF: ?Port.Configuration = null,
-    GPIOG: ?Port.Configuration = null,
+    GPIOA: ?if (util.has_port('A')) Port.Configuration else undefined = null,
```
mattnite, line 384: Same here as above

mattnite (changes requested):

mattnite:
Hey @JAicewizard heads up, the update to zig master has taken place

JAicewizard:
Thanks for the heads up! I am currently overcome by events, definitly want to finish this and get this merged.

port/stmicro/stm32/src/hals/common/pins_v2.zig
JAicewizard: I fixed this, although it is a small breaking change. Any usage of `Port.Configuration` should be replaced with `PortConfiguration`. This is not something that is implementable with generated enums

PR #940, opened 2026-04-29
rp2xxx: `pins.zig`: Refactor code to use less eval branch quota

tact1m4n3:


Grazfather (approved):
Is this avoiding having to add the add eval func, or is there some place you could remove one?

tact1m4n3:
> Is this avoiding having to add the add eval func, or is there some place you could remove one?

Someone asked a question on discord when they got an branch eval quota error. This patch is just an optimization so that the code uses less eval branch quota, but still if you pass a certain number of pin definitions, I suspect you would still get the same error. My guess is that the default branch quota of 1000 will be far out of reach with the number of pins this chip has with this patch. Previously the required a branch quota was proportional to n^2, now it is proportional to n, where n is the number of pins passed in the argument.

PR #941, opened 2026-04-29
fix spi for esp

mappoTrell:
change that clock_pin connects to the right signal (fspiclk instead of spiclk)

fix the fifo_fill function so that the rest bytes when the length of the message is not divisible by 4 get also transmitted

### DIFF
--- a/port/espressif/esp/src/hal/spi.zig
+++ b/port/espressif/esp/src/hal/spi.zig
@@ -380,7 +380,7 @@ pub const SPI = enum(u2) {
             }
         }
 
-        //add remaining bytes
+        // Add remaining bytes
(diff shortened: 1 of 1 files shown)


tact1m4n3 (approved):

port/espressif/esp/src/hal/spi.zig
```diff
@@ -379,6 +379,11 @@ pub const SPI = enum(u2) {
                 word = 0;
             }
         }
+
+        //add remaining bytes
```
Grazfather:
```suggestion
        // Add remaining bytes
```

Grazfather (commented):
good catch

PR #943, opened 2026-05-01
test: verify linker error without export_startup

mattnite:
Temporary PR to verify that `forceUndefinedSymbol("microzig_main")` produces a clear linker error when a user forgets to call `microzig.export_startup()`. Will be closed immediately after.


mattnite:
Test complete. forceUndefinedSymbol does NOT catch missing export_startup() — see analysis.

PR #944, opened 2026-05-01
test: verify linker ASSERT catches missing export_startup

mattnite:
Temporary PR to verify that the linker script `ASSERT(DEFINED(microzig_main), ...)` produces a clear linker error when a user forgets to call `microzig.export_startup()`. Will be closed immediately after verification.


mattnite:
Test complete. The linker script ASSERT successfully catches missing export_startup() with a clear error message.

PR #953, opened 2026-05-27
Zig 0.16.0

mattnite:
Besides the mechanical upgrades, I've made the following choices:

- Create a simple and maintainable `flags` library that we will standardize on for CLI tools. This will reduce churn from dependencies such as `zig-args` and `clap`.
- Remove `boxzer` packaging. It was a cool experiment, but honestly we should be improving user experience by contributing upstream to zig itself.


Grazfather:
Taking a look at this, won't submit my review yet (plus it's not done) since this is a draft, but ping me when you'd like eyes on this. Looks great so far.

PR #954, opened 2026-05-29
Zig 0.17.0 master

mattnite:


mattnite:
@Grazfather This is ready to be looked at. Once [this PR](https://codeberg.org/ziglang/zig/pulls/35515) lands and the master binary is available, this PR should pass CI

.github/workflows/ci.yml
```diff
@@ -164,6 +164,14 @@ jobs:
       - name: Build
         run: zig build -Doptimize=ReleaseSmall
 
+  # Make sure that we are building examples and invoking tests for each port
```
mattnite: Just realized that this is not complete. The idea is to have a check to make sure that every port has examples and unit tests set up in CI using some calls to `yq`.

port/espressif/esp/src/hal/usb_serial_jtag.zig
```diff
@@ -83,8 +80,13 @@ pub const logger = struct {
             else => " (" ++ @tagName(scope) ++ "): ",
         };
 
+        // TODO: do we use debug_io?
+        _ = format;
+        _ = args;
+        _ = prefix;
+
```
tact1m4n3: Can you clarify what the TODO means? Also, I have an `Io.Writer` implemented for this. If you want I can commit it.

port/raspberrypi/rp2xxx/src/hal/uart.zig
```diff
@@ -144,20 +144,83 @@ pub const instance = struct {
 pub const UART = enum(u1) {
     _,
 
-    pub const UART_With_Timeout = struct {
-        instance: UART,
+    pub const Writer = struct {
+        uart: UART,
         deadline: mdf.time.Deadline,
+        intf: std.Io.Writer,
```
tact1m4n3: Why intf as opposed to interface (as it is in the `std.Io.File.Writer`)?

core/src/microzig.zig
```diff
@@ -37,21 +37,25 @@ pub const Allocator = @import("allocator.zig");
 pub const panic = std.debug.FullPanic(struct {
     pub fn panic_fn(message: []const u8, first_trace_address: ?usize) noreturn {
         std.log.err("panic: {s}", .{message});
+        _ = first_trace_address;
 
-        var frame_index: usize = 0;
-        if (@errorReturnTrace()) |trace| frame_index = utilities.dump_stack_trace(trace);
+        // TODO: sit down and determine if we want to provide our own SelfInfo.
+        // Is that what plugs into the standard panic?
```
tact1m4n3: After a quick skim of std.debug, at least on freestanding, the standard panic just traps. Imho, we should keep a custom panic with also a custom StackIterator that doesn't depend on io. SelfInfo might not make sense for embedded as there isn't any debug info available on the device.

port/raspberrypi/rp2xxx/src/hal/uart.zig
mattnite: `interface` always feels a bit long so I wanted to see how it looked with a condensed version. I'm ambivalent on it, so when I do another pass at this patch I'll keep to `interface`.

port/espressif/esp/src/hal/usb_serial_jtag.zig
mattnite: This TODO is stale. There's a separate `std.Io` instance for the debug system, passed in through options:

https://codeberg.org/ziglang/zig/src/commit/0c41edf4978cc45b59909fce69885e310360cbfa/lib/std/debug.zig#L664

There are two boundaries we could choose for implementing the logger. The first is setting a logging function, which we've always done, and the second is to implement a `std.Io` for the standard library debug system.

I'm leaning towards implementing a logging function, similar to how you mentioned we should have our own stack iterator. I have a feeling that implementing a `std.Io` will have some unintended side effects like binary size. I've cut a ticket for research so we can have an informed answer, to be done sometime after 0.17.0 lands:

https://github.com/ZigEmbeddedGroup/microzig/issues/956

PR #957, opened 2026-06-02
rp2xxx: Implement readv_then_writev for SPI driver.

knightpp:
The implementation is missing, is this a mistake? It seems `transceive_vecs_blocking` is what it should have been.


knightpp:
It seems the method should not be implemented using transcieve.

knightpp:
After short discussion, it seems this is incorrect approach. microzig needs new methods (or new SPI interface) for full duplex communication

PR #959, opened 2026-06-13
rp2xxx fixes

Grazfather:
1. Flash size was incorrect for pico 2 and some third-party RP2350 boards. Fix them.
2. Fix link type in boot meta: It is an offset, not a pointer.
3. Explicitly set the size of the `Item` types so that we don't accidentally break them.

### DIFF
--- a/port/raspberrypi/rp2xxx/src/hal/bootmeta.zig
+++ b/port/raspberrypi/rp2xxx/src/hal/bootmeta.zig
@@ -3,7 +3,7 @@ const std = @import("std");
 const microzig = @import("microzig");
 const arch = @import("compatibility.zig").arch;
 
-pub const image_def_block = if (microzig.config.ram_image and arch == .arm) Block(packed struct {
+pub const image_def_block = if (microzig.config.ram_image and arch == .arm) Block(packed struct(u128) {
     image_def: ImageDef,
     entry_point: EntryPoint(false),
 }){
@@ -24,7 +24,7 @@ pub const image_def_block = if (microzig.config.ram_image and arch == .arm) Bloc
             .sp = microzig.utilities.get_end_of_stack(),
         },
     },
-} else Block(packed struct {
+} else Block(packed struct(u32) {
     image_def: ImageDef,
 }){
     .items = .{
--- a/port/raspberrypi/rp2xxx/src/hal/bootmeta.zig
+++ b/port/raspberrypi/rp2xxx/src/hal/bootmeta.zig
@@ -3,7 +3,7 @@ const std = @import("std");
 const microzig = @import("microzig");
 const arch = @import("compatibility.zig").arch;
 
-pub const image_def_block = if (microzig.config.ram_image and arch == .arm) Block(extern struct {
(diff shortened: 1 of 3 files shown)


port/raspberrypi/rp2xxx/src/hal/bootmeta.zig
```diff
@@ -3,7 +3,7 @@ const std = @import("std");
 const microzig = @import("microzig");
 const arch = @import("compatibility.zig").arch;
 
-pub const image_def_block = if (microzig.config.ram_image and arch == .arm) Block(extern struct {
+pub const image_def_block = if (microzig.config.ram_image and arch == .arm) Block(packed struct {
```
mattnite: Please include the backing integer for the packed struct, that's required upstream

port/raspberrypi/rp2xxx/src/hal/bootmeta.zig
```diff
@@ -24,8 +24,7 @@ pub const image_def_block = if (microzig.config.ram_image and arch == .arm) Bloc
             .sp = microzig.utilities.get_end_of_stack(),
         },
     },
-    .link = microzig.options.hal.bootmeta.next_block,
-} else Block(extern struct {
+} else Block(packed struct {
```
mattnite: ditto

mattnite (changes requested):
These then I'll force myself to read about bootmeta before merging.

port/raspberrypi/rp2xxx/src/hal/bootmeta.zig
```diff
@@ -100,8 +101,8 @@ pub const ImageDef = packed struct {
 
 pub fn EntryPoint(with_stack_limit: bool) type {
     if (with_stack_limit) {
-        return extern struct {
-            header: packed struct {
+        return packed struct(u128) {
+            header: packed struct(u32) {
```
tact1m4n3, line 105: What is the reason for replacing all externs with packed (just wondering)?

port/raspberrypi/rp2xxx/src/hal/bootmeta.zig
Grazfather: Do you mean that all packed structs will require the size? I'd rather not put it here, since it's derived from the size of its fields, and I set it there. If we were to change this, we'd have to manually add up the sizes of all of fields.

port/raspberrypi/rp2xxx/src/hal/bootmeta.zig
Grazfather: as I understand, packed structs are 'now' the correct way to bit pack. And by using them, we can say how large they are and catch in the compile if the size changes. I accidentally changed the size of a field and it bit me :)

port/raspberrypi/rp2xxx/src/hal/bootmeta.zig
mattnite: Yes, that's exactly what I mean

Grazfather: OK done.

mattnite (approved):

PR #960, opened 2026-06-13
rp2xxx: Fix Adafruit boards

Grazfather:
I noticed that my Adafruit Feather rp2350 worked when I first flashed it, but not after a reset or power cycle.

Digging through some code, I found a commit that added support for the Adafruit Feather to the arduino-pico core:

https://github.com/earlephilhower/arduino-pico/commit/163b209d#diff-d0d48a8d47fd5d31e3bc70b12159bdcb23a5ab82460bb5972cf61776f4d6768aR34-R42

It sets `PICO_XOSC_STARTUP_DELAY_MULTIPLIER=64`.

Seems that these boards have shakier clocks.

This change adds support for an optional
`xosc_startup_delay_multiplier`, which increases the startup_delay_ms when setting up the external oscillator.


mattnite (approved):
Great find, and addition! smh I hope it was worth it adafruit

PR #963, opened 2026-06-20
nrf5x: Add nice!nano board

kotokrad:
- Add nice!nano board

Tested on an original nice!nano and a "promicro nrf52840" clone.

But this board is shipped with Adafruit UF2 bootloader which supports SoftDevice, so the flashing offsets are different.
Does it make sense to add it separately from #921?

### DIFF
--- a/port/nordic/nrf5x/build.zig
+++ b/port/nordic/nrf5x/build.zig
@@ -20,6 +20,9 @@ boards: struct {
         microbit_v1: *const microzig.Target,
         microbit_v2: *const microzig.Target,
     },
+    nicekeyboards: struct {
+        nice_nano: *const microzig.Target,
+    },
 },
 
 pub fn init(dep: *std.Build.Dependency) Self {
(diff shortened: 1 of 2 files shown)


port/nordic/nrf5x/build.zig
```diff
@@ -191,6 +194,30 @@ pub fn init(dep: *std.Build.Dependency) Self {
...
+                    .preferred_binary_format = .{ .uf2 = .{ .family_id = .NRF52840 } },
+                    .chip = .{
+                        .name = "nrf52840",
+                        .url = "https://www.nordicsemi.com/products/nrf52840",
+                        .register_definition = .{
+                            .svd = nrfx.path("mdk/nrf52840.svd"),
+                        },
+                        .memory_regions = &.{
+                            .{ .tag = .flash, .offset = 0x00026000, .length = 0x100000 - 0x26000, .access = .rx },
+                            .{ .tag = .ram, .offset = 0x20002800, .length = 0x40000 - 0x2800, .access = .rwx },
```
Grazfather: Where are you getting these? This is your attempts to not trample the uf2 bootloader? Not sure if we'd want to add a separate target for if/when we want to flash directly. Is the ram allocation just for the softdevice? Does your uf2 bootloader fill that up?

kotokrad: Yeah the ram offset was from my attempts to make it work. The 0x2800 is from #921 - this is ram reserved for softdevice. Now I discovered that for disabled softdevice it should start at 0x20000008 - MBR reserves 8 bytes for interrupt forwarding.
(Also with `cpu.ram_vector_table = true` it's not needed, interrupts work with ram starting at 0x20000000).

As for flash, 0x0000-0x1000 is MBR, then everything until 0x00026000 is reserved for softdevice (S140 6.1.1 in this case, different for other versions). Writing there breaks everything.

Many popular boards use this bootloader, I think there should be support for building uf2 for it without flashing actual softdevice binaries and allocating ram.

mattnite: Heads up @kotokrad that PR was AI generated. It's from when I was doing too much of that, so I need to comb through the entire thing and validate it on hardware before any of it gets in. This PR looks good to me though

mattnite (approved):

PR #966, opened 2026-06-25
sorcerer: Update to zig 0.16

piotrfila:
Also makes regz and all the build scripts compilable with both zig 0.16 and 0.17.

The dual zig version support is probably not a great idea long term, but makes sense until dvui updates. The anyverz module could be deleted then.

I had to remove the translate-c dependency from libxml2 so that it's not neccessary to switch between dependency versions, which is not trivial.


mattnite:
The solution to this is to create a branch for microzig targetting 0.16.0. With the changes to translate-c, it is not worth the challenges of supporting both 0.16 and 0.17 in the same version.

From my upgrading, 0.16 had some issues so we're going to follow 0.17.0-dev until the next release. Once we have that next release we'll be in a better place to have sorcerer caught up.

PR #972, opened 2026-07-13
new port: MSPM0

piotrfila:
This PR adds support for the MSPM0 family as well as some improvements to regz. There is HAL support for gpio and uart so far.

I only have the MSPM0C1104 on hand, but other variants should be easy to support as well.

One issue that will arise if we want to support the variants with more peripheral instances is that all the generated registers have the peripheral instance number prefixed to their name, which should probably be removed with a new type of regz patch. I tried to add this kind of patch, but my familiarity with SQL is very limited, so I postponed it for now.

### DIFF
--- a/port/texasinstruments/mspm0/src/hal/Uart.zig
+++ b/port/texasinstruments/mspm0/src/hal/Uart.zig
@@ -91,6 +91,88 @@ pub fn read(self: Uart) ?u8 {
     return @intFromEnum(self.regs.UART0_TXDATA.read().DATA);
 }
 
+pub const Writer = struct {
+    const vtable: *const std.Io.Writer.VTable = &.{
+        .drain = drain,
+    };
+
+    instance: Uart,
+    seek: usize = 0,
+    interface: std.Io.Writer,
+
+    pub fn drain(w: *std.Io.Writer, data: []const []const u8, splat: usize) std.Io.Writer.Error!usize {
+        _ = splat;
+        const self: *Writer = @fieldParentPtr("interface", w);
+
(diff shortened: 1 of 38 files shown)


port/texasinstruments/mspm0/src/hal/Uart.zig
```diff
@@ -0,0 +1,207 @@
...
+    const reg_idx = std.meta.fieldIndex(peri_types.uart0, register_name) orelse
+        @compileError("No register " ++ register_name ++ " in uart0.");
+    const Reg = std.meta.fieldTypes(peri_types.uart0)[reg_idx].underlying_type;
+
+    const fld_idx = std.meta.fieldIndex(Reg, field_name) orelse
+        @compileError("No field " ++ field_name ++ " in uart0." ++ register_name ++ ".");
+    return std.meta.fieldTypes(Reg)[fld_idx];
+}
+
+// Maybe move this into core?
```
mattnite: Yes please do this, maybe in utilities for now?

mattnite (changes requested):
🤌 absolutely lovely patch!

Just that one suggestion for the frac div and we're good to go.

port/texasinstruments/mspm0/src/hal/Uart.zig
piotrfila: Done. I also updated some existing code to use this helper. There are more places where this could be used, but they are less trivial and I don't want to risk regressions.

mattnite (approved):
AND you added it in other places! 10/10

PR #979, opened 2026-07-15
fix(cortex-m): add missing return in set_pending/clear_pending for m0…

by965738071:
fix(cortex-m): add missing return in set_pending/clear_pending for m0…


mattnite (approved):

Grazfather:
Why return? Can't we just put the `@compileError` in an else case? That way we follow the semantics of all the switch cases.

mattnite (changes requested):
Yeah, @Grazfather is right, please put it in an else

PR #981, opened 2026-07-15
fix(msp430): correct memset and memcpy assembly operand order for TI …

by965738071:


mattnite:
How did you validate this fix?

by965738071:
The operand order in the old code was reversed: `ADD R5, R14` meant `R14 += R5` (causing count corruption), while `SUB R12, R14` left `R12` unrecoverable. The fix changes both to conform to TI's `OP src, dst → dst = dst OP src` semantics.

PR #983, opened 2026-07-17
regz: Move VirtualFilesystem into /modules

piotrfila:
I originally intended for this to just be a simple PR moving the files but the VirtualFilesystem code was offensive enough that I decided to rewrite most of it, adding more error checking along the way.

I also found and fixed a bug where Dir.openFile() only matched on the file name, but not on directory names, so if both
`foo/bar/main.zig`
and
`alice/bob/main.zig`
existed, handle to any of them could be returned.

### DIFF
--- a/modules/virtual-io/src/root.zig
+++ b/modules/virtual-io/src/root.zig
@@ -89,7 +89,14 @@ pub const vtable: *const std.Io.VTable = blk: {
             _: Dir.OpenFileOptions,
         ) File.OpenError!File {
             const vio: *VirtualIo = @ptrCast(@alignCast(userdata.?));
-            return try vio.dir_open_file(dir, sub_path) orelse error.FileNotFound;
+            const node = try vio.get_node(dir, sub_path);
+            return switch (node.value_ptr.kind) {
+                .file => .{
+                    .handle = node.key_ptr.*,
+                    .flags = .{ .nonblocking = false },
+                },
+                .dir => error.IsDir,
+            };
         }
 
         fn file_close(_: ?*anyopaque, _: []const File) void {}
@@ -141,7 +148,7 @@ pub fn deinit(vio: *VirtualIo) void {
     vio.hierarchy.deinit(vio.gpa);
 }
 
-pub fn total_file_count(vio: *VirtualIo) usize {
+pub fn total_file_count(vio: *const VirtualIo) usize {
     var ret: usize = 0;
     var it = vio.nodes.iterator();
     while (it.next()) |entry| {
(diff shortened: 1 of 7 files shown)


Grazfather (approved):
Assuming that root.zig is basically an unchanged VirtualFilesystem.zig, it looks good to me.

Why virtual-io instead virtual fs?

modules/virtual-io/src/root.zig
```diff
@@ -0,0 +1,291 @@
+//! In-memory filesystem implementation
+const std = @import("std");
+const builtin = @import("builtin");
+
+const Allocator = std.mem.Allocator;
+const assert = std.debug.assert;
```
mattnite, line 6: This is unused according to our check

mattnite (changes requested):
I don't see us needing this to become a full virtual IO system, it's just an in-memory filesystem for easy testing, so I'd keep the `VirtualFilesystem` name for both the module directory name and the file name.

piotrfila:
I figured the name was picked before the introduction of the Io interface, so if it was created now it would be called VirtualIo.

But I don't have a strong opinion on this, I'll just revert this change.

Grazfather:
I do not mind the naming, I was only curious. I will actually review the changes now. Good work on the bug fixes!

modules/virtual-io/src/root.zig
```diff
@@ -0,0 +1,291 @@
...
+        else => @compileError("Expected File or Dir, got " ++ @typeName(T)),
+    };
+}
+
+pub const Dir = struct {
+    pub const empty: Node = .{ .dir = .{ .inner = .empty } };
+    pub const kind: Node.Kind = empty;
+
+    pub const ID = enum(Node.ID) {
+        // Reserve first 256 file handles for os-specific values, such as sitdin/out/err
```
Grazfather, line 22:
```suggestion
        // Reserve first 256 file handles for os-specific values, such as stdin/out/err
```

modules/virtual-io/src/root.zig
```diff
@@ -0,0 +1,291 @@
...
+        }
+        return ret;
+    }
+
+    pub fn file_contents(vio: *const VirtualIo, file: Io.File) !*std.ArrayList(u8) {
+        return &(try (try File.ID.from_std(file)).get(vio)).inner;
+    }
+};
+
+pub fn from_handle(T: type, handle: std.posix.fd_t) !T {
```
Grazfather, line 274: Shouldn't these return `std.Io.Operation.FileWriteStreaming.Error`? Since we `try` in `write_streaming` which returns that error type.

modules/virtual-io/src/root.zig
```diff
@@ -0,0 +1,291 @@
...
+        .windows => @intFromPtr(handle),
+        else => handle,
+    })) |int|
+        @enumFromInt(int)
+    else
+        error.Unexpected;
+}
+
+pub fn to_handle(id: anytype) std.posix.fd_t {
+    // const ID = @TypeOf(id);
```
Grazfather, line 286: xxx

mattnite (approved):

piotrfila:
Was this merge intentional? There were still issues to be resolved :sweat_smile: 
I can just add them to my next PR though

modules/virtual-io/src/root.zig
piotrfila: Only assert is unused, fixed

modules/virtual-io/src/root.zig
piotrfila: from_handle is used in a few places and FileWriteStreaming.Error set just does not contain the FileNotFound error. I guess the closest would be NotOpenForWriting?
I didn't spend too much time picking these error names, in case they are encountered the trace should point to the exact error location anyway.

PR #985, opened 2026-07-17
fix(riscv32-common): use .@"struct" instead of .Struct in modify()

by965738071:
The @typeInfo union tag for structs is .@"struct" in Zig 0.17, not .Struct. This went unnoticed because modify() was never called with a packed struct CSR type.
Also add build.zig test step with 6 unit tests


mattnite:
riscv32-common needs a big overhaul, so we don't need this right now.

PR #987, opened 2026-07-17
cdc: handle short writes in flush instead of asserting

by965738071:
ep_writev can return fewer bytes than requested. The old assert was stripped in release builds, silently dropping data on short writes. Now we keep the unsent tail in the buffer and retry on the next flush.


mattnite:
How did you test?

mattnite:
This needs manual testing and your other answers around this are indicating you're finding easy bugs with an LLM.

I'm not going to accept this one without testing, so please open an issue instead.

PR #990, opened 2026-07-18
core: Mmio helpers and raw access cleanup

piotrfila:
This PR adds helper functions for manipulating Mmio and replaces references to Mmio.raw with raw.read() and raw.write()
This is a prerequisite to making Mmio aware of register access type.

I also tried to generally improve the code I touched by removing direct .raw access.

This made me aware that there is a lot of duplicate code in gpio drivers, maybe we can consolidate them somehow? At least for similar families like stm32/gd32/ch32v

### DIFF
--- a/core/src/cpus/cortex_m.zig
+++ b/core/src/cpus/cortex_m.zig
@@ -227,7 +227,7 @@ pub const interrupt = struct {
                 3 => &ppb.SHPR3,
             };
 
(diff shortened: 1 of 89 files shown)


piotrfila:
Looks like the esp examples work on debug builds but break with --release=small?? I'll take a look tommorow

piotrfila:
Main branch has the same issue:(

PR #993, opened 2026-07-23
Fix UART on rp2xxx

pcolladosoto:
When trying to run some of the examples at `examples/raspberrypi/rp2xxx/src/usb_cdc.zig` I noticed data was written into the UART in an infinite loop. After looking into it it seems data being primed into the UART TX FIFO is not being taken into account when accumulating the amount of bytes written. This (I believe) throws [`std.Io.Writer.writeAll`](https://ziglang.org/documentation/master/std/#std.Io.Writer.writeAll) into an infinite loop.

This PR simply increments `written` by whatever amount of data is primed into the TX FIFO. At least on my Pico (i.e. RP20240) this makes the UART functional again.

BTW, thanks a million for making microzig :P


mattnite (approved):
Thank you for the catch! I definitely missed that while doing the 0.17 upgrade.

pcolladosoto:
Thank you for putting microzig together! Have a great day :)

PR #997, opened 2026-07-25
board-specific uart and other config

Grazfather:
This is an attempt to make examples more generic, but putting more of the board-specific config (e.g. which pins are easy to use for uart) in a board file, while still providing an easy way for it to be overridden.


mattnite (approved):
Hey is this still wip?

Grazfather:
We can merge it. I wanted to give everyone an opportunity to weigh in / comment on the design. If you guys like it, then I'd like to use it in our other ports.

PR #1000, opened 2026-07-27
Add missing root module dependencies

piotrfila:
Right now it's a bit cumbersome to use one of microzig's modules - you have to clone the repo locally or add it as a git submodule and import it by path.

This PR exports all the modules in the root build.zig.

I'm not sure if updating all of microzig to import modules through the root project is a good idea - if we leave the imports as they are (by path) it's easier to override them for development (with zig command line options).

Also, I still need to find a good way to re-export libraries, not just modules

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -37,7 +37,6 @@ const port_list: []const struct {
 };
 
 pub fn build(b: *Build) void {
-    const target = b.standardTargetOptions(.{});
     const optimize = b.standardOptimizeOption(.{});
 
     const generate_linker_script_mod = b.createModule(.{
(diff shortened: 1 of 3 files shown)


mattnite:
I chatted to a friend outside the project about this, and ultimately I think we need to return to creating tarballs, but generating them far more often (like every push to main). This gets us the lazy dependencies and the standalone modules. It's either that or we contribute package manager features to Zig.

build.zig
```diff
@@ -822,3 +882,17 @@ inline fn custom_find_import_pkg_hash_or_fatal(comptime dep_name: []const u8) []
 
     @panic("dependency not found");
 }
+
+fn pass_through_module(b: *std.Build, name: []const u8, module: *std.Build.Module) void {
```
mattnite: Please don't do this pass through thing, I will accept the changes to the root zon file though.

mattnite (changes requested):

piotrfila:
@mattnite Is this what you have in mind?

mattnite (approved):
Yes, thank you

PR #1003, opened 2026-08-01
Patches some stuff for Ashet development

ikskuh:
These are some fixes and changes i've made for Ashet. This PR is just so i don't forget pushing them, and we can merge them eventually

### DIFF
--- a/port/raspberrypi/rp2xxx/src/hal/pio/assembler/encoder.zig
+++ b/port/raspberrypi/rp2xxx/src/hal/pio/assembler/encoder.zig
@@ -501,12 +501,23 @@ pub fn Encoder(comptime chip: Chip, comptime options: Options) type {
             var instr_index: ?u5 = program.origin orelse 0;
             for (self.tokens[self.index..]) |token| {
                 switch (token.data) {
-                    .label => |label| try program.labels.append(.{
-                        .name = label.name,
-                        .public = label.public,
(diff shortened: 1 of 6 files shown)


port/raspberrypi/rp2xxx/src/hal/pio/assembler/encoder.zig
```diff
@@ -501,12 +501,23 @@ pub fn Encoder(comptime chip: Chip, comptime options: Options) type {
...
+                        });
+                    },
                     .instruction, .word => {
+                        if (instr_index == null) {
+                            diags.* = Diagnostics.init(token.index, "out of instructions", .{});
+                            return error.Overflow;
+                        }
+
                         const result, const ov = @addWithOverflow(instr_index.?, 1);
                         instr_index = if (ov != 0) null else result;
```
Grazfather, line 522: Can't we just return an error here instead of using an optional type and adding these checks? or maybe at the check once at the top of the loop?

Grazfather (commented):
Not actually a review, just a Q about this part

port/raspberrypi/rp2xxx/src/hal/pio/assembler/encoder.zig
mattnite: @ikskuh ping

Issue #955, opened 2026-05-29
RP2040 (and likely other) broken with multiple interrupts when disabling


Kytezign:
From the RP2040 documentation, the Interrupt Clear-Enable Register will clear (disable) all enabled interrupts when written with the or as it currently is in:
[https://github.com/ZigEmbeddedGroup/microzig/blob/main/core/src/cpus/cortex_m.zig#L417](url)

This seems to be a pervasive bug assuming this is true across all the cortex_m cpus and multiple register accesses.  

<img width="1774" height="1394" alt="Image" src="https://github.com/user-attachments/assets/7d9720fa-04a6-404f-bcda-984ce009c147" />

Grazfather:
Fixed in #1006

PR #1006, opened 2026-08-06
cortex m0+: Fix bug doing read modify write on set only registers

Kytezign:
This bug can be especially bad on the NVIC_ICER where a read, modify, write can actually disable all previously enabled interrupts (rather than just the intended interrupt)

I'm not that familiar with the code base but I suspect this is an issue in other places.  I might try to look closer later but I just wanted to get this out there as at least a partial fix for an issue I ran into earlier this year (and worked around at the time).

https://github.com/ZigEmbeddedGroup/microzig/issues/955

Update to do write only rather than read, modify, write since 0 is meaningless and the read-back 1 may not have the desired effect.


mattnite (approved):
Good catch, this is definitely an area where we could add some safety abstractions to/like Mmio

PR #1015, opened 2026-08-14
elf2image: update ELF header field names for Zig 0.17

by965738071:
elf2image: update ELF header field names for Zig 0.17


mattnite:
Please note that you're updating to a newer version of zig than what we're targetting. Please use the root build.zig.zon to see what version. We're not going to update the zig version until the 0.17 release, so we don't need these changes right now.

PR #1018, opened 2026-08-25
feat: support timeouts and deadlines

### LINKED ISSUE #1017: Incorporating deadline into UART/writer interfaces in MicroZig
> Quick question that popped up when using the interfaces. Is this deadline correct? It's initialised when instantiating the `Reader` struct, but at least for the RP2040 I'm creating these deadlines with `time.deadline_in_ms()` which simply adds the argument with the current time to define an absolute timestamp when the deadline elapses.
> 
> Now, given the deadline's instantiated at the very beginning and never recreated it basically becomes a no-op as it's always elapsed later down the road.
> 
> Would it make sense to instead of passing an instantiated `mdf.time.Deadline` pass an integer (i.e. `deadline_ms`) to instantiate the deadline whenever we make a call such as a in `stream()`?
> 
> Maybe I'm just using the API wrong...
> 
> Should this be an error I'd be more than happy to provide a separate PR handling this as it affects both the `Reader` and `Writer`.
> 
> Thanks a ton for your time. 

 _Originally posted by @pcolladosoto in [#1008](https://github.com/ZigEmbeddedGroup/microzig/pull/1008/changes#r3747495467)_

pcolladosoto:
This commit adds support for specifying deadlines for read and write operations over UARTs in one of two ways:

  1. By providing a deadline as an absolute time in the future as was done before. These deadlines can be 'refreshed' through the `set_deadline` methods.

  2. By providing a timeout with which a new deadline will be computed upon performing a read or write operation.

This allows for a precise handling of deadlines when leveraging the std.Io.{Reader,Writer} interface semantics.

Closes #1017

### DIFF
--- a/port/raspberrypi/rp2xxx/src/hal/uart.zig
+++ b/port/raspberrypi/rp2xxx/src/hal/uart.zig
@@ -152,28 +152,28 @@ pub const UART = enum(u1) {
 
     pub const Writer = struct {
         uart: UART,
-        timeFrontier: TimeFrontier,
+        time_frontier: TimeFrontier,
         interface: std.Io.Writer,
 
         pub fn set_deadline(self: *Writer, deadline: mdf.time.Deadline) void {
-            self.*.timeFrontier = TimeFrontier{ .deadline = deadline };
+            self.*.time_frontier = TimeFrontier{ .deadline = deadline };
         }
     };
 
     pub const Reader = struct {
         uart: UART,
-        timeFrontier: TimeFrontier,
+        time_frontier: TimeFrontier,
         interface: std.Io.Reader,
 
         pub fn set_deadline(self: *Reader, deadline: mdf.time.Deadline) void {
-            self.*.timeFrontier = TimeFrontier{ .deadline = deadline };
+            self.*.time_frontier = TimeFrontier{ .deadline = deadline };
         }
     };
 
-    pub fn writer(uart: UART, timeFrontier: TimeFrontier, buffer: []u8) Writer {
+    pub fn writer(uart: UART, time_frontier: TimeFrontier, buffer: []u8) Writer {
         return .{
             .uart = uart,
-            .timeFrontier = timeFrontier,
+            .time_frontier = time_frontier,
             .interface = .{
                 .buffer = buffer,
                 .vtable = &.{
@@ -183,10 +183,10 @@ pub const UART = enum(u1) {
         };
     }
 
-    pub fn reader(uart: UART, timeFrontier: TimeFrontier, buffer: []u8) Reader {
+    pub fn reader(uart: UART, time_frontier: TimeFrontier, buffer: []u8) Reader {
         return .{
             .uart = uart,
-            .timeFrontier = timeFrontier,
+            .time_frontier = time_frontier,
             .interface = .{
                 .buffer = buffer,
                 .seek = 0,
@@ -203,7 +203,7 @@ pub const UART = enum(u1) {
         const uart = uart_writer.uart;
 
         var deadline: mdf.time.Deadline = undefined;
-        switch (uart_writer.timeFrontier) {
+        switch (uart_writer.time_frontier) {
             .deadline => |d| deadline = d,
             .timeout_us => |t| deadline = time.deadline_in_us(t),
         }
@@ -232,7 +232,7 @@ pub const UART = enum(u1) {
         const uart = uart_reader.uart;
 
         var deadline: mdf.time.Deadline = undefined;
-        switch (uart_reader.timeFrontier) {
+        switch (uart_reader.time_frontier) {
             .deadline => |d| deadline = d,
             .timeout_us => |t| deadline = time.deadline_in_us(t),
         }
--- a/port/raspberrypi/rp2xxx/src/hal/uart.zig
+++ b/port/raspberrypi/rp2xxx/src/hal/uart.zig
@@ -206,7 +206,6 @@ pub const UART = enum(u1) {
             .deadline => |d| d,
             .timeout_us => |t| time.deadline_in_us(t),
         };
-        // logg.debug("hello from drain with deadline {any}", .{deadline});
 
         // bytes from buffer are not included in count.
         w.end -= uart.write_blocking(w.buffer[0..w.end], deadline) catch |err| switch (err) {
@@ -234,7 +233,6 @@ pub const UART = enum(u1) {
             .deadline => |d| d,
             .timeout_us => |t| time.deadline_in_us(t),
         };
-        // logg.debug("hello from stream with deadline {any}", .{deadline});
 
         return switch (limit) {
             .nothing => 0,
--- a/port/raspberrypi/rp2xxx/src/hal/uart.zig
+++ b/port/raspberrypi/rp2xxx/src/hal/uart.zig
@@ -156,7 +156,7 @@ pub const UART = enum(u1) {
         interface: std.Io.Writer,
 
         pub fn set_deadline(self: *Writer, deadline: mdf.time.Deadline) void {
-            self.*.time_frontier = TimeFrontier{ .deadline = deadline };
+            self.time_frontier = TimeFrontier{ .deadline = deadline };
         }
     };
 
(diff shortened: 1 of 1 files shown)


pcolladosoto:
Hi @Grazfather, @mattnite! Feel free to make any comments and/or changes: I'm by no means a Zig expert!

For instance, I believe the explicit dereference in the `set_deadline` methods might not be needed and I'd be fine with deleting it. Also, the naming can surely be improved (i.e. I'm not completely sold on the `TimeFrontier` type name) and some documentation can be added; I just deferred that a bit to get a feel for what might be needed to meet your requirements.

Thanks a ton for your time!

port/raspberrypi/rp2xxx/src/hal/uart.zig
```diff
@@ -135,6 +135,8 @@ pub const instance = struct {
     }
 };
 
+pub const TimeFrontier = union(enum) { timeout_us: u64, deadline: mdf.time.Deadline };
```
tact1m4n3: nit: maybe format this on multiple lines?

port/raspberrypi/rp2xxx/src/hal/uart.zig
```diff
@@ -148,20 +150,28 @@ pub const UART = enum(u1) {
 
     pub const Writer = struct {
         uart: UART,
-        deadline: mdf.time.Deadline,
+        timeFrontier: TimeFrontier,
```
tact1m4n3: nit: fields should be snake case

port/raspberrypi/rp2xxx/src/hal/uart.zig
```diff
@@ -148,20 +150,28 @@ pub const UART = enum(u1) {
...
         interface: std.Io.Writer,
+
+        pub fn set_deadline(self: *Writer, deadline: mdf.time.Deadline) void {
+            self.*.timeFrontier = TimeFrontier{.deadline = deadline};
```
tact1m4n3: Doing self.* here is not necessary. Field access dereferences the pointer automatically

port/raspberrypi/rp2xxx/src/hal/uart.zig
tact1m4n3: You can add here a 
```zig
pub const no_deadline: TimeFrontier = .{ .deadline = .no_deadline };

```
so you can do .no_deadline directly when you don't want any deadline.

port/raspberrypi/rp2xxx/src/hal/uart.zig
```diff
@@ -190,18 +200,24 @@ pub const UART = enum(u1) {
         const uart_writer: *Writer = @alignCast(@fieldParentPtr("interface", w));
         const uart = uart_writer.uart;
 
+        var deadline: mdf.time.Deadline = undefined;
+        switch (uart_writer.timeFrontier) {
+            .deadline => |d| deadline = d,
+            .timeout_us => |t| deadline = time.deadline_in_us(t)
+        }
```
tact1m4n3: nit: switch can be used as an expression so you could do
```zig
const deadline: mdf.time.Deadline = switch (uart_writer.time_frontier) {
    .deadline => |d| d,
    .timeout_us => |t| time.deadline_in_us(t),
};
```

port/raspberrypi/rp2xxx/src/hal/uart.zig
```diff
@@ -559,7 +582,7 @@ var uart_logger: ?UART.Writer = null;
 ///     .logFn = hal.uart.log,
 /// };
 pub fn init_logger(uart: UART) void {
-    uart_logger = uart.writer(.no_deadline, &.{});
+    uart_logger = uart.writer(.{.deadline = .no_deadline}, &.{});
```
tact1m4n3: Can become `uart.writer(.no_deadline)`. See the suggestion above

tact1m4n3 (approved):
Solid work! I like the direction of this. I only have a few style suggestions.

tact1m4n3:
Also CI formatting seems to fail. You should run zig fmt on the file. Or you can just setup your editor to do that on save so you don't have to think about it.

port/raspberrypi/rp2xxx/src/hal/uart.zig
pcolladosoto: Apparently `zig fmt` doesn't like line 138 being formatter on multiple lines...

Also, I defined the `no_deadline` constant: thanks for the idea!

port/raspberrypi/rp2xxx/src/hal/uart.zig
pcolladosoto: Yup! Thanks for spotting the deviation from the naming convention. It should be corrected.

port/raspberrypi/rp2xxx/src/hal/uart.zig
pcolladosoto: Yeah... I actually read that on [the docs](https://ziglang.org/documentation/master/#struct) but preferred the explicit dereference. I'm happy to adhere to the standard approach though. Thanks a ton!

port/raspberrypi/rp2xxx/src/hal/uart.zig
pcolladosoto: Much better! Thanks a ton for the suggestion; it should be applied.

port/raspberrypi/rp2xxx/src/hal/uart.zig
pcolladosoto: Thanks! Actually `.no_deadline` caused the CI to fail though. I'm currently trying to run with `uart.writer(no_deadline)` which will hopefully make the pipelines pass.

pcolladosoto:
Hi @tact1m4n3! Thanks a ton for the input. I guess everything should be addressed and the pipelines are passing except for a the website generation which apparently fails when not being able to pull an external dependency.

I'd be more than happy to squash all the commits before merging so as not to pollute `main` too much: up to you :P

Thanks again for the input!

port/raspberrypi/rp2xxx/src/hal/uart.zig
Grazfather: Put a comma before the last `}` an it'll format into separate lines. The zig formatter only formats on multiple lines if the last entry has a trailing comma

pcolladosoto: Done! Thanks a ton for the tip :P

port/raspberrypi/rp2xxx/src/hal/uart.zig
```diff
@@ -135,6 +135,13 @@ pub const instance = struct {
     }
 };
 
+pub const TimeFrontier = union(enum) {
+    timeout_us: u64,
+    deadline: mdf.time.Deadline,
+};
+
+pub const no_deadline: TimeFrontier = .{ .deadline = .no_deadline };
```
mattnite: One last thing, please move this into the TimeFrontier body, it will achieve what @tact1m4n3  was after.

Because its true name would be `TimeFrontier.no_deadline`, that means any time you pass this type into a function, you could simply provide `.no_deadline`.

When you see things start with a dot in zig, the programmer is telling the compiler to infer the type. As a function parameter, it knows the expected type, so it try's to match the dot name to one of the declarations in the type. The way this is currently set up, the declaration is in the parent namespace, the compiler only looks in the type itself.

Here is a good thread on it: https://ziggit.dev/t/whats-the-magic-behind-the-function-call-inference/14823

tact1m4n3: I meant to add no_deadline as a decl inside the TimeFrontier container. No worries, sorry I wasn't very specific. This has a very nice property (see zig declaration literals; because the type of the constant is the same as the type of the container) that wherever you would initialize a `TimeFrontier` you can use the .no_deadline notation (the same syntax as for an enum field).

```suggestion
pub const TimeFrontier = union(enum) {
    timeout_us: u64,
    deadline: mdf.time.Deadline,
    
    pub const no_deadline: TimeFrontier = .{ .deadline = .no_deadline };
};
```

port/raspberrypi/rp2xxx/src/hal/uart.zig
tact1m4n3: See my comment bellow.

port/raspberrypi/rp2xxx/src/hal/uart.zig
pcolladosoto: Super! Thanks a ton for the information on leveraging this notation: it really helps with writing idiomatic Zig :P

I just committed the changes.

Issue #221, opened 2024-07-14
ESP support tracking


DNedic:
This issue tracks the support status for ESP32 series Espressif SoCs and will be updated with new tasks continously:

- [x] [Add GPIO driver](https://github.com/ZigEmbeddedGroup/microzig/pull/216)
- [ ] Add UART driver
- [ ] Add support for the ESP32-C6
- [x] Interrupt support
- [ ] Add support for digital addressable RGB LEDs found on many devboards (WS2811 compatible)
- [x] Add support for the ESP-IDF Application Image Format
- [ ] Enable caches

Grazfather:
#415 should knock off this one when it's ready.

> Add support for the ESP-IDF Application Image Format

Grazfather:
We also need an i2c driver, and most of these, to follow our standard hal api, will need a time driver so that we can use deadlines, so I think the time device should be next.

Uart was done by @tact1m4n3 in #436

tact1m4n3:
Interrupts can be checked 😄

PR #1024, opened 2026-09-04
port/espressif: add an esp32c6 target and a XIAO ESP32C6 board

zmscode:
Adds `esp32_c6`, `esp32_c6_direct_boot` and `esp32_c6_flashless` next to the esp32c3 targets, plus a Seeed Studio XIAO ESP32C6 board definition. This closes out the "Add support for the ESP32-C6" item on #221.

## What works

Chip bring up (clocks, cache, watchdogs, interrupt controller), gpio, the system timer and the usb serial/jtag logger. `blinky`, `custom_clock_config`, `gpio_input` and `systimer` build for the C6 across all three boot modes; `i2c`, `spi`, `ledc`, `rtos` and `radio` are still esp32c3 only and are gated out of the C6 example matrix.

## Structure

Where a peripheral differs the hal now forwards to a per chip module the way `clocks.zig` already did. `hal/system`, `hal/gpio` and `cpus/esp_riscv` gained chip subdirectories, and the cpu learns which chip it is on through a new option in its `cpu-config` module. The esp32c3 code moved into those subdirectories unchanged except where noted below.

## Chip differences worth knowing about

- The esp32c6 gates each peripheral from its own PCR register instead of the two `PERIP_CLK_ENx` bitmasks, so `PeripheralMask` became a flag set with a comptime mapping table onto PCR registers, plus a comptime check that every mask field is actually backed by a mapping.
- Its pll is fixed at 480 MHz with the cpu frequency taken from a divider, and its apb clock stays at 40 MHz rather than following the cpu.
- Its rom does not export the regi2c helpers, so the analog side of the pll is left as the rom bootloader calibrated it. Re-running that calibration needs a `LP_I2C_ANA_MST` driver and is only necessary once the hal starts powering the pll down.
- `INTERRUPT_CORE0` and `INTPRI` split the interrupt matrix from the cpu interrupt controller, and there are 77 sources instead of 62.
- The rtc watchdogs moved into the low power domain.
- 31 pads instead of 22, usb serial/jtag on GPIO12/13, and no pads with a swapped 10/20 mA drive strength.
- Instruction and data share one address space for both flash and sram, so `ld/esp32_c6/image_boot_sections.ld` puts the rodata first (the app descriptor has to start the first mapped segment) and walks a single region instead of offsetting the two views against each other with dummy sections, which overlap when the two windows coincide.

## Incidental fixes

Three bugs in the shared cpu interrupt code fell out of the split, all of which also affected the esp32c3:

- The interrupt priority registers were indexed through an array sized by the *bit width* of the interrupt enum (5) rather than its range (32).
- `get_type` had an operand precedence error (`raw & (1 << num) >> num`) that made it always read bit 0.
- `get_mapped_interrupt` truncated the mapped interrupt number to four bits, losing interrupts 16 to 31.

## Vendored SVD

`src/chips/ESP32-C6.svd` is the esp-pacs base svd with one local change, noted in the port README: a `<dimIndex>1,2</dimIndex>` on `DMA.OUT_CONF0_CH%s`. Without it that array is numbered from zero and collides with the explicit `OUT_CONF0_CH0`.

## Testing

**Not tested on hardware — I do not have a board, which is why this is a draft.** What has been verified:

- All three C6 boot modes build, as does the XIAO board example.
- All 40 existing esp32c3 firmware images still build unchanged.
- The generated esp image parses correctly and matches the esp32c3 image structure: app descriptor at the start of the first mapped segment (`0x42000020`), text on the next mmu page (`0x42010020`), chip id `0x000d`, entry point matching the text segment.

Someone with a C6 or a XIAO putting this on silicon would be very welcome. The most likely first-boot problem is the pll path, since the analog calibration is inherited from the rom rather than redone.


mattnite:
If you don't have this hardware, why did you make this patch?

Grazfather:
cool that you added this, but less cool that you're throwing this over the wall and trying to get us to test it (on hardware!)

Why don't you buy those boards and test them out? Then we can merge this.

PR #1049, opened 2026-09-15
drivers: ICM_20948: Support disabling logging

Grazfather:
Support disabling logging.

I needed to do this for ch32v203. The logging added something like 10kB to the binary!


drivers/src/sensor/ICM_20948.zig
```diff
@@ -35,865 +35,879 @@
...
-    fn err(comptime format: []const u8, args: anytype) void {
-        if (!builtin.is_test)
-            std.log.err(format, args);
-    }
+pub const Options = struct {
+    /// Controls which log messages the driver compiles in.
+    /// Works like `std_options.log_level`: messages at this severity or
+    /// higher are included. Set to `null` to compile out all logging.
+    /// Default `.err` includes only error messages.
+    log_level: ?std.log.Level = .err,
```
mattnite, line 43: Are you aware of std.log.scoped and how you can configure the log level for different scopes in std_options?

Grazfather: I am, but the logging in this function is most all error level, so I wanted a way to disable it explicitly.

mattnite: Good enough for me

mattnite (approved):

PR #1052, opened 2026-09-16
cortex_m: Various fixes and cleanup

tact1m4n3:
* Add dcb and dwt (register defs generated using llm, I tested only a few registers I needed)
* Rename ExternalInterrupt to just Interrupt (also on hazard3 for consistency)
* Fix enable, disable, set_pending, clear_pending. They previously used an `or` operation doing a rmw which was unnecessary and not atomic
* Fix interrupt priorities. Now we get the priority bits from chip properties
* Cleanup interrupt.exception: the implementation didn't work, used peripherals.ppb which didn't exist. In the future this could be added directly in the interrupt functions with a comptime switch


mattnite:
Please verify all the definitions by reading through the references.

To what extent did you test the rest of the code?

tact1m4n3:
Hey! Sorry, I should've checked the register definitions. If even this an LLM can't get right... 😄 I pushed the revised registers. The interrupt function changes I tested in my flight controller firmware where I use interrupts to schedule tasks. There was quite an annoying bug a while ago because of the rwm in the interrupt pend functions where some tasks wouldn't get scheduled because the interrupt wouldn't get reliably pended. The `Interrupt` renaming and exceptions up for debate, I can revert them if they are not wanted.

mattnite:
FWIW, I didn't find any issues, when I read:

> register defs generated using llm

I take it to mean "I have not validated this, nor do I want to be accountable if it's wrong". And that's why I asked you to review them. It's okay for PRs to be less than perfect, a work in progress, or a process of collaboration. It's not okay for them to be low effort.

I'm pretty busy with SYCL, I'll have more time to review this once it's done.

PR #1053, opened 2026-09-16
core: Bring back stack traces

tact1m4n3:


core/src/microzig.zig
```diff
@@ -85,6 +93,8 @@ pub const Options = struct {
     /// reduce code size as the string literals for error names no longer have to
     /// be included in the executable.
     simple_panic_if_main_errors: bool = false,
+
+    panic_stack_trace: bool = builtin.optimize.runtimeSafety(),
```
tact1m4n3: should we enable stack traces by default in both debug and safe modes? Or just in debug? I noticed that in release safe the addresses are off by a lot anyway so it might only be worth it in debug. also error traces are enabled by default by the zig compiler only in debug

mattnite (approved):

core/src/microzig.zig
mattnite: let's keep them on for debug and safe by default

